PR #1, opened 2017-12-16
Add some Player related method implementations

Sauilitired:
Specifically related to banning and OfflinePlayers.


Sauilitired:
Yeah, I'm sorry about breaking the test. For some reason my setup didn't detect any tests, and so it declared that everything passed. I've fixed that locally.

I'll make sure to actually implements tests in the future, I was just in a rush to get this pushed. I am not normally this messy, promise :cow: 

Anywho, you've got a nice project going here. I would really like to use it in PlotSquared once all the required components are implemented.

seeseemelk:
Ah well that's okay.
I added a couple of tests for BanList, found one method that threw an exception but aside from that one nothing else.
It's so nice to see someone is actually interested in using it.

PR #3, opened 2018-07-02
Update property PlayerMock and update version gradle

BigTows:


seeseemelk:
I've merged this as it looks pretty good.
I also added two small unit tests for the display name as sometimes the Bukkit API can be awkward to reimplement and I accidentally break tiny stuff from time to time.

PR #21, opened 2019-05-16
Enable Player EXP mocking

james9909:
Implements the various experience related methods that belong to `Player`. Let me know if I missed anything!

### DIFF
--- a/src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
+++ b/src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
@@ -27,6 +27,8 @@ import org.bukkit.event.block.BlockDamageEvent;
 import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.event.entity.PlayerDeathEvent;
 import org.bukkit.event.inventory.InventoryType;
+import org.bukkit.event.player.PlayerExpChangeEvent;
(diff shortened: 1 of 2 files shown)


src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -504,20 +508,162 @@ public void chat_AnyMessage_AsyncEventFired()
...
+			@EventHandler
+			public void onLevelChangeEvent(PlayerLevelChangeEvent event)
+			{
+				levelCount.incrementAndGet();
+			}
 
+			@EventHandler
+			public void onExpChangeEvent(PlayerExpChangeEvent event)
+			{
+				fail("PlayerExpChangeEvent should not be called");
```
MiniDigger: is this bukkit behavior? I looked at the javadocs and didn't find anything about that, as I user I would expect both events to be called (since the total exp changed?)

james9909: According to [the docs](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/player/PlayerExpChangeEvent.html), `PlayerExpChangeEvent` is only called "when a players experience changes naturally". Artifically calling `giveExp` through the API is not natural behavior. I also found [a thread](https://bukkit.org/threads/playerexpchangeevent-not-running.367378/) where somebody asked about this behavior themselves.

PR #35, opened 2020-02-18
Fix lower case issues

ShadowJonathan:
`getPlayerExact` was not properly emulating craftbukkit, craftbukkit makes search of player names explicitly case-insensitive.

`matchPlayer` could potentially have a bug regarding Locale where one side had been `toLowerCase`'d with a different `Locale` compared to the other.


seeseemelk:
Looks good. Thanks!

seeseemelk:
If I remember to do so later today, I will publish the version to Maven Central.

PR #41, opened 2020-04-09
Implement some Eye related functions

HeroCC:
Defaults to standard Bukkit player height, is changed by sneaking. Tests to verify all pass on my end, though I'd love it if someone could independently verify these constants on various server versions. Eventually we should use the headHeight rather than constants, but for now this is good.


seeseemelk:
Looks awesome, and the constants seem to be correct if the Minecraft wiki can be trusted.

PR #49, opened 2020-04-18
Feature/multi plugin

seeseemelk:
This pull request adds supports for multi-plugin unit tests.
As some plugins depend on other plugins, it is necessary to be able to load those other plugins into memory.

This pull request adds `MockBukkit.loadJar(string jarFile)`, which can load a plugin from a jar file.


HeroCC:
Could this be used to do, say, `testImplementation 'my.plugin:here:version'` in gradle, and load it in MockBukkit?

seeseemelk:
That would be difficult. The plugin itself (not the API jar) would have to be on some maven repository and it would have to have the extra constructor as required by MockBukkit. This is because the default super constructor from Bukkit expects a special ClassLoader to be used, which we cannot control for any Gradle-managed dependencies.
So it would only work if the plugin is available from a repository and has a special constructor.

PR #53, opened 2020-05-03
Added EnchantedBookMeta mock

TheBusyBiscuit:
This Pull Request adds a mock-version for the EnchantmentStorageMeta class which is used by Enchanted Books.
I have updated the ItemFactory accordingly to incorporate this new class too.

I also implemented the Entity#getType() method for the MockZombie class to return EntityType.ZOMBIE

Thanks for creating this library btw!


seeseemelk:
Looks good. Would be nice to have some unit tests to make sure that this doesn't break at some point in the future, but it's better to have no tests than to not have the functionality.
I also updated it as Bukkit changed stuff without bumping version numbers, just so the CI wouldn't complain.

TheBusyBiscuit:
Yeah I can add adequate Unit Tests for that in a follow up PR.
Thanks!

PR #55, opened 2020-05-03
Fixed constructors for ItemMetas

TheBusyBiscuit:
I first didn't realize the ItemMeta Mock classes would need constructors, since none of them had any.
However while writing Unit Tests for them I ran into a `NoSuchMethodException`.
Apparently every Mock ItemMeta class requires a second constructor which takes its original counterpart as an argument as seen here:
https://github.com/seeseemelk/MockBukkit/blob/v1.15/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemFactoryMock.java#L102

This Pull Request adresses and fixes this.
Also had I written Unit Tests before, this would have been noticed quicker but oh well.

P.S. Sorry in advance for bombarding you with pull requests, there will be more down the line because I intend to use this library for a big project.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
@@ -31,7 +31,7 @@ public class BookMetaMock extends ItemMetaMock implements BookMeta {
         
         this.title = meta.getTitle();
         this.author = meta.getAuthor();
-        this.pages = meta.getPages();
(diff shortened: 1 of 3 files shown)


seeseemelk:
I'm perfectly okay with being bombarded with pull request. I'm actually super glad that this library is useful to someone else but me!
Note that during the week my response times will be a lot longer because work.

TheBusyBiscuit:
Same for me, I also need to work during the week, albeit from home. So there probably won't be many pull requests in the first place.

PR #57, opened 2020-05-04
Persistent Data & Unit Tests

TheBusyBiscuit:
Good evening, I have added Unit Tests for the newly added ItemMeta classes respectively.
Furthermore I also implemented a Mock class for the PersistentDataContainer in Bukkit, here's a full rundown of all the changes:

* Updated README
  * Since many people may actually use Maven over Gradle, I have added instructions on how to add MockBukkit via Maven too
* PersistentData
  * A new Mock class for the PersistentDataContainer was introduced
  * Also a Mock class for PersistentDataAdapterContext, although that one is a little... ridiculous
  * The PersistentDataContainer Mock was also implemented for all Entity and ItemMeta Mocks
* EnderChestInventoryMock
  * Just a simple extension of the InventoryMock class
  * Each PlayerMock now has an instance of this mock and returns/instantiates it upon calling #getEnderChest()
* Unit Tests
  * The Unit Tests for ItemMeta have been expanded upon
  * New Unit Tests for the newly added ItemMeta mocks have been added
  * New Unit Tests for the PersistentDataContainer Mock were also added
* ItemFactory
  * Some stuff in the ItemFactory Mock was fixed, the BookMeta is now correctly assigned to Written books instead of normal books
  * The "asMetaFor" class was improved, we now get the most suitable constructor via reflection rather than relying on a specific one and falling back onto a standard one.


seeseemelk (approved):

seeseemelk:
If you ever want to see a proper release, just open an issue and I'll get on it once I have time.

TheBusyBiscuit:
Alright, I personally don't have any needs in that regard though. With jitpack I am targeting the specific commit anyway, so I am not really in need of any releases.
I will let you handle your release management, I'm good :D

PR #58, opened 2020-05-06
Updated to latest Bukkit version and fixed compilation

TheBusyBiscuit:
Bukkit has apparently introduced two new methods to the "PlayerInventory" interface which prevented the project from compiling.
The new methods were added to the PlayerInventoryMock to account for that.


seeseemelk:
Ugh, I hate this stuff. Why can't Bukkit use proper versioning? It would prevent these sorts of issues

TheBusyBiscuit:
Fully agree on this, it causes more headaches than benefits.

PR #60, opened 2020-05-07
Implemented keepInventory, added an OfflinePlayerMock and more methods to Player/Inventory Mocks

TheBusyBiscuit:
Alright, here's the next batch.
For one of the changes I would like to hear your opinion on it:
As of now the Player interface is solely represented by PlayerMock and the Mock implementation for Server#getOfflinePlayer(...) seems to just refer to the Player implementation. For the time being I just added a OfflinePlayerMock class which has no relation to the PlayerMock class, it's only purpose is providing throwaway OfflinePlayer mocks for methods such as Skullmeta#getOwningPlayer().
This is by far not ideal and I would like to hear your opinion on this matter, for example on where the Server#getOfflinePlayer() method should return those mocks too or whether there is some way the PlayerMock and OfflinePlayerMock could work together without any issues.
Anyway, other than this little problem, the following things were introduced:

* KeepInventory Gamerule
  * Upon Player death the Inventory of the Player will be cleared unless KeepInventory was enabled in the current world, this allows Unit Tests to correctly get the state of the Player after death, since they would drop their inventory
  * The Inventory#clear() methods have also been implemented while I was at it
* The SkullMetaMock was finished (see above for the OfflinePlayerMock)
* Compass Targets and Bed Spawn locations were added to the PlayerMock class
* More Unit Tests (There is no such thing as "too much" unit tests anyway, is there?)

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
@@ -1255,11 +1255,10 @@ public class WorldMock implements World
 		throw new UnimplementedOperationException();
 	}
 	
-	@SuppressWarnings("unchecked")
 	@Override
 	public <T> T getGameRuleValue(GameRule<T> rule)
 	{
-		return (T) gameRules.get(rule);
+		return rule.getType().cast(gameRules.get(rule));
 	}
 	
 	@Override
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
@@ -468,8 +468,7 @@ public abstract class EntityMock implements Entity, MessageTarget
 	@Override
 	public boolean isValid()
 	{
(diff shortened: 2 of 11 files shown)


seeseemelk (approved):
The changes look good. Sometimes the formatting is a bit odd, I think there might be some tab/spaces conflict going on, but that's okay.

seeseemelk:
Currently there is the method `ServerMock#addPlayer()`. Maybe an additional `ServerMock#addOfflinePlayer()` which adds an offline player to the server. Then, it could be possible to add an `OfflinePlayerMock#join()` which causes that player to 'join' the server, and a `PlayerMock#leave()` which causes that player to become an offline player.

After checking the CraftBukkit source code, it seems that `getOfflinePlayers` will return a normal `CraftPlayer` object **if** that player is online. It will only return a `CraftOfflinePlayer` if the player is both offline and no `CraftPlayer` instance exists for that player. (Or so I think).

So adding a `ServerMock#addOfflinePlayer` that creates an `OfflinePlayerMock` is *probably* the best course of action.

Issue #61, opened 2020-05-10
Allow custom implementations of ServerMock


Silthus:
Currently the `ServerMock` instance is statically created and there is no option to pass a custom implementation of the mock.

**Feature Request**
It would be nice to be able to pass custom implementations of `ServerMock` to the `MockBukkit.mock()` method. That would also solve the issue to rely on Pull Requests being merged to implement missing methods in the Bukkit implementation.

I will send a PR for that shortly.

PR #62, opened 2020-05-10
feat(core): overload mock() to allow custom ServerMock implementations

Silthus:
Implemented #61 

* Overloaded `MockBukkit.mock(...)` to accept custom implementations of `ServerMock`
* Added documentation on how to use this

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
@@ -56,7 +56,6 @@ public class MockBukkit
 	 * @param serverMockImplementation your custom {@link ServerMock} implementation.
 	 * @return The provided {@link ServerMock}.
 	 */
-	@SuppressWarnings("unchecked")
 	public static <TServerMock extends ServerMock> TServerMock mock(TServerMock serverMockImplementation)
 	{
 		if (mock != null)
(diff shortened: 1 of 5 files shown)


src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
```diff
@@ -44,20 +44,34 @@ protected static void setServerInstanceToNull()
...
+		mock = serverMockImplementation;
+
 		Level defaultLevel = mock.getLogger().getLevel();
 		mock.getLogger().setLevel(Level.WARNING);
 		Bukkit.setServer(mock);
 		mock.getLogger().setLevel(defaultLevel);
-		
-		return mock;
+
+		return (TServerMock) mock;
```
TheBusyBiscuit: You could also just return "serverMockImplementation" here since its the same object. This would also allow you to remove this unnecessary cast and the warnings supressor. Other than that I find this to be a good idea and hope this gets merged!

Silthus: Yeah thanks for catching that. I updated the code.

seeseemelk (approved):

seeseemelk:
This looks really useful. I'm gonna merge it and then push a new release to maven central.

PR #67, opened 2020-05-14
Fixes #59, implements serialization for the ItemMetaMock

### LINKED ISSUE #59: ItemMeta Serialization Methods
Currently, the ItemMeta serialization (and I presume deserialization) methods are not implemented. My use case is serializing and deserializing inventory items and using tests to make sure that happens as expected. This involves ItemMeta, and I really don't want to maintain my own implementation.

If I find the time, I can probably implement this myself and PR it, but in case someone else wants to get to it before me, I'll just leave this issue here. :)

maxphillipsdev:
Fixes #59 
Implemented the `serialize()` and `deserialize()` methods for `ItemMetaMock` and wrote tests for them. `ServerMock` was modified to register the `ItemMetaMock` as being serializable with the `ConfigurationSerializable` interface.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java
@@ -21,7 +21,7 @@ import java.util.*;
 import static java.util.Objects.nonNull;
 
 @SuppressWarnings("deprecation")
-public class ItemMetaMock implements ItemMeta, Damageable, ConfigurationSerializable {
+public class ItemMetaMock implements ItemMeta, Damageable {
     
     private String displayName = null;
     private List<String> lore = null;
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java
@@ -7,6 +7,7 @@ import com.google.common.collect.Multimap;
 import com.google.common.collect.MultimapBuilder;
 import org.bukkit.attribute.Attribute;
 import org.bukkit.attribute.AttributeModifier;
+import org.bukkit.configuration.serialization.ConfigurationSerializable;
 import org.bukkit.enchantments.Enchantment;
 import org.bukkit.inventory.EquipmentSlot;
 import org.bukkit.inventory.ItemFlag;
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java
```diff
@@ -19,7 +21,7 @@
 import static java.util.Objects.nonNull;
 
 @SuppressWarnings("deprecation")
-public class ItemMetaMock implements ItemMeta, Damageable {
+public class ItemMetaMock implements ItemMeta, Damageable, ConfigurationSerializable {
```
TheBusyBiscuit: ItemMeta already extends `ConfigurationSerializable` so this would be somewhat redundant. Is there any particular reason you implemented it here too?

src/test/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMockTest.java
```diff
@@ -345,4 +343,40 @@ public void assertCustomModelData() {
 		assertTrue(meta.hasCustomModelData());
 		assertEquals(100, meta.getCustomModelData());
 	}
+
+	@Test
+	public void assertSerialize() {
+
+		// Tests for displayName, Lore, enchants, unbreakable status, and damage
+		meta.setDisplayName("Test name");
+		meta.setLore(new ArrayList<String>() {{ add("Test lore"); }});
```
TheBusyBiscuit: Should probably rather use Arrays.asList() here.

src/test/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMockTest.java
```diff
@@ -345,4 +343,40 @@ public void assertCustomModelData() {
...
+		meta.setUnbreakable(true);
+		meta.setDamage(5);
+
+		HashMap<String, Object> expected = new HashMap<String, Object>() {
+			{
+				put("displayName","Test name");
+				put("lore", new ArrayList<String>() {{ add("Test lore"); }});
```
TheBusyBiscuit: Should probably rather use Arrays.asList() here.

TheBusyBiscuit (changes requested):

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java
maxphillipsdev: Ah my bad! Removing it and replacing the ArrayLists in the next commit.

seeseemelk (approved):

seeseemelk:
Thanks for the addition!

PR #69, opened 2020-05-18
Added Sound Assertions for PlayerMock

TheBusyBiscuit:
As a follow up to my last PR, there is now a method to assert whether a Player has heard a certain Sound or not.

In addition to that I fixed some of my earlier mistakes with PotionMetaMock and the drops for the EntityDeathEvent.
Also addd a .clearEvents() method to the PluginManagerMock in order to clear the "memory" of the PluginManager from any events.

+ Implemented InventoryCloseEvent when calling player.closeInventory()


seeseemelk (approved):
I love the class name `AudioExperience` for some reason :stuck_out_tongue_closed_eyes:

PR #73, opened 2020-06-05
Added time/environment methods to WorldMock

syldium:
This pull request adds the following things :
- an `Bukkit.isPrimaryThread()` implementation which redirects to the existing `#isOnMainThread()`  method
- an implementation of time-related methods in WorldMock
- environment getter and setter for WorldMock

I also added the corresponding unit tests.


seeseemelk (approved):
Thanks! The changes look really good! I actually didn't even know that these functions existed until this pull request :+1:

PR #77, opened 2020-06-16
Fixed #76, deprecated build.gradle, and incomplete javadocs

### LINKED ISSUE #76: Inventory#getItem returns AIR instead of null
The method is explicitly nullable, and will return null if the inventory doesn't have an item at that slot.

From CraftInventory:
`@Override
public ItemStack getItem(int index) {
    net.minecraft.server.ItemStack item = getInventory().getItem(index);
    return item.isEmpty() ? null : CraftItemStack.asCraftMirror(item);
}`

BigBadE:
I didn't update the gradle wrapper from 6.3 to 6.5, but the build.gradle is not compatible with Gradle 7.

I made it so the jar isn't automatically signed on build, but it is signed on publish (publishMockBukkitToRepository/publishMockBukkitToSnapshotRepository).

#76 is fixed too.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'java-library'
 apply plugin: 'eclipse'
 apply plugin: 'idea'
 apply plugin: 'jacoco'
-apply plugin: 'maven'
+apply plugin: 'maven-publish'
 apply plugin: 'signing'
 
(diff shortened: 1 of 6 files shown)


seeseemelk:
Awesome, thanks for these changes!

Issue #74, opened 2020-06-07
BlockMock#setBlockData missing


hafarooki:

WalshyDev:
PR it ;)

LinoxGH:
This issue has been fixed.

PR #78, opened 2020-06-26
Added the velocity methods for EntityMock and BlockDataMock.

LinoxGH:
Added simple and surface level velocity methods to EntityMock.
Added BlockDataMock for Issue #74.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
@@ -29,6 +29,7 @@ import org.bukkit.util.Vector;
 
 import be.seeseemelk.mockbukkit.UnimplementedOperationException;
 import junit.framework.AssertionFailedError;
+import org.jetbrains.annotations.NotNull;
 import org.junit.Assert;
 
 public class BlockMock implements org.bukkit.block.Block
@@ -374,9 +375,8 @@ public class BlockMock implements org.bukkit.block.Block
 	}
 
 	@Override
-	public void setBlockData(BlockData data)
+	public void setBlockData(@NotNull BlockData data)
 	{
-		Assert.assertNotNull(data);
 		this.blockData = data;
 	}
 
--- a/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
@@ -5,6 +5,9 @@ import static org.junit.Assert.fail;
 import java.util.Collection;
 import java.util.List;
 
+import be.seeseemelk.mockbukkit.MockBukkit;
+import be.seeseemelk.mockbukkit.block.data.BlockDataMock;
+import org.bukkit.Bukkit;
 import org.bukkit.Chunk;
 import org.bukkit.FluidCollisionMode;
 import org.bukkit.Location;
@@ -26,6 +29,7 @@ import org.bukkit.util.Vector;
 
 import be.seeseemelk.mockbukkit.UnimplementedOperationException;
 import junit.framework.AssertionFailedError;
+import org.junit.Assert;
 
 public class BlockMock implements org.bukkit.block.Block
 {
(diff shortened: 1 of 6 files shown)


src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
```diff
@@ -363,15 +369,14 @@ public boolean breakNaturally(ItemStack tool)
...
-		throw new UnimplementedOperationException();
+		return blockData;
 	}
 
 	@Override
 	public void setBlockData(BlockData data)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Assert.assertNotNull(data);
```
TheBusyBiscuit: Never assert in the implementations of normal methods. This will make any test fail which may actually intentionally provoke a null argument here.

You should always throw an `IllegalArgumentException` for illegal arguments instead of asserting this early on unless specifically intended.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -50,6 +50,7 @@
 	private String name = "entity";
 	private final Queue<String> messages = new LinkedTransferQueue<>();
 	private final Set<PermissionAttachment> permissionAttachments = new HashSet<>();
+	private Vector velocity;
```
TheBusyBiscuit: This should probably be initialized as 0,0,0 instead of null.

TheBusyBiscuit (changes requested):

src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
seeseemelk: You could also use `Objects.requireNonNull(data)` instead

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -0,0 +1,61 @@
...
+    public @NotNull Material getMaterial()
+    {
+        return type;
+    }
+
+    @Override
+    public @NotNull String getAsString()
+    {
+        // TODO Auto-generated method stub
+        throw new UnsupportedOperationException();
```
seeseemelk: Maybe change these from `UnsupportedOperationException` to `UnimplementedOperationException`. The latter type of exception will cause JUnit to not show any test using the function as having a failure, but rather to be ignored as it uses functionality that cannot be tested yet.

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public @NotNull String getAsString(boolean hideUnspecified)
+    {
+        // TODO Auto-generated method stub
+        throw new UnsupportedOperationException();
```
seeseemelk: Ditto

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public @NotNull BlockData merge(@NotNull BlockData data)
+    {
+        // TODO Auto-generated method stub
+        throw new UnsupportedOperationException();
```
seeseemelk: Ditto

seeseemelk (changes requested):
Looks good aside from a couple of small points

LinoxGH:
I'm sorry for the shallow addition. I'm not so experienced with Java and totally not experienced enough with automated tests. So I just tried to implement what I could for the issue #74 and what I needed in a plugin using this.

seeseemelk (approved):

seeseemelk:
Don't worry about it being to be too shallow. The whole pull request is here so that code can be checked and improved, and I'm honestly really happy every time someone opens a new pull request. (even for the smallest changes :smile: )
So thank you for contributing!

PR #80, opened 2020-06-29
Hopefully fixed publishing publications for good

LinoxGH:
You should most likely review and go through with this


seeseemelk:
Seems to work, thanks!

PR #87, opened 2020-07-20
Added isEmpty and isLiquid to the block class.

ohowe1:


seeseemelk (approved):

src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
```diff
@@ -308,15 +308,13 @@ public int getBlockPower()
...
-		throw new UnimplementedOperationException();
+		return this.getType().equals(Material.AIR);
 	}
 
 	@Override
 	public boolean isLiquid()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return this.getType().equals(Material.LAVA) || this.getType().equals(Material.WATER);
```
TheBusyBiscuit, line 317: Enum comparisons should be done using == instead of .equals() as it is faster and safer since it can do premature exclusions at compile time instead of runtime. Plus you get the benefit of avoiding to run into NullPointerException.

TheBusyBiscuit, line 317: Also you forgot one liquid: `BUBBLE_COLUMN`, water above magma blocks has its own Material even though it looks and behaves like water.

TheBusyBiscuit (commented):
Even though this Pull Request has already been merged, I have found two things which should be adressed to avoid further mistakes.

src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
seeseemelk: Ah, I looked over this. Thanks for noticing!

seeseemelk: Wow, I didn't even know this liquid existed.

seeseemelk:
I opened a new pull request to fix these issues

PR #89, opened 2020-08-02
Added default Potion Effect Types

TheBusyBiscuit:
This can be seen as a follow-up Pull Request to #75.
It turns out that, similar to the Material Tags, the Potion Effect Types need to be registered by the Server itself.
So I decided that it would be best to just register all default Potion Effect Types in the ServerMock class, with a - rather primitive - safeguard to avoid double-registration and IllegalArgumentExceptions.
Without this Pull Request, all Potion Effects would have to be registered manually before being able to use them inside a Unit Test, so this is definitely worth adding in my opinion.

I have also added a rather simple Unit Test for this.


TheBusyBiscuit:
Also, the colors, names, ids and stuff were simply taken from Spigot's default implementation, I just plugged it into a regex and made it spit it out like that. Doing it manually was a bit too tedious for me XD

seeseemelk:
This in completely within the spirit of MockBukkit, so it is definitly worth adding!

PR #94, opened 2020-08-07
Added a simple Item Entity Mock + Player first/last played

TheBusyBiscuit:
A simple Item Entity mock has been added.
It is a pretty small class and only really holds the ItemStack, as it is supposed to.
I have also moved some methods up to the EntityMock class since they belong to the Entity class, not to LivingEntity.

I have also implemented the methods `getFirstPlayed()`,  `getLastPlayed()` and  `hasPlayedBefore()`.
The "lastPlayed" time will be set everytime ServerMock#addPlayer() is invoked. "firstPlayed" will only be set on the first invocation. I have also included a Unit Test to test this behaviour.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -101,6 +101,8 @@ public class PlayerMock extends LivingEntityMock implements Player
 	private Location compassTarget;
 	private Location bedSpawnLocation;
 	private ItemStack cursor = null;
+	private long firstPlayed = 0;
+	private long lastPlayed = 0;
 
 	private final List<AudioExperience> heardSounds = new LinkedList<>();
 
@@ -144,7 +146,7 @@ public class PlayerMock extends LivingEntityMock implements Player
 		int result = super.hashCode();
 		result = prime * result + Objects.hash(attributes, exp, expLevel, expTotal, displayName, gamemode, getHealth(),
 				foodLevel, saturation, inventory, enderChest, inventoryView, getMaxHealth(), online, whitelisted,
-				compassTarget, bedSpawnLocation, cursor);
+				compassTarget, bedSpawnLocation, cursor, firstPlayed, lastPlayed);
 		return result;
 	}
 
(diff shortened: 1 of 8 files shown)


src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -928,4 +928,30 @@ public void testMultiplePotionEffects()
...
+		
+		server.addPlayer(player);
+		long firstPlayed = player.getFirstPlayed();
+		
+		assertTrue(player.hasPlayedBefore());
+		assertTrue(firstPlayed > 0);
+		assertEquals(player.getFirstPlayed(), player.getLastPlayed());
+		
+		// This forces the time to be different by at least 1ms
+		Thread.sleep(1);
```
seeseemelk: I'm personally not a big fan of something like this in a unit test, but I don't see any immediate way to prevent this.

seeseemelk (approved):

seeseemelk:
This looks good, but the test `ServerMockTest#addPlayer_TwoPlayers_SizeIsTwo` is failing for some reason.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -144,7 +146,7 @@ public int hashCode()
 		int result = super.hashCode();
 		result = prime * result + Objects.hash(attributes, exp, expLevel, expTotal, displayName, gamemode, getHealth(),
 				foodLevel, saturation, inventory, enderChest, inventoryView, getMaxHealth(), online, whitelisted,
-				compassTarget, bedSpawnLocation, cursor);
+				compassTarget, bedSpawnLocation, cursor, firstPlayed, lastPlayed);
```
seeseemelk, line 149: The `PlayerMock#equals` method should also test for `firstPlayed` and `lastPlayed`

seeseemelk (changes requested):

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
TheBusyBiscuit: Whoops, my bad.

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
TheBusyBiscuit: Yeah I am not a great fan of this either,
But I couldn't come up with any more elegant solution for this at the time, this was all done in a hurry.

But now after my first coffee, I may have come up with a better solution.

TheBusyBiscuit:
So what I have done now is the following:
Instead of directly invoking `System.currentTimeMillis()`, I added a protected method to get the server time.
This method simply defaults to `System.currentTimeMillis()`.

However in our unit test (where we need the time to actually be different, I simply provided a custom ServerMock implementation which overrides this method to always add 1 on every invocation.
This shouldn't cause any problems or be of a big disruptance since only this single method utilises this method so far.

I think this would be a much better solution than a simple Thread.sleep() which I lazily did yesterday.
Not that it would matter but the tests will now also be at least 1ms faster XD

seeseemelk (approved):

PR #101, opened 2020-08-23
Add pull request template

seeseemelk:


.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+# Checklist
```
TheBusyBiscuit, line 4: I really like the addition of a template here!
But I think you should also add a "# Description (State your changes)" section above that, since that is pretty important for pull request documentation anyway!

It might also be wise to include the points of "Adding Unit Tests" and "Following the existing code format" in that checklist.

seeseemelk: Does the updated description seem better?

TheBusyBiscuit: Looks good to me!

PR #108, opened 2020-09-09
Fixes #107 + Unit Test

### LINKED ISSUE #107: ServerMock#getRecipesFor does not match ItemStack with unequal amount
The [Javadocs](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/Server.html#getRecipesFor(org.bukkit.inventory.ItemStack)) describe the behavior of Server#getRecipesFor as:
`Get a list of all recipes for a given item. The stack size is ignored in comparisons.`

However, ServerMock does not ignore the stack size in the comparison, which makes it hard to test a plugin which relies upon that functionality.

TheBusyBiscuit:
# Description
This Pull Request fixes Issue #107 and adds a Unit Test to verify this behaviour for future versions.

# Checklist
- [x] Unit tests added.


seeseemelk:
#109 fixes the build error this PR has. After it has been merged, update this branch and bump the version number again to `0.5.2`

seeseemelk:
Welp, I just remembered I hadn't actually checked the code of this. Looks good though, the only thing I noticed is that maybe the `while` loop in `removeRecipe` could maybe be a `foreach`, but it seems the previous code was already doing it like that.

TheBusyBiscuit:
`removeRecipe` will remove recipes from our collection though, hence why an Iterator would be a better choice here since you can't remove items inside a `foreach` loop without triggering a `ConcurrentModificationException`.

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -707,10 +708,54 @@ public boolean addRecipe(Recipe recipe)
...
+	public List<Recipe> getRecipesFor(@NotNull ItemStack item)
 	{
 		assertMainThread();
-		return recipes.stream().filter(recipe -> recipe.getResult().equals(result)).collect(Collectors.toList());
+
+		return recipes.stream().filter(recipe -> {
+			ItemStack result = recipe.getResult();
+			// Amount is explicitly ignored here
+			return result.getType() == item.getType() && result.getItemMeta().equals(item.getItemMeta());
+		}).collect(Collectors.toList());
```
clmcdonald, line 719: I know this PR is already closed but I just wanted to recommend that you replace these lines with something like this.
I haven't tested it, but I think you should use ItemStack#isSimilar instead of doing your own comparison
```
return recipes.stream()
	.map(Recipe::getResult)
	.filter(item::isSimilar)
	.collect(Collectors.toList());
```
EDIT: supposed to return Recipe not ItemStack...but you get the idea

TheBusyBiscuit: Oh right, I forgot that .isSimilar() ignores the amount too

seeseemelk: This would return a `List<ItemStack` though instead of a `List<Recipe>`.
Perhaps
```java
return recipes.stream()
    .filter(item -> item.isSimilar(item.getResult()))
    .collect(Collectors.toList());
```

clmcdonald:
```
return recipes.stream()
    .filter(recipe -> item.isSimilar(recipe.getResult()))
    .collect(Collectors.toList());
```

TheBusyBiscuit: This would of course result in a NPE if item is null, is this inline with Spigot behaviour too?

TheBusyBiscuit: Oh nevermind, it is marked as NonNull anyway, looks good!

PR #117, opened 2020-09-30
Added bossbar

ohowe1:
# Description
- Added the `BossBar` class
- Added the `KeyedBossBar` class
- Added the methods `createBossBar`, `getBossBars`, `getBossBar`, and `removeBossBar` to `ServerMock`
- Added tests for the `Bossbar` and `KeyedBossBar` classes.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -75,6 +75,8 @@ import org.bukkit.potion.PotionEffectType;
 import org.bukkit.util.CachedServerIcon;
 import org.jetbrains.annotations.NotNull;
 
+import be.seeseemelk.mockbukkit.boss.BossBarMock;
+import be.seeseemelk.mockbukkit.boss.KeyedBossBarMock;
 import be.seeseemelk.mockbukkit.command.CommandResult;
 import be.seeseemelk.mockbukkit.command.ConsoleCommandSenderMock;
 import be.seeseemelk.mockbukkit.command.MessageTarget;
@@ -110,6 +112,7 @@ public class ServerMock implements Server
 	private final Set<EntityMock> entities = new HashSet<>();
 	private final List<World> worlds = new ArrayList<>();
 	private final List<Recipe> recipes = new LinkedList<>();
+	private final Map<NamespacedKey, KeyedBossBarMock> bossBars = new HashMap<>();
 	private final ItemFactory factory = new ItemFactoryMock();
 	private final PlayerMockFactory playerFactory = new PlayerMockFactory(this);
 	private final PluginManagerMock pluginManager = new PluginManagerMock(this);
@@ -1214,8 +1217,8 @@ public class ServerMock implements Server
 	@Override
 	public BossBar createBossBar(String title, BarColor color, BarStyle style, BarFlag... flags)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		BossBar bar = new BossBarMock(title, color, style, flags);
+		return bar;
 	}
 
 	@Override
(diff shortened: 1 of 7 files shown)


src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -23,6 +23,7 @@
 import org.bukkit.Tag;
 import org.bukkit.command.Command;
 import org.bukkit.entity.Player;
+import org.bukkit.entity.Horse.Style;
```
TheBusyBiscuit: Where did this import suddenly come from?

src/main/java/be/seeseemelk/mockbukkit/boss/BossBarMock.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Override
+    public double getProgress()
+    {
+        return progress;
+    }
+
+    @Override
+    public void addPlayer(Player player)
```
TheBusyBiscuit: player is marked as `@NotNull` in the Spigot API, please continue using this annotation here and prevent null from being passed.

src/main/java/be/seeseemelk/mockbukkit/boss/BossBarMock.java
```diff
@@ -0,0 +1,153 @@
...
+    {
+        this.players.add(player);
+    }
+
+    @Override
+    public void removePlayer(Player player)
```
TheBusyBiscuit: player is marked as `@NotNull` in the Spigot API, please continue using this annotation here and prevent null from being passed.

src/main/java/be/seeseemelk/mockbukkit/boss/BossBarMock.java
TheBusyBiscuit: The original Spigot implement throws an IllegalArgumentException for null parameters, you can (and should) do the same, you can just use `Validate.notNull(player, message)` for this. It would be awesome to add a unit test for this case too

src/main/java/be/seeseemelk/mockbukkit/boss/BossBarMock.java
TheBusyBiscuit: The original Spigot implement throws an IllegalArgumentException for null parameters, you can (and should) do the same, you can just use `Validate.notNull(player, message)` for this. It would be awesome to add a unit test for this case too

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -1393,29 +1396,30 @@ public ItemStack createExplorerMap(World world, Location location, StructureType
...
+		bossBars.put(key, bar);
+		return bar;
 	}
 
 	@Override
 	public Iterator<KeyedBossBar> getBossBars()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return ((List) new ArrayList<KeyedBossBarMock>(bossBars.values())).iterator();
```
TheBusyBiscuit: This isn't an ideal solution, I'd go for something like this:
```suggestion
		return bossBars.values().stream().map(bossbar -> (KeyedBossBar) bossbar).iterator();
```

I haven't tested it but this should be a bit cleaner as it avoids unsafe type casting and generic types.

TheBusyBiscuit (changes requested):
Haven't looked at the Unit Tests but there are a few minor things I have noticed :)
Otherwise looking really good, you could also add some documentation to your methods or classes but I think it's good as is.

src/main/java/be/seeseemelk/mockbukkit/boss/BossBarMock.java
ohowe1: I was not doing that because some of the methods with that annotation in the spigot did not have it in MockBukkit. 
Thanks for the heads up and will fix.

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
ohowe1: Ok. Good to know thanks!

seeseemelk (approved):

seeseemelk (approved):

PR #121, opened 2020-10-05
[Feature] new Get or Create Method for Mock Server

nikgil:
# Description
Developed a `getOrCreateMock` method for ServerMock.

## Justification behind reason:
on the consumer-side reduces amount of boilerplate needed when creating and removing server instances for unit tests (avoid having to check if server is mocked every time). Alternative would be to make `setServerInstanceToNull` public as trying to unmock every time will cause a null pointer to be thrown in the finally block if no instance exists.

# Checklist
- [x] Unit tests added.


TheBusyBiscuit (approved):

seeseemelk:
We should also probably fix the `NullPointerException` that is thrown by `unmock`.

PR #127, opened 2020-10-09
Added basic functionality of Block#breakNaturally

### LINKED ISSUE #126: Implement Block#breakNaturally functionally
I have a plugin that requires breaking the block (with the plugin) and dropping the drop of the item without triggering the block break event, but it seems to not be implemented, preventing parts of the plugin from being unit tested. I have tried a solution without changing the source code (extended classes), but it took too much effort (of figuring it out), and it still didn't work.

A PR that adds functionality should:

- Simulate a block break without triggering the block break event
- Drop the item that would drop normally from "mining" the block (for Block#breakNaturally, may also require impl of Block#getDrops)

JacksonChen666:
# Description
Added basic functionality of Block#breakNaturally (breaks block, no event).
This does not include block drops that can be picked up by the player.
This PR fixes #126

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
@@ -344,15 +344,18 @@ public class BlockMock implements Block
 	@Override
 	public boolean breakNaturally()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		if (this.isEmpty())
+		{
+			return false;
+		}
+		this.setType(Material.AIR);
+		return true;
 	}
 
 	@Override
 	public boolean breakNaturally(ItemStack tool)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return this.breakNaturally();
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
```diff
@@ -344,15 +344,18 @@ public PistonMoveReaction getPistonMoveReaction()
 	@Override
 	public boolean breakNaturally()
 	{
```
nikgil, line 346: I'm seeing in craftblock that it does actually drop items. Does it call the `BlockDropItemEvent` in that case? But also would the item drop need to be mocked?

JacksonChen666: The PR doesn't simulate either the `BlockDropItemEvent` event (missing `Block#getDrop`) and the item drop, and item drops would also have to be mocked (if the player can pick up items from the world).

seeseemelk: Adding that functionality wouldn't be too hard I think. It would require an `ItemMock` entity which gets added to the list of entities in a world, and implement some simple methods. However, it would probably be best to implement this in a separate PR.

seeseemelk (approved):

Issue #113, opened 2020-09-12
Implement additional inventory mocks


seeseemelk:
We're currently only supporting three inventory types, let's implement the rest!

A PR should add mock implementaties for any of these inventory types (the ones with a check mark have already been implemented):
 - [x] Anvil
 - [x] Barrel
 - [x] Beacon
 - [x] Blast Furnace
 - [x] Brewing
 - [x] Cartography
 - [X] Chest
 - [x] Crafting
 - [x] Creative
 - [x] Dispenser
 - [x] Dropper
 - [x] Enchanting
 - [x] Ender Chest
 - [x] Furnace
 - [x] Grindstone
 - [x] Hopper
 - [x] Lectern
 - [x] Loom
 - [x] Merchant
 - [X] Player
 - [x] Shulker Box
 - [x] Smithing
 - [x] Smoker
 - [x] Stonecutter
 - [x] Workbench

Note that some of these are not creatable (check `org.bukkit.event.inventory.InventoryType`), so `ServerMock#createInventory` should throw an error. Many of the ones that are creatable can most likely be easily implemented using the `SimpleInventoryMock` class.

TheBusyBiscuit:
:eyes: I am working on Dispenser / Dropper mocks!
https://github.com/TheBusyBiscuit/MockBukkit/commit/1d8dc334fdfcd3203a9a8cb06938fcce7b1e148a

seeseemelk:
I increased the scope of it a wee bit because I thought it would maybe be insane to have 23 separate tickets just for inventories. But the issue can be solved by multiple people and multiple PRs.

TheBusyBiscuit:
Ah okay, that's perfect!<br>
*I also updated my comment to emphasize what exact part I am working on*

TheBusyBiscuit:
@seeseemelk The checklist needs to be updated.

seeseemelk:
Which once have been implemented?

seeseemelk:
Ah, it seems to be Barrel, Dispenser, Dropper, and Shulker Box

svew:
@seeseemelk Looks like lectern has been added as well

thelooter:
This is officially done!

PR #128, opened 2020-10-10
Lectern Inventory

nikgil:
# Relevant Issue: #113 

# Description
- Add Lectern Inventory
- Add constraints to inventory creation as present in gradle

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/LecternMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/LecternMock.java
@@ -10,11 +10,13 @@ import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.meta.BookMeta;
 import org.jetbrains.annotations.NotNull;
 
-public class LecternMock extends ContainerMock implements Lectern {
+public class LecternMock extends ContainerMock implements Lectern
+{
 
     private int currentPage;
 
-    public LecternMock(@NotNull Material material) {
+    public LecternMock(@NotNull Material material)
+    {
         super(material);
     }
 
@@ -29,22 +31,26 @@ public class LecternMock extends ContainerMock implements Lectern {
     }
 
     @Override
-    protected InventoryMock createInventory() {
+    protected InventoryMock createInventory()
+    {
         return new LecternInventoryMock(this);
     }
 
     @Override
-    public BlockState getSnapshot() {
+    public BlockState getSnapshot()
+    {
         return new LecternMock(this);
     }
 
     @Override
-    public int getPage() {
+    public int getPage()
+    {
         return this.currentPage;
     }
 
     @Override
-    public void setPage(int page) {
+    public void setPage(int page)
+    {
         ItemStack book = this.getInventory().getItem(0);
 
         int maxPages = book != null && getBookMeta(book) != null
@@ -54,12 +60,15 @@ public class LecternMock extends ContainerMock implements Lectern {
         this.currentPage = Math.min(Math.max(0, page), maxPages - 1);
     }
 
-    private BookMeta getBookMeta(ItemStack book) {
-        if(book == null) {
+    private BookMeta getBookMeta(ItemStack book)
+    {
+        if (book == null)
+        {
             return null;
         }
 
-        if(book.getType() != Material.WRITABLE_BOOK && book.getType() != Material.WRITTEN_BOOK) {
+        if (book.getType() != Material.WRITABLE_BOOK && book.getType() != Material.WRITTEN_BOOK)
+        {
             return null;
         }
 
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/LecternMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/LecternMock.java
@@ -1,0 +1,62 @@
+package be.seeseemelk.mockbukkit.block.state;
+
+import be.seeseemelk.mockbukkit.inventory.InventoryMock;
+import be.seeseemelk.mockbukkit.inventory.LecternInventoryMock;
+import org.bukkit.Material;
+import org.bukkit.block.BlockState;
+import org.bukkit.block.Lectern;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.BookMeta;
+import org.jetbrains.annotations.NotNull;
+
+public class LecternMock extends ContainerMock implements Lectern {
+
+    private int currentPage;
+
+    public LecternMock(@NotNull Material material) {
+        super(material);
+    }
+
+    protected LecternMock(@NotNull LecternMock state)
+    {
+        super(state);
+    }
+
+    @Override
+    protected InventoryMock createInventory() {
+        return new LecternInventoryMock(this);
+    }
+
+    @Override
+    public BlockState getSnapshot() {
+        return new LecternMock(this);
+    }
+
+    @Override
+    public int getPage() {
+        return this.currentPage;
+    }
+
+    @Override
+    public void setPage(int page) {
+        ItemStack book = this.getInventory().getItem(0);
+
+        int maxPages = book != null && getBookMeta(book) != null
+                ? getBookMeta(book).getPageCount()
+                : 1;
+
+        this.currentPage = Math.min(Math.max(0, page), maxPages - 1);
+    }
+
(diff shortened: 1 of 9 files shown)


src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
```diff
@@ -29,12 +29,24 @@
 
 	public InventoryMock(InventoryHolder holder, int size, InventoryType type)
 	{
+		assertTrue(
+				"Size for custom inventory must be a multiple of 9 between 9 and 54 slots (got " + size + ")",
+				9 <= size && size <= 54 && size % 9 == 0);
```
TheBusyBiscuit: I think a Validate would be more appropriate here.

Especially because this is what the Bukkit APi does:
`Validate.isTrue(9 <= size && size <= 54 && size % 9 == 0, "Size for custom inventory must be a multiple of 9 between 9 and 54 slots (got " + size + ")");`

Almost an exact copy.

```suggestion
		Validate.isTrue(9 <= size && size <= 54 && size % 9 == 0,
				"Size for custom inventory must be a multiple of 9 between 9 and 54 slots (got " + size + ")");
```

src/test/java/be/seeseemelk/mockbukkit/inventory/InventoryMockTest.java
```diff
@@ -45,6 +45,41 @@ public void constructor_SetsSize()
...
+
+		try
+		{
+			new SimpleInventoryMock(null, 63, InventoryType.CHEST);
+		} catch(AssertionError e) {
+			errorCount++;
+		}
+
+		assertEquals(3, errorCount);
+	}
```
TheBusyBiscuit: ugh, that is really not a good practice to go about things.
Especially because you lose the information about **where** the error happened which is very much not the purpose of a Unit Test.

To test for an exception you would generally make a seperate Test and then use `@Test(expected = SomeException.class)`.

TheBusyBiscuit (changes requested):

src/main/java/be/seeseemelk/mockbukkit/block/state/LecternMock.java
```diff
@@ -0,0 +1,62 @@
...
+        return this.currentPage;
+    }
+
+    @Override
+    public void setPage(int page) {
+        ItemStack book = this.getInventory().getItem(0);
+
+        int maxPages = book != null && getBookMeta(book) != null
+                ? getBookMeta(book).getPageCount()
+                : 1;
```
TheBusyBiscuit, line 58: This could maybe be refactored or even put into a dedicated method, but not really a necessity.

TheBusyBiscuit (commented):
One thing that could still be added would be Unit tests for the actual Lectern BlockState.

And more importantly the Lectern should be added to the switch/case in BlockStateMock.
But otherwise this is looking pretty good!

nikgil:
Couple of notes:
Issue #113 should be updated a bit. Since there are inventories that are in code but not updated in the issue (dropper and barrel off the top of my head).

Didn't notice this last time but code formatting is inconsistent, some styling is K&R, some is Allman and some do Java camel case naming conventions while others do snake case.

Some of the older tests don't really follow a Given-When-Then flow which falls in line with what was mentioned above, not testing one characteristic at one time.

Could do with a formatting Issue somewhere down the line.

src/main/java/be/seeseemelk/mockbukkit/block/state/LecternMock.java
nikgil: depends on how hard to read you feel that is. Method is <10 lines which is what I aim for, but only problem with this I can see is potentially readability issues.

TheBusyBiscuit: The ternary operator is generally not considered the most readable asset compared to if/else structures.
Again, I said that I don't see it as a necessity, but it's generally good practice to rely on if/else structures or to make a dedicated method for this.
Just a general comment on it, it would also be fine as is.

seeseemelk: Be especially careful with ternary operators when using them with more complex expressions. Many people (including me) don't know of the top of their head which operations take precedence. (E.g.: is `A && B ? C : D` the same as `(A && B) ? C : D` or is it the same as `A && (B ? C : D)`. Or what about `A ? B : C + D`, is that `(A ? B : C) + D` or `A ? B : (C + D)`).

For simple situations I do use it myself though.

src/main/java/be/seeseemelk/mockbukkit/block/state/LecternMock.java
```diff
@@ -0,0 +1,68 @@
...
+            return null;
+        }
+
+        if(book.getType() != Material.WRITABLE_BOOK && book.getType() != Material.WRITTEN_BOOK) {
+            return null;
+        }
+
+        return (BookMeta) book.getItemMeta();
+    }
+}
```
seeseemelk: Several parts of this class don't really follow the existing code style guidelines. The project uses [Allman style](https://en.wikipedia.org/wiki/Indentation_style#Allman_style), which means that all opening braces are on their own line, and you always have a space between an `if` and it's opening bracket. (E.g: `if (...)` instead of `if(...)`)

src/main/java/be/seeseemelk/mockbukkit/inventory/LecternInventoryMock.java
```diff
@@ -0,0 +1,13 @@
...
+import org.bukkit.inventory.InventoryHolder;
+
+/**
+ * This {@link InventoryMock} mocks an {@link org.bukkit.block.Lectern} but pretty much behaves like any small chest.
+ */
+public class LecternInventoryMock extends InventoryMock {
+    public LecternInventoryMock(InventoryHolder holder) {
+        super(holder, InventoryType.LECTERN);
+    }
+}
```
seeseemelk: Same comment about code style as in `LecternMock`.

src/test/java/be/seeseemelk/mockbukkit/block/state/LecternMockTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+        //When
+        lectern.setPage(2);
+
+        //Then
+        assertEquals(2, lectern.getPage());
+    }
+
+    @Test
+    public void testSetPageInvalid() {
```
seeseemelk: Same comment about code style as for `LecternMock`.

seeseemelk (changes requested):
Functionally it looks really good. I did leave some comment about code style on at a couple of places. If you could fix those I'll merge it! :rocket:

seeseemelk:
> Couple of notes:
> Issue #113 should be updated a bit. Since there are inventories that are in code but not updated in the issue (dropper and barrel off the top of my head).
> 
> Didn't notice this last time but code formatting is inconsistent, some styling is K&R, some is Allman and some do Java camel case naming conventions while others do snake case.
> 
> Some of the older tests don't really follow a Given-When-Then flow which falls in line with what was mentioned above, not testing one characteristic at one time.
> 
> Could do with a formatting Issue somewhere down the line.

I should have read this before doing the code review :facepalm:. You are right, the code formatting is quite inconsistent. The reason for this is early in the project when I started getting the very first PRs, I was so excited that people were actually using MockBukkit that I didn't really care about them using different coding styles. I was afraid I'd scare people off if I'd be too pedantic about that.

However, now that I'm getting more and more PRs I'm enforcing a consistent code style a bit more. You're always welcome to clean up some of the code if/when you come across it, it would be appreciated!

nikgil:
@seeseemelk will do. won't do any clean up (other than the lectern stuff) here just for sanity sake and not having a bunch of extra files everywhere.

src/main/java/be/seeseemelk/mockbukkit/block/state/LecternMock.java
nikgil: I think here it should be trivial enough. Would never compound them or anything like that but I see the point.

seeseemelk:
That's fine, it's probably best to clean up those issues while modifying those files for other issues anyway. That way there is minimal clutter while still have some net improvement.

seeseemelk (approved):

PR #132, opened 2020-10-13
reformat and create code style editorconfig for IntelliJ

nikgil:
# Description
Reformatted all code to be in allman style and created an editorconfig to be imported by IntelliJ IDE to automatically format code this way.

# Checklist
- [x] Unit tests added (although not applicable).


src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
```diff
@@ -5,49 +5,48 @@
  */
 public class Coordinate
 {
-	public int x;
-	public int y;
-	public int z;
+    public int x;
```
nikgil, line 8: This is a recurring theme but basically code was done in tabs not spaces which isn't normalized. config has 4 spaces indent I believe.

src/main/java/be/seeseemelk/mockbukkit/block/state/ContainerMock.java
```diff
@@ -8,92 +9,88 @@
...
+        super(material);
+        this.inventory = createInventory();
+    }
 
-	public ContainerMock(@NotNull Material material)
-	{
-		super(material);
-		this.inventory = createInventory();
-	}
+    protected ContainerMock(@NotNull Block block)
```
nikgil, line 30: TODO: change back to public

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
```diff
@@ -22,273 +13,277 @@
...
+    }
+
+    @After
+    public void tearDown()
+    {
+        MockBukkit.unmock();
+    }
+
+    @Test
+    public void getBlockAt_StandardWorld_DefaultBlocks()
```
nikgil, line 38: Still not sure how to approach these snake + camelCase scenarios

allman-editorconfig
```diff
@@ -0,0 +1,1330 @@
+[*]
```
nikgil, line 1: @seeseemelk I would recommend taking a play through with this to see if everything fits the required style

JacksonChen666 (changes requested):
Using the filename `allman-editorconfig` would not work for IntelliJ IDEA (at least, for me), but renaming the file to `.editorconfig` would work (anything before the . will be rejected by IntelliJ), and the editor would follow the formatting specified in the file instead of the user's format.

nikgil:
@JacksonChen666 you would need to manually import it through settings, reasonably simple but would need manual intervention. Main reason that was done is because .editorconfig was on the gitignore, so rather than remove and push I want to see the reasoning it was on the list first.

JacksonChen666:
It was ignored because it matched the pattern `.*` with a comment saying, "Ignore hidden files." Adding `!.editorconfig` to `.gitignore` and renaming `allmen-editorconfig` to `.editorconfig` should make IntelliJ follow it automatically and allow the file in the repo.

nikgil:
Fair enough. I'll change that once the other changes get their blessing

src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
seeseemelk: Oof. I'm personally more of a fan of tabs rather than spaces. I know some people who always set their IDE to render a tab as 2 space, but that makes my eyes bleed. So it's quite nice that we can set it to different settings.
Then again, I'm not sure if I want to start a tabs vs spaces discussion. What do you think @TheBusyBiscuit?

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
seeseemelk: This is a certain method of naming your unit tests. It's essentially `method_scenario_result()`, but most pull requests don't follow this and I don't use it either anymore. I don't work in Java all that much anymore, so I don't have any specific preference on how these should be treated.

src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
TheBusyBiscuit: There are good arguments for both.

With **spaces** you get universal control of how your code is indented, no matter the platform or settings, which is really handy.
The space character also has the benefit of being easier to use on mobile platforms, albeit not the norm, sometimes I do find myself pushing a quick little fix from my phone.
Spaces simply provide a very huge level of consistency, which is very important for general code style discussions.

With **tabs** however you give every individual the option to set the level of indentation themselves, you hand over this control.
From a memory/data standpoint this will also be slightly more efficient as a tab character is much more lightweight to store than say four individual space characters, even if the compiled result may not care about this. (And easier to read of course since you can just set it to render as two spaces for large files)
From a historical standpoint, the tab character was also designed specifically for indentation, whereas the space character is meant to be used for word seperations.

It really comes down to the preference of the project owner.
I personally prefer spaces, simply as I think they are the most suitable for big open-source projects. They are universal and your code will look the same no matter where you view it, many popular standards like the Google Java Convention have a spaces-only policy too.
But that is just my personal opinion, I don't have anything against following the standard a project owner defines, as there are valid arguments for both variants.

It really doesn't *necessarily* come down to what you choose, as long as it is consistent. Because mixing the two characters creates really bad-looking results most of the time.

seeseemelk: Let's make it 4 spaces then. Half of the incoming pull requests use it anyway.

Narimm:
Why use `.editorconfig` - the accepted standard for standardised java file formatting would be checkstyle.xml.

Narimm:
The goal here should not be to automatically import a file which modifies someones ide...that is sort of frowned upon.  But to stop them or automatically check code that they might PR...Using Codacy would be a good example of how checkstyle can be automatically checked and comments applied to a PR.

PR #135, opened 2020-11-04
Added missing methods for inventory contains

JacksonChen666:
# Description
This PR adds "missing" methods for inventory contains methods.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,5 @@
 gradle.ext.apiVersion = '1.16'
 gradle.ext.apiVersionFull = '1.16.3-R0.1-SNAPSHOT'
-gradle.ext.version = '0.13.1'
+gradle.ext.version = '0.14.0'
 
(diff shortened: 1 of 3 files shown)


seeseemelk:
I merged #134, so you will have to bump the version number up to 0.15.0.
Other than that, it looks promising so far.

JacksonChen666:
It seems like there is `InventoryMock#assertContainsAtLeast` which is used for assertions, but for some reason there isn't `InventoryMock#containsAtLeast`? Very odd stuff.

seeseemelk:
> It seems like there is `InventoryMock#assertContainsAtLeast` which is used for assertions, but for some reason there isn't `InventoryMock#containsAtLeast`? Very odd stuff.

There are indeed some inconsistency in the library. At the beginning of the project I just added functions as I needed them. Feel free to split the function into two!

seeseemelk (approved):

PR #138, opened 2020-11-20
Update Plugin Event mananger to support Priority as well as improved Event handling

Narimm:
### Description

Improves Listener registration handling so that priority is properly handled and the listeners are stored properly.  Cancellation is functional 

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
@@ -386,23 +386,27 @@ public class PluginManagerMock implements PluginManager
 		events.add(event);
 		HandlerList handlers = event.getHandlers();
 		RegisteredListener[] listeners = handlers.getRegisteredListeners();
-		for(RegisteredListener l : listeners) {
+		for(RegisteredListener l : listeners)
+		{
 			callRegisteredListener(l,event);
 		}
 	}
 
 	private void callRegisteredListener(RegisteredListener registration, Event event){
-		if (!registration.getPlugin().isEnabled()) {
+		if (!registration.getPlugin().isEnabled())
+		{
 			return;
 		}
-		try {
+		try
+		{
 			registration.callEvent(event);
-		} catch (AuthorNagException ex) {
+		}
+		catch (AuthorNagException ex)
+		{
 			Plugin plugin = registration.getPlugin();
-
-			if (plugin.isNaggable()) {
+			if (plugin.isNaggable())
+			{
 				plugin.setNaggable(false);
-
 				server.getLogger().log(Level.SEVERE, String.format(
 						"Nag author(s): '%s' of '%s' about the following: %s",
 						plugin.getDescription().getAuthors(),
@@ -410,7 +414,9 @@ public class PluginManagerMock implements PluginManager
 						ex.getMessage()
 				));
 			}
-		} catch (Throwable ex) {
+		}
+		catch (Throwable ex)
+		{
 			String msg = "Could not pass event " + event.getEventName() + " to " + registration.getPlugin().getDescription().getFullName();
 			server.getLogger().log(Level.SEVERE, msg, ex);
 		}
@@ -576,7 +582,8 @@ public class PluginManagerMock implements PluginManager
 	@Override
 	public void registerEvents(Listener listener, Plugin plugin)
 	{
-		if (!plugin.isEnabled()) {
+		if (!plugin.isEnabled())
+		{
 			throw new IllegalPluginAccessException("Plugin attempted to register " + listener + " while not enabled");
 		}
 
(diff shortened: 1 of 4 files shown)


src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -5,49 +5,29 @@
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
+import java.lang.reflect.*;
```
TheBusyBiscuit: Please do **not** use wildcard imports, it clutters the local namespace and is subject to breaking changes when the package gets ever updated.

This doesn't necessarily apply for java's internal packages but it is generally a good practice to always avoid wildcard imports whenever you can.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -5,49 +5,29 @@
...
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
TheBusyBiscuit: Please do **not** use wildcard imports, it clutters the local namespace and is subject to breaking changes when the package gets ever updated.

This doesn't necessarily apply for java's internal packages but it is generally a good practice to always avoid wildcard imports whenever you can.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -5,49 +5,29 @@
...
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang.Validate;
 import org.bukkit.command.PluginCommand;
 import org.bukkit.command.PluginCommandUtils;
 import org.bukkit.configuration.file.YamlConfiguration;
-import org.bukkit.event.Event;
-import org.bukkit.event.EventHandler;
-import org.bukkit.event.EventPriority;
-import org.bukkit.event.Listener;
+import org.bukkit.event.*;
```
TheBusyBiscuit: Please do **not** use wildcard imports, it clutters the local namespace and is subject to breaking changes when the package gets ever updated.

It is generally a good practice to always avoid wildcard imports whenever you can.
But this applies **especially** here for the Bukkit package, importing the entire package will give Bukkit/Spigot the ability to pollute this local namespace whenever there is a new Event class added and that does happen quite often.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -5,49 +5,29 @@
...
 import org.bukkit.permissions.PermissionDefault;
-import org.bukkit.plugin.EventExecutor;
-import org.bukkit.plugin.InvalidDescriptionException;
-import org.bukkit.plugin.InvalidPluginException;
-import org.bukkit.plugin.Plugin;
-import org.bukkit.plugin.PluginDescriptionFile;
-import org.bukkit.plugin.PluginLoader;
-import org.bukkit.plugin.PluginManager;
-import org.bukkit.plugin.UnknownDependencyException;
+import org.bukkit.plugin.*;
```
TheBusyBiscuit: Please do **not** use wildcard imports, it clutters the local namespace and is subject to breaking changes when the package gets ever updated.

It is generally a good practice to always avoid wildcard imports whenever you can.
But this applies **especially** here for the Bukkit package, importing the entire package will give Bukkit/Spigot the ability to pollute this local namespace.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -380,32 +359,38 @@ private PluginDescriptionFile findPluginDescription(Class<? extends JavaPlugin>
...
-		{
-			for (ListenerEntry entry : listeners)
-			{
-				if (entry.isCompatibleFor(event))
-				{
-					entry.invokeUnsafe(event);
-				}
+		HandlerList handlers = event.getHandlers();
+		RegisteredListener[] listeners = handlers.getRegisteredListeners();
+		Arrays.stream(listeners).forEach(l ->callRegisteredListener(l,event));
```
TheBusyBiscuit: Try to avoid Streams and inline `.forEach` whenever you can.
They are significantly slower and much less optimized by the JVM than a standard for-loop.
We are talking micro-optimizations here but since the Bukkit API is heavily centered around the Event contract, this method will be a hot point of contact and should be handled with care.
So if you wanna reduce the load of this method, you should consider refactoring it back to a normal for-loop.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -380,32 +359,38 @@ private PluginDescriptionFile findPluginDescription(Class<? extends JavaPlugin>
...
+			if (plugin.isNaggable()) {
+				plugin.setNaggable(false);
+
+				server.getLogger().log(Level.SEVERE, String.format(
+						"Nag author(s): '%s' of '%s' about the following: %s",
+						plugin.getDescription().getAuthors(),
+						plugin.getDescription().getFullName(),
+						ex.getMessage()
+				));
 			}
```
TheBusyBiscuit: I would suggest making a `nag(Plugin plugin, String reason)` method here.
It seems like this may come in useful at another time too.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -560,20 +545,62 @@ public void clearEvents()
...
+		}
+	}
+
 	@Override
 	public void registerEvent(Class<? extends Event> event, Listener listener, EventPriority priority,
 			EventExecutor executor, Plugin plugin)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		registerEvent(event,listener,priority,executor,plugin,false);
```
TheBusyBiscuit: Formatting. Please be consistent.

```suggestion
		registerEvent(event, listener, priority, executor, plugin, false);
```

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -719,8 +746,7 @@ public void unsubscribeFromDefaultPerms(boolean op, Permissible permissible)
 	@Override
 	public boolean useTimings()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return false;
```
TheBusyBiscuit, line 798: On what grounds do you return false here?
If there is a valid reason for returning a constant here then please explain so with a comment.

TheBusyBiscuit (changes requested):
Lots of changes I would like to suggest/request

In addition to this, please follow the existing bracket style (Allman) of the project and try to be consistent with the formatting.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
Narimm: Timings refer to the SpigotTimings/ Papers Timing that revolves around tick timers for reporting lag.  Its should be false in a testing environment ..because 

1. Timings are pointless in a testing environ
2.  Lag cant be used consistently anyway in a test server.

 However, this method is called internally in the JavaPluginLoader that Mockbukkit uses and causes any test that uses that call to fail.

TheBusyBiscuit: I agree but like I said in my review, things like that should be explained in a comment.
Good docs and comments on why something is there help make it clear for future contributors to understand why it's there.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
Narimm: TO be honest this code is pretty much C+P'd from the Spigot source, tweaked to support a testing environ.  Again nagging can probably just be removed for testing purposes.  I mean for CI purposes logging is a waste of time.

seeseemelk (approved):

seeseemelk (approved):

PR #142, opened 2020-12-06
Implement HelpMap

NeumimTo:
# Description
Implementation of HelpMap

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/help/HelpMapMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/help/HelpMapMock.java
@@ -8,7 +8,11 @@ import org.bukkit.help.HelpTopic;
 import org.bukkit.help.HelpTopicComparator;
 import org.bukkit.help.HelpTopicFactory;
 
-import java.util.*;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.List;
 
 public class HelpMapMock implements HelpMap
 {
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/help/HelpMapMock.java
```diff
@@ -0,0 +1,67 @@
...
+
+import be.seeseemelk.mockbukkit.UnimplementedOperationException;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandExecutor;
+import org.bukkit.help.HelpMap;
+import org.bukkit.help.HelpTopic;
+import org.bukkit.help.HelpTopicComparator;
+import org.bukkit.help.HelpTopicFactory;
+
+import java.util.*;
```
TheBusyBiscuit: Please do not use wildcard imports, there is absolutely no need to import the entire java.util library.
It serves no benefit here other than polluting this namespace.

TheBusyBiscuit (changes requested):
There is one thing that should definitely be addressed.
Other than that this is looking pretty good!

*Can't really say much about it, I haven't ever used the HelpMap before actually*

NeumimTo:
> There is one thing that should definitely be addressed.
> Other than that this is looking pretty good!
> 
> _Can't really say much about it, I haven't ever used the HelpMap before actually_

Me neither, but i need to test a plugin that relies on ACF, which requires it under the hood somehow.

seeseemelk (approved):

PR #147, opened 2021-01-15
Skip artifact signing on forks

NeumimTo:
# Description
Skip artifact signing on forked repositories

This skips signing artifacts, unless the git origin url contains  seeseemelk

Makes it possible for jitpack to build forked repositories

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,10 @@ artifacts {
     archives javadocJar, sourcesJar
 }
 
+def isFork = isFork()
+
 signing {
+    required { !isFork }
     sign publishing.publications
 }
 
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
Seems like you based your branch off of PR #146.
I reckon it would be best to keep the two branches seperate, so can you re-base your branch off of v1.16?

NeumimTo:
Just merge #146  first, does not really matter

seeseemelk (approved):
After merging the version number was no longer updated, but given the nature of the PR the version number is not really necessary. I don't think there is a need to publish this version  separately on maven central.

PR #148, opened 2021-01-16
Added equals method to BlockStateMock. Fixed player hashCode method.

Lauriethefish:
# Description
- Added an equals method to BlockStateMock to make comparing it in unit tests work.
- Changed the player hashCode method to just hash the unique ID, as CraftBukkit does it. This is important since otherwise those storing players in a map won't work, and will just return null if _anything_ about the player changes. (the exact reason I forked this actually)
If either of these are intended, then I'm sorry.

# Checklist
- [Done] Version number updated in `settings.gradle`.
- [Done] Unit tests added.
- [Done] Code follows existing code format.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
@@ -293,4 +293,17 @@ public class BlockStateMock implements BlockState, Cloneable
 			return new BlockStateMock(block);
 		}
 	}
+
+	@Override
+	public boolean equals(Object other) {
+		if(other == null) {return false;}
+		if(!(other instanceof BlockState)) {return false;}
+
+		BlockState otherState = (BlockState) other;
+		if(!(otherState.getType() == getType())) {return false;}
+		if(otherState.isPlaced() != isPlaced()) {return false;}
+		if(!isPlaced()) {return true;}
+
+		return otherState.getBlock().getLocation().equals(getBlock().getLocation());
+	}
 }
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
@@ -296,20 +296,37 @@ public class BlockStateMock implements BlockState, Cloneable
 	}
 
 	@Override
-	public boolean equals(Object other) {
-		if(other == null) {return false;}
-		if(!(other instanceof BlockState)) {return false;}
+	public boolean equals(Object other)
+	{
+		if (other == null)
+		{
+			return false;
+		}
+		if (!(other instanceof BlockState))
+		{
+			return false;
+		}
 
 		BlockState otherState = (BlockState) other;
-		if(otherState.getType() != getType()) {return false;}
-		if(otherState.isPlaced() != isPlaced()) {return false;}
-		if(!isPlaced()) {return true;}
+		if (otherState.getType() != getType())
+		{
+			return false;
+		}
+		if (otherState.isPlaced() != isPlaced())
+		{
+			return false;
+		}
+		if (!isPlaced())
+		{
+			return true;
+		}
 
 		return otherState.getBlock().getLocation().equals(getBlock().getLocation());
 	}
 
 	@Override
-	public int hashCode() {
+	public int hashCode()
+	{
 		return Objects.hash(material, block == null ? null : block.getLocation());
 	}
 }
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
@@ -300,7 +300,7 @@ public class BlockStateMock implements BlockState, Cloneable
 		if(!(other instanceof BlockState)) {return false;}
 
 		BlockState otherState = (BlockState) other;
-		if(!(otherState.getType() == getType())) {return false;}
+		if(otherState.getType() != getType()) {return false;}
 		if(otherState.isPlaced() != isPlaced()) {return false;}
 		if(!isPlaced()) {return true;}
 
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
@@ -1,6 +1,7 @@
 package be.seeseemelk.mockbukkit.block.state;
 
 import java.util.List;
+import java.util.Objects;
 
 import org.bukkit.Chunk;
 import org.bukkit.Location;
@@ -305,5 +306,10 @@ public class BlockStateMock implements BlockState, Cloneable
 		if(!isPlaced()) {return true;}
 
 		return otherState.getBlock().getLocation().equals(getBlock().getLocation());
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(material, block == null ? null : block.getLocation());
 	}
 }
--- a/src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
+++ b/src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
@@ -108,7 +108,7 @@ public class BlockStateMockTest
 		Block blockB = new BlockMock(Material.AIR, new Location(null, 0, 65, 0));
 		Block blockC = new BlockMock(Material.STONE, new Location(null, 0, 65, 0));
 
-		assertNotEquals(blockA.getState(), blockB.getState());
+		assertNotEquals(blockA.getState(), blockC.getState());
 		assertNotEquals(blockB.getState(), blockC.getState());
 	}
 
--- a/src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
+++ b/src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
@@ -94,7 +94,8 @@ public class BlockStateMockTest
 
 	// Tests that different instances of the placed block state of a block, both with the same material, are equal
 	@Test
-	public void testEqualsAndHashCode() {
+	public void testEqualsAndHashCode()
+	{
 		Block block = new BlockMock(Material.STONE, new Location(null, 0, 64, 0));
 
 		BlockState stateA = block.getState();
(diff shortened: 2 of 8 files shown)


src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
```diff
@@ -94,4 +94,16 @@ public void testUpdateForce()
 		assertTrue(chest.update(true));
 		assertTrue(block.getState() instanceof Chest);
 	}
+
+	@Test
+	public void testEquals() {
+		Block block = new BlockMock(Material.STONE, new Location(null, 0, 64, 0));
+
+		assertEquals(block.getState(), block.getState());
```
nuriofernandez: What's the meaning of this test?

Do you want to test if the equals work fine with two references to the same instance? In that case, you may have to specify it at the test method name or explain it in a comment or somewhere.

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
```diff
@@ -293,4 +293,17 @@ public static BlockStateMock mockState(@NotNull Block block)
...
 		}
 	}
+
+	@Override
+	public boolean equals(Object other) {
+		if(other == null) {return false;}
+		if(!(other instanceof BlockState)) {return false;}
+
+		BlockState otherState = (BlockState) other;
+		if(!(otherState.getType() == getType())) {return false;}
```
nuriofernandez: You can simplify this with `if(otherState.getType() != getType())` instead.

Lauriethefish: Oh yeah sorry I just write weird sometimes.

src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
Lauriethefish: Ah, I didn't realise that calling getState multiple times in MockBukkit returns the same object.
The intent of the test was to check if different instances of the state are tested as the same, but yes, I'll clarify it as your comment.

Lauriethefish:
I'd like to clarify, is the unique ID hashCode change fine, or is it supposed to hash everything?

src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
nuriofernandez: I took a look into the code of the `BlockMock#getState()` and will end with a call to `BlockStateMock#getSnapshot()`

https://github.com/seeseemelk/MockBukkit/blob/40e006b9ca9d783537695ed6a748bfaaba3daa21/src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java#L226-L235

That will create a new instance for each call, so your method may make sense.

Lauriethefish: OK, thanks for the clarification.
I'll comment the test and make that other change in a bit.

nuriofernandez: I would add some explanatory comments to avoid a future confusion.

Something like this one:
https://github.com/seeseemelk/MockBukkit/blob/6672eca92ec41e58121c84ef2ed92c9a98ec3301/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java#L245-L250

src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
```diff
@@ -94,4 +91,36 @@ public void testUpdateForce()
...
+
+	// Tests that different instances of the placed block state of a block, with different materials, are unequal
+	// Also tests that block states with the same material, but with different parent blocks are unequal
+	@Test
+	public void testNotEquals() {
+		Block blockA = new BlockMock(Material.STONE, new Location(null, 0, 64, 0));
+		Block blockB = new BlockMock(Material.AIR, new Location(null, 0, 65, 0));
+		Block blockC = new BlockMock(Material.STONE, new Location(null, 0, 65, 0));
+
+		assertNotEquals(blockA.getState(), blockB.getState());
```
seeseemelk: Here you compare if two blocks with different type and different parents are equal. Did you mean to compare `blockA` and `blockC` here, as you are already comparing two blocks with different types on the next line.

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
```diff
@@ -293,4 +293,17 @@ public static BlockStateMock mockState(@NotNull Block block)
...
+		if(other == null) {return false;}
+		if(!(other instanceof BlockState)) {return false;}
+
+		BlockState otherState = (BlockState) other;
+		if(otherState.getType() != getType()) {return false;}
+		if(otherState.isPlaced() != isPlaced()) {return false;}
+		if(!isPlaced()) {return true;}
+
+		return otherState.getBlock().getLocation().equals(getBlock().getLocation());
+	}
```
seeseemelk: If you add an `equal` method, you should also add a `hashCode` method.

seeseemelk (changes requested):

src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
Lauriethefish: Correct yes, thanks for pointing this out.

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
Lauriethefish: I'll add one.

Lauriethefish: hashCode method added - I also modified the existing tests to test hashCode as well.

seeseemelk (approved):

nuriofernandez (approved):

src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
```diff
@@ -18,6 +13,8 @@
 import be.seeseemelk.mockbukkit.WorldMock;
 import be.seeseemelk.mockbukkit.block.BlockMock;
 
+import static org.junit.Assert.*;
```
TheBusyBiscuit, line 16: Ugh, wildcard imports are generally not the best idea.

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
```diff
@@ -293,4 +294,22 @@ public static BlockStateMock mockState(@NotNull Block block)
...
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(material, block == null ? null : block.getLocation());
+	}
```
TheBusyBiscuit: 1. This method breaks the contract here as .hashCode() should always be inline with what .equals() does.
2. Look at the Spigot implementation, it not only takes the location (world + position) into account, but also the `BlockData`, so when mocking this behaviour should be mirrored.

```java
    @Override
    public int hashCode() {
        int hash = 7;
        hash = 73 * hash + ((this.world != null) ? this.world.hashCode() : 0);
        hash = 73 * hash + ((this.position != null) ? this.position.hashCode() : 0);
        hash = 73 * hash + ((this.data != null) ? this.data.hashCode() : 0);
        return hash;
    }
```

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
```diff
@@ -293,4 +294,22 @@ public static BlockStateMock mockState(@NotNull Block block)
...
+		if(other == null) {return false;}
+		if(!(other instanceof BlockState)) {return false;}
+
+		BlockState otherState = (BlockState) other;
+		if(otherState.getType() != getType()) {return false;}
+		if(otherState.isPlaced() != isPlaced()) {return false;}
+		if(!isPlaced()) {return true;}
+
+		return otherState.getBlock().getLocation().equals(getBlock().getLocation());
+	}
```
TheBusyBiscuit: Again, not inline with what Spigot does.
Spigot compares Location and BlockData, type/isPlaced are completely ignored.

TheBusyBiscuit (changes requested):
I like that this is getting implemented.
However this implementation strongly deviates from what Spigot does.

src/test/java/be/seeseemelk/mockbukkit/block/state/BlockStateMockTest.java
Lauriethefish: Sorry, IntelliJ autochanged it, I should turn that off.

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
Lauriethefish: BlockPosition isn't a class that I can use - it's NMS. So using the location of the internal Block makes more sense. (or null if none)
BlockData isn't yet implemented, so I can't use that.

I can change it to remove the Objects.hash call and try to mirror the Bukkit method as closely a possible in that way if you wish, but other than that I can't get much closer without implementing BlockData.

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
Lauriethefish: Spigot compares the Location and BlockData. I can't compare the BlockData as it isn't implemented, so Material is as close as I can get. BlockStateMock doesn't store a BlockPosition and World, so I can't compare those separately as CraftBukkit does. Checking if isPlaced is different (returning false if the case) and then comparing Locations afterwards is the equivalent of what Spigot does in this case. Comparing the Locations checks if the World is the same as well as the BlockPosition. 
At least, as far as I can see this is the case.

I'm not sure what you want me to change here.

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
seeseemelk: There is a `BlockDataMock`, however it is currently not being used by the `BlockStateMock`. Some work would be required to instantiate the correct `BlockDataMock` for any given `BlockStateMock`.

Lauriethefish: Ah there is, I see. I just assumed that there wasn't, since I saw that trying to get it from a state mock just threw a not implemented exception.
I don't have much time atm, but maybe I suppose.
Thanks for the insight

seeseemelk:
It seems that most of the issues with this PR are caused by the fact that `BlockDataMock` is really being used by anything yet. So in order the current issues with this PR, we should first fix the issues `BlockDataMock`.

Lauriethefish:
Agreed, that would make sense.
I assume this wouldn't be part of this PR.
Should I close it?

PR #152, opened 2021-01-21
Create Gradle workflow

seeseemelk:
# Description
This adds a workflow which automatically builds the project, runs unit tests, and validates the formatting of the source code.

### DIFF
--- a/.astylerc
+++ b/.astylerc
@@ -1,0 +1,7 @@
+--style=allman
+--mode=java
+--indent=tab
+--pad-header
+--pad-oper
+--suffix=none
+--recursive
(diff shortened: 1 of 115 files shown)


seeseemelk:
@TheBusyBiscuit This would automatically reformat any incoming pull requests.
Do the reformat files look okay to you?

TheBusyBiscuit (approved):
Formatting is a bit wonky sometimes but I hope this can be improved/configured later on.
All in all this is definitely a step in the right direction though!

I think the gradle workflow should also become a "required workflow" for pull requests, so that they cannot be merged until it completed successfully.

PR #153, opened 2021-01-22
Don't sign jar when building on Jitpack

HeroCC:
# Description
Builds are [broken](https://jitpack.io/com/github/seeseemelk/MockBukkit/v1.16-f7107abd49-1/build.log) on Jitpack due to the signing requirement. This simple PR checks for the presence of $JITPACK, and if it exists, doesn't sign.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ artifacts {
 def isFork = isFork()
 
 signing {
-    required { !isFork }
+    required { !isFork && System.getenv('JITPACK') == null }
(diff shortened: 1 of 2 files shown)


TheBusyBiscuit (approved):

TheBusyBiscuit:
*Now that Travis CI finally decided to finish...*

Thanks for the PR!
Good to know that jitpack sets an environment variable we can check for.

PR #156, opened 2021-01-24
Fix ItemMeta#clone not copying all fields

Jikoo:
# Description
ItemMeta#clone does not copy all fields, so certain data is lost on clone.

# Checklist
- [ ] Unit tests added.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,5 @@
 gradle.ext.apiVersion = '1.16'
 gradle.ext.apiVersionFull = '1.16.3-R0.1-SNAPSHOT'
-gradle.ext.version = '0.22.0'
+gradle.ext.version = '0.22.1'
 
 rootProject.name = 'MockBukkit-' + gradle.ext.apiVersion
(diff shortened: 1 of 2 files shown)


Jikoo:
I specifically didn't add another test here because it's something that's hard to continuously test correctly - each new field would need the test updated unless tests were restructured to reuse the same meta and then reflectively test fields, i.e.:

```Java
@Test
public void cloneSetsFields()
{
	ItemMetaMock clone = meta.clone();
	for (Field field : ItemMetaMock.class.getDeclaredFields()) {
		field.setAccessible(true);
		try {
			assertEquals(field.get(clone), field.get(meta));
		} catch (IllegalAccessException e) {
			fail("Encountered exception getting field content: " + e.getMessage());
		}
	}
}
```

TheBusyBiscuit (approved):
LGTM

Now we just need to fix .equals() and .hashCode() too 😅

Jikoo:
I'm willing to tack those onto this PR if you like, but I'm uncertain about how to futureproof them in tests. I don't think that https://github.com/seeseemelk/MockBukkit/pull/156#issuecomment-766452831 is necessarily a good idea because it introduces extra variation, but it would make it easier from a maintenance perspective - future implementations would just have to implement their tests prior to the declaration of the clone/equals/hashcode tests instead of having to remember to modify those tests as well.

PR #161, opened 2021-01-31
Implement plugin manager plugin loading

TheDrakonir:
# Description
I still encountered problems trying to load third party plugins with MockBukkit in order to setup a test pipeline for a plugin. However this time the upcoming issue is related to classloading. Due to this, the aim of this pull request is to adapt the plugin loading to the way Bukkit itself does it by using the PluginManager. I implemented the `loadPlugins` method on the PluginManagerMock in a way it should respect the Bukkit way of loading dependencies. I hope this will finally fix the issues I have.
In addition to the implementation of the method, I changed the `MockBukkit.loadJar()` method to use the PluginManagerMock as well. 

I still need to add the unit tests though. 

Any feedback on this pull request is highly appreciated.

# Checklist
- [ ] Unit tests added.


src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -544,15 +547,158 @@ public boolean isPluginEnabled(Plugin plugin)
...
-		throw new UnimplementedOperationException();
+		if (!file.getName().endsWith(".jar")) 
+		{
+			throw new InvalidPluginException("Plugin must be a .jar file");
+		}
+
+		try 
+		{
+			File tempDir = createTemporaryDirectory("MockBukkit-plugin-" + file.getName().replace(".", "-"));
+			Path targetPath = Files.copy(file.toPath(), tempDir.toPath().resolve(file.getName()), StandardCopyOption.REPLACE_EXISTING);
```
seeseemelk, line 661: Why is the Jar file to be loaded copied to a temporary directory? Is it not possible to load it directly from it's original location?

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -544,15 +547,158 @@ public boolean isPluginEnabled(Plugin plugin)
...
+					for (String loadBefore : pdf.getLoadBefore())
+					{
+						Set<String> softDepends = softDependencies.getOrDefault(loadBefore, new HashSet<>());
+						softDepends.add(pluginName);
+						softDependencies.put(loadBefore, softDepends);
+					}
+				}
+				catch(InvalidDescriptionException e)
+				{
+					e.printStackTrace();
```
seeseemelk, line 727: You might want to wrap and rethrow the exception here to make sure that the unit tests will fail properly, instead of just printing a stack trace and continuing. Many test runners hide any output unless the test actually fails.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -544,15 +547,158 @@ public boolean isPluginEnabled(Plugin plugin)
...
+				String pluginToLoad = pluginsToLoad.keySet().iterator().next();
+				loadedPlugins.addAll(loadPluginWithDependencies(pluginToLoad, pluginsToLoad, dependencies, softDependencies, 10));
+			}
+			return loadedPlugins.toArray(Plugin[]::new);
+		} 
+		catch (IOException | InvalidPluginException e)
+		{
+			e.printStackTrace();
+		}
+		return new Plugin[0];
```
seeseemelk, line 745: Did you mean to return an empty array here?

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -544,15 +547,158 @@ public boolean isPluginEnabled(Plugin plugin)
...
+		{
+			throw new InvalidPluginException(String.format("Error loading plugin %s: ", pluginToLoad), e);
+		}
+		finally {
+			pluginsToLoad.remove(pluginToLoad);
+		}
+		return result;
+	}
+
+	private Set<Plugin> loadDependencies(String key, Map<String, File> pluginsToLoad, Map<String, Set<String>> dependencies, Map<String, Set<String>> globalDependencies, Map<String, Set<String>> globalSoftDependencies, int recursiveDepthLeft)
```
seeseemelk, line 790: What is the `key` parameter exactly? Is the name of a plugin?

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -544,15 +547,158 @@ public boolean isPluginEnabled(Plugin plugin)
...
+				}
+			}
+
+			Set<Plugin> loadedPlugins = new HashSet<>();
+
+			//load in order gathered before
+			while(!pluginsToLoad.isEmpty())
+			{
+				String pluginToLoad = pluginsToLoad.keySet().iterator().next();
+				loadedPlugins.addAll(loadPluginWithDependencies(pluginToLoad, pluginsToLoad, dependencies, softDependencies, 10));
```
seeseemelk, line 737: It took me a while to realise that `pluginsToLoad` is being modified by the `loadPluginWithDependencies` method. Is there a way you to write it without modifying it within `loadPluginWithDependencies`?

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -544,15 +547,158 @@ public boolean isPluginEnabled(Plugin plugin)
...
+			throw new InvalidDescriptionException(String.format("Plugin %s wasn't loaded: The plugin name isn't allowed to have spaces!", originFile.getName()));
+		}
+	}
+
+	private Set<Plugin> loadPluginWithDependencies(String pluginToLoad, Map<String, File> pluginsToLoad, Map<String, Set<String>> dependencies, Map<String, Set<String>> softDependencies, int recursiveDepthLeft) throws InvalidPluginException
+	{
+		Set<Plugin> result = new HashSet<>();
+		result.addAll(loadDependencies(pluginToLoad, pluginsToLoad, dependencies, dependencies, softDependencies, recursiveDepthLeft));
+		try{
+			result.addAll(loadDependencies(pluginToLoad, pluginsToLoad, softDependencies, dependencies, softDependencies, recursiveDepthLeft));
```
seeseemelk, line 769: You're loading all plugin dependencies twice. Did you mean for it to work like this?

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -544,15 +547,158 @@ public boolean isPluginEnabled(Plugin plugin)
...
+			}
+		}
+	
+		if (pluginName.contains(" "))
+		{
+			throw new InvalidDescriptionException(String.format("Plugin %s wasn't loaded: The plugin name isn't allowed to have spaces!", originFile.getName()));
+		}
+	}
+
+	private Set<Plugin> loadPluginWithDependencies(String pluginToLoad, Map<String, File> pluginsToLoad, Map<String, Set<String>> dependencies, Map<String, Set<String>> softDependencies, int recursiveDepthLeft) throws InvalidPluginException
```
seeseemelk, line 764: What does the `recursiveDepthLeft` parameter do? Is it to prevent cyclic dependencies from causing an infinite loop/stack overflow?
If it is, you might be better off to keep a set of all plugins you've already attempted to load. Whenever you try to load a plugin, you can check if it is the map. If it is in the map, you return immediately, else you add it to the map then try to load that plugin's dependencies.
This way you'll be able to immediately detect any recursive dependencies, and you'll be able to handle more than 10 dependencies deep (not that such a scenario would occur often).

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -423,7 +423,7 @@ public String getName()
 	@Override
 	public String getVersion()
 	{
-		return "0.1.0";
+		return getBukkitVersion() + "-MockBukkit";
```
seeseemelk: I'm not sure if it is safe to add text to this all willy-nilly. Some plugins and frameworks (e.g.: Paper) perform regex matches on this. @TheBusyBiscuit probably knows what version strings can actually be found in the wild. We should try to conform to that format as much as possible.

seeseemelk (commented):
I have some questions about some of the things you do here. Mostly small stuff though. It's good that this is being moved to the proper class though, thank you for your PR!

The issue with `ClassLoader`s is most likely Bukkit's fault. The default constructor of Bukkit's `Plugin` class checks if the plugin was loaded with a specific class loader, and if not it throws an exception. This is the sole reason any tested plugin _must_ provide a custom constructor.
The best way to fix this would be to modify the default constructor of the `Plugin` class itself. (although perhaps we could add such a custom constructor to third-party plugins at runtime using some reflection magic and/or modifying a plugin's Jar bytecode at runtime.)

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
TheDrakonir: I wanted to recreate the already existing loading mechanism of the "internal" plugins as close as possible. As they are passed a temporary data directory I thought of recreating that here. It would be no problem though to change that and load the plugin directly in that folder.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
TheDrakonir: Yes, but I think it's the same problem like in 622-624. So instead of having an empty array returned on error, I should just pass on the exception.

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
TheDrakonir: I think I might revert this change and open another issue/pull request to address this as it's just indirectly related to this topic. (WorldEdit doesn't loading cause it thinks the server version is too old was the origin)

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
TheDrakonir: Yes, this was my intention. But your approach sounds a lot better, gonna work on implementing that. (And resolving the other problems of course :D)

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
TheBusyBiscuit: 👍🏻 Feel free to open a new Pull Request for this and I will merge it asap.
Since the library PaperLib also relies on the version following a certain format:
https://github.com/PaperMC/PaperLib/blob/master/src/main/java/io/papermc/lib/environments/Environment.java#L45-L68

And many plugins utilise this library, so following that (unspoken) convention would be nice.
"MockBukkit (MC: ${bukkitVersion})" would probably be ideal.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
seeseemelk: We do indeed need a temporary data directory in case the plugin wants to store some configuration or some data. However, we do not have to copy the jar file itself to this directory.

We also have to make sure to remove that data after the test has finished in order to prevent someone's disk from filling up if they have to run hundreds of tests. On my Linux system, for instance, it's all being stored in `/tmp` which is just a ram disk. It's really fast and won't wear out my SSD, but the amount of data I can store there is limited by how much memory I have. Running a test suite consisting of many tests should not be filling that up.

The current implementation has a `List<File> temporaryFiles` which contains all temporary directories and files that were created during a single unit test. When `MockBukkit.unload()` is called, each directory and file in this list gets removed in order to avoid this issue.

thelooter:
@TheDrakonir any update on this?

thelooter:
I don't think there will be any progress on this so I'm gonna close this for now. If @TheDrakonir or anyone else wants to continue to work on this, fell free to reopen

PR #168, opened 2021-02-26
Update scoreboard methods and implement proper state and argument validation

TheBusyBiscuit:
# Description
This pull request implements the overloaded `ScoreboardMock#registerNewObjective()` methods which are long overdue for implementation.
This implementation gap was discovered by a user on discord in the `#mockbukkit-help` channel.
I also took this oppurtunity to implement all the state and argument validations that the documented `@throws` clauses dictate.
An `IllegalStateException` will be thrown for any `ObjectiveMock` method that is invoked while the objective is not registered.
Furthermore, argument validation was implemented by the usage of Apache's `Validate` class which throws an `IllegalArgumentException` (inline with the Bukkit documentation).
Even character limits are now enforced.

# Checklist
- [ ] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/scoreboard/ScoreboardMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/scoreboard/ScoreboardMock.java
@@ -1,6 +1,11 @@
 package be.seeseemelk.mockbukkit.scoreboard;
 
-import java.util.*;
+import java.util.Collections;
+import java.util.EnumMap;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.bukkit.OfflinePlayer;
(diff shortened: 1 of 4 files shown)


src/main/java/be/seeseemelk/mockbukkit/scoreboard/ScoreboardMock.java
```diff
@@ -37,9 +55,8 @@ public ObjectiveMock getObjective(String name) throws IllegalArgumentException
 	@Override
 	public Set<Objective> getObjectivesByCriteria(String criteria) throws IllegalArgumentException
 	{
-		return objectives.values().stream()
-		       .filter(objective -> objective.getCriteria().equals(criteria))
-		       .collect(Collectors.toSet());
+		return objectives.values().stream().filter(objective -> objective.getCriteria().equals(criteria))
+				.collect(Collectors.toSet());
```
seeseemelk, line 59: This is much harder to read imo

seeseemelk (approved):

src/main/java/be/seeseemelk/mockbukkit/scoreboard/ScoreboardMock.java
TheBusyBiscuit: I agree.
Didn't change that though, probably my auto fornatter.
We really need to get the formatter workflow running again.

PR #170, opened 2021-02-27
[ADD] Add support for Bukkit.getServicesManager() (Minecraft 1.15)

0ddlyoko:
# Description
Add support for Bukkit.getServicesManager() method

# Checklist
- [x] Unit tests added.

Version number isn't in file settings.gradle

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -69,7 +69,7 @@ import org.bukkit.inventory.Recipe;
 import org.bukkit.loot.LootTable;
 import org.bukkit.map.MapView;
 import org.bukkit.plugin.Plugin;
-import org.bukkit.plugin.ServicesManager;
+import org.bukkit.plugin.SimpleServicesManager;
 import org.bukkit.plugin.messaging.Messenger;
 import org.bukkit.potion.PotionEffectType;
 import org.bukkit.util.CachedServerIcon;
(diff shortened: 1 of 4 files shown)


0ddlyoko:
Should I add support for Minecraft 1.14 & lower ?

seeseemelk:
We're not really actively back-porting features to older version of Minecraft as we haven't really gotten any requests from anyone to do so. However, we do keep try to keep our infrastructure set up so that we can easily back port features if we need to.

seeseemelk:
> However, we do keep try to keep our infrastructure set up so that we can easily back port features if we need to.

And looking at our current infrastructure, I'm pretty sure we broke it somewhat recently...
I'll merge it in case it helps someone, but I don't think this will end up Maven central.

0ddlyoko:
> > However, we do keep try to keep our infrastructure set up so that we can easily back port features if we need to.
> 
> And looking at our current infrastructure, I'm pretty sure we broke it somewhat recently...
> I'll merge it in case it helps someone, but I don't think this will end up Maven central.

I was trying to cherry pick the commit to Minecraft 1.14 but I got some conflicts and the test took a long time to run

seeseemelk:
Hmm, it seems our builds are not being executed for the `v1.15`. This is probably because we moved from Travis CI to GitHub actions very recently, and this branch has not yet been configured for that. Until that is fixed, this PR will sadly not be built automatically.

0ddlyoko:
Okay, no problem, I just need it for Minecraft 1.16 :)

Vinetos:
> Hmm, it seems our builds are not being executed for the `v1.15`. This is probably because we moved from Travis CI to GitHub actions very recently, and this branch has not yet been configured for that. Until that is fixed, this PR will sadly not be built automatically.

What do you need exactly ? Actually, the github Actions are triggered only on 1.16 branch

seeseemelk:
Some GitHub actions are being executed as of #182, so we should be able to merge it now.

PR #171, opened 2021-02-27
Minor Changes

0ddlyoko:
# Description
Some minor changes

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
@@ -177,7 +177,7 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public BlockMock getBlockAt(int x, int y, int z)
+	public @NotNull BlockMock getBlockAt(int x, int y, int z)
 	{
 		Coordinate coordinate = new Coordinate(x, y, z);
 		if (blocks.containsKey(coordinate))
@@ -191,13 +191,13 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public BlockMock getBlockAt(Location location)
+	public @NotNull BlockMock getBlockAt(Location location)
 	{
 		return getBlockAt(location.getBlockX(), location.getBlockY(), location.getBlockZ());
 	}
 
 	@Override
-	public String getName()
+	public @NotNull String getName()
 	{
 		return name;
 	}
@@ -213,13 +213,13 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public UUID getUID()
+	public @NotNull UUID getUID()
 	{
 		return uuid;
 	}
 
 	@Override
-	public Location getSpawnLocation()
+	public @NotNull Location getSpawnLocation()
 	{
 		if (spawnLocation == null)
 		{
@@ -251,7 +251,7 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public List<Entity> getEntities()
+	public @NotNull List<Entity> getEntities()
 	{
 		// MockBukkit.assertMocking();
 		List<Entity> entities = new ArrayList<>();
@@ -263,46 +263,46 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public ChunkMock getChunkAt(int x, int z)
+	public @NotNull ChunkMock getChunkAt(int x, int z)
 	{
 		ChunkMock chunk = new ChunkMock(this, x, z);
 		return chunk;
 	}
 
 	@Override
-	public void sendPluginMessage(Plugin source, String channel, byte[] message)
+	public void sendPluginMessage(@NotNull Plugin source, @NotNull String channel, byte[] message)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public Set<String> getListeningPluginChannels()
+	public @NotNull Set<String> getListeningPluginChannels()
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public void setMetadata(String metadataKey, MetadataValue newMetadataValue)
+	public void setMetadata(@NotNull String metadataKey, @NotNull MetadataValue newMetadataValue)
 	{
 		metadataTable.setMetadata(metadataKey, newMetadataValue);
 	}
 
 	@Override
-	public List<MetadataValue> getMetadata(String metadataKey)
+	public @NotNull List<MetadataValue> getMetadata(@NotNull String metadataKey)
 	{
 		return metadataTable.getMetadata(metadataKey);
 	}
 
 	@Override
-	public boolean hasMetadata(String metadataKey)
+	public boolean hasMetadata(@NotNull String metadataKey)
 	{
 		return metadataTable.hasMetadata(metadataKey);
 	}
 
 	@Override
-	public void removeMetadata(String metadataKey, Plugin owningPlugin)
+	public void removeMetadata(@NotNull String metadataKey, @NotNull Plugin owningPlugin)
 	{
 		metadataTable.removeMetadata(metadataKey, owningPlugin);
 	}
@@ -315,42 +315,42 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public int getHighestBlockYAt(Location location)
+	public int getHighestBlockYAt(@NotNull Location location)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public Block getHighestBlockAt(int x, int z)
+	public @NotNull Block getHighestBlockAt(int x, int z)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public Block getHighestBlockAt(Location location)
+	public @NotNull Block getHighestBlockAt(@NotNull Location location)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public Chunk getChunkAt(Location location)
+	public @NotNull Chunk getChunkAt(@NotNull Location location)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public Chunk getChunkAt(Block block)
+	public @NotNull Chunk getChunkAt(@NotNull Block block)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public boolean isChunkLoaded(Chunk chunk)
+	public boolean isChunkLoaded(@NotNull Chunk chunk)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
@@ -364,7 +364,7 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public void loadChunk(Chunk chunk)
+	public void loadChunk(@NotNull Chunk chunk)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
@@ -399,7 +399,7 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public boolean unloadChunk(Chunk chunk)
+	public boolean unloadChunk(@NotNull Chunk chunk)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
@@ -442,7 +442,7 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public Item dropItem(@NotNull Location loc, ItemStack item)
+	public @NotNull Item dropItem(@NotNull Location loc, @NotNull ItemStack item)
 	{
 		if (item == null || item.getType() == Material.AIR)
 		{
@@ -456,7 +456,7 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public Item dropItemNaturally(@NotNull Location location, ItemStack item)
+	public @NotNull Item dropItemNaturally(@NotNull Location location, @NotNull ItemStack item)
 	{
 		Random random = ThreadLocalRandom.current();
 
@@ -473,28 +473,28 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public Arrow spawnArrow(Location location, Vector direction, float speed, float spread)
+	public @NotNull Arrow spawnArrow(@NotNull Location location, @NotNull Vector direction, float speed, float spread)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public boolean generateTree(Location location, TreeType type)
+	public boolean generateTree(@NotNull Location location, @NotNull TreeType type)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public boolean generateTree(Location loc, TreeType type, BlockChangeDelegate delegate)
+	public boolean generateTree(@NotNull Location loc, @NotNull TreeType type, @NotNull BlockChangeDelegate delegate)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public Entity spawnEntity(Location loc, EntityType type)
+	public @NotNull Entity spawnEntity(@NotNull Location loc, @NotNull EntityType type)
 	{
 		EntityMock entity = mockEntity(type);
 		entity.setLocation(loc);
@@ -527,21 +527,21 @@ public class WorldMock implements World
 	}
 
 	@Override
-	public LightningStrike strikeLightning(Location loc)
+	public @NotNull LightningStrike strikeLightning(@NotNull Location loc)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public LightningStrike strikeLightningEffect(Location loc)
+	public @NotNull LightningStrike strikeLightningEffect(@NotNull Location loc)
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public List<LivingEntity> getLivingEntities()
+	public @NotNull List<LivingEntity> getLivingEntities()
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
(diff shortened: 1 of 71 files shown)


src/main/java/be/seeseemelk/mockbukkit/attribute/AttributeInstanceMock.java
```diff
@@ -1,13 +1,18 @@
 package be.seeseemelk.mockbukkit.attribute;
 
 import java.util.Collection;
+import java.util.Objects;
+
+import org.junit.Assert;
```
TheBusyBiscuit, line 6: unused imports

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -1,5 +1,7 @@
 package be.seeseemelk.mockbukkit.block.data;
 
+import java.util.Objects;
```
TheBusyBiscuit, line 3: Unused import

src/main/java/be/seeseemelk/mockbukkit/entity/ArmorStandMock.java
```diff
@@ -38,22 +40,23 @@ public ArmorStandMock(ServerMock server, UUID uuid)
...
 
 	@Override
-	public EntityType getType()
+	public @NotNull EntityType getType()
 	{
 		return EntityType.ARMOR_STAND;
 	}
 
 	@Override
+	@NotNull
```
TheBusyBiscuit, line 49: Inconsistently placed annotations, on all methods you added it inline before the return type, here you added it to the method body.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityEquipmentMock.java
```diff
@@ -71,30 +72,38 @@ public void setItem(@NotNull EquipmentSlot slot, @Nullable ItemStack item, boole
...
+			break;
 		case OFF_HAND:
-			return getItemInOffHand();
+			itemStack = getItemInOffHand();
+			break;
 		default:
 			// This should never be reached unless Mojang adds new slots
 			throw new UnimplementedOperationException("EquipmentSlot '" + slot + "' has no implementation!");
 		}
+		return itemStack == null ? new ItemStack(Material.AIR) : itemStack;
```
TheBusyBiscuit, line 102: This implementation is redundant.
There is no point in storing the variable when we can just return it immediately.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityEquipmentMock.java
```diff
@@ -231,8 +240,12 @@ public void setBoots(ItemStack boots, boolean silent)
 	@Override
 	public void setArmorContents(ItemStack[] items)
```
TheBusyBiscuit, line 241: No annotation here?

src/main/java/be/seeseemelk/mockbukkit/entity/EntityEquipmentMock.java
```diff
@@ -231,8 +240,12 @@ public void setBoots(ItemStack boots, boolean silent)
...
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		if (items.length == 4) {
+			setHelmet(items[0]);
+			setChestplate(items[1]);
+			setLeggings(items[2]);
+			setBoots(items[3]);
+		}
```
TheBusyBiscuit, line 248: else?

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -341,48 +345,45 @@ public boolean hasPermission(String name)
...
 
 	@Override
-	public void removeAttachment(PermissionAttachment attachment)
+	public void removeAttachment(@NotNull PermissionAttachment attachment)
 	{
-		if (attachment == null)
-		{
-			throw new IllegalArgumentException("Attachment cannot be null");
-		}
+		Validate.notNull(attachment);
```
TheBusyBiscuit, line 386: You should add a message for context to the notNull() call

src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -11,11 +11,7 @@
 import java.util.Set;
 import java.util.UUID;
 
-import org.bukkit.Bukkit;
-import org.bukkit.FluidCollisionMode;
-import org.bukkit.GameRule;
-import org.bukkit.Location;
-import org.bukkit.Material;
+import org.bukkit.*;
```
TheBusyBiscuit, line 14: Never use wildcard imports please.

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
```diff
@@ -271,12 +271,9 @@ public void setStorageContents(ItemStack[] items) throws IllegalArgumentExceptio
...
 
 	@Override
-	public boolean contains(Material material) throws IllegalArgumentException
+	public boolean contains(@NotNull Material material) throws IllegalArgumentException
 	{
-		if (material == null)
-		{
-			throw new IllegalArgumentException("Material cannot be null.");
-		}
+		Validate.notNull(material);
```
TheBusyBiscuit, line 276: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
```diff
@@ -294,12 +291,9 @@ public boolean contains(ItemStack item)
...
 
 	@Override
-	public boolean contains(Material material, int amount) throws IllegalArgumentException
+	public boolean contains(@NotNull Material material, int amount) throws IllegalArgumentException
 	{
-		if (material == null)
-		{
-			throw new IllegalArgumentException("Material cannot be null.");
-		}
+		Validate.notNull(material);
```
TheBusyBiscuit, line 296: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
```diff
@@ -128,25 +129,22 @@ private boolean isValidPage(int page)
 	}
 
 	@Override
-	public void setPage(int page, String text)
+	public void setPage(int page, @NotNull String text)
 	{
+		Validate.notNull(text);
```
TheBusyBiscuit, line 134: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/EnchantedBookMetaMock.java
```diff
@@ -71,6 +73,7 @@ public EnchantedBookMetaMock clone()
 	@Override
 	public boolean addStoredEnchant(@NotNull Enchantment ench, int level, boolean ignoreLevelRestriction)
 	{
+		Validate.notNull(ench);
```
TheBusyBiscuit, line 76: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/EnchantedBookMetaMock.java
```diff
@@ -82,7 +85,13 @@ public boolean addStoredEnchant(@NotNull Enchantment ench, int level, boolean ig
 		}
 
 		Integer prev = storedEnchantments.put(ench, level);
-		return prev == null || prev.intValue() != level;
+		return prev == null || prev != level;
+	}
+	@Override
+	public boolean removeStoredEnchant(@NotNull Enchantment ench) throws IllegalArgumentException
+	{
+		Validate.notNull(ench);
```
TheBusyBiscuit, line 93: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/EnchantedBookMetaMock.java
```diff
@@ -82,7 +85,13 @@ public boolean addStoredEnchant(@NotNull Enchantment ench, int level, boolean ig
 		}
 
 		Integer prev = storedEnchantments.put(ench, level);
-		return prev == null || prev.intValue() != level;
+		return prev == null || prev != level;
```
TheBusyBiscuit, line 88: Revert this please.
```suggestion
		return prev == null || prev.intValue() != level;
```

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/KnowledgeBookMetaMock.java
```diff
@@ -101,6 +101,7 @@ public boolean hasRecipes()
 	@Override
 	public void setRecipes(@NotNull List<NamespacedKey> recipes)
 	{
+		Validate.notNull(recipes);
```
TheBusyBiscuit, line 104: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -442,9 +443,11 @@ public boolean refreshChunk(int x, int z)
 	}
 
 	@Override
-	public Item dropItem(@NotNull Location loc, ItemStack item)
+	public @NotNull Item dropItem(@NotNull Location loc, @NotNull ItemStack item)
 	{
-		if (item == null || item.getType() == Material.AIR)
+		Validate.notNull(loc);
```
TheBusyBiscuit, line 448: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -442,9 +443,11 @@ public boolean refreshChunk(int x, int z)
...
+		Validate.notNull(item);
```
TheBusyBiscuit, line 449: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -1072,15 +1075,16 @@ public String getGameRuleValue(String rule)
...
 	}
 
 	@SuppressWarnings("unchecked")
 	@Override
-	public boolean setGameRuleValue(String rule, String value)
+	public boolean setGameRuleValue(@NotNull String rule, @NotNull String value)
 	{
-		if (rule == null)
-			return false;
+		Validate.notNull(rule);
```
TheBusyBiscuit, line 1086: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -1072,15 +1075,16 @@ public String getGameRuleValue(String rule)
...
+		Validate.notNull(value);
```
TheBusyBiscuit, line 1087: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -1104,112 +1108,113 @@ else if (gameRule.getType().equals(Integer.TYPE))
 	}
 
 	@Override
-	public boolean isGameRule(String rule)
+	public boolean isGameRule(@NotNull String rule)
 	{
-		return rule != null && GameRule.getByName(rule) != null;
+		Validate.notNull(rule);
```
TheBusyBiscuit, line 1113: Please add a message for context.

src/main/java/be/seeseemelk/mockbukkit/entity/ArmorStandMock.java
```diff
@@ -38,22 +40,23 @@ public ArmorStandMock(ServerMock server, UUID uuid)
...
 		return equipment;
 	}
 
 	@Override
 	@Deprecated
-	public ItemStack getBoots()
+	public @NotNull ItemStack getBoots()
 	{
-		return getEquipment().getBoots();
+		return getEquipment().getItem(EquipmentSlot.FEET);
```
TheBusyBiscuit, line 59: Why?

src/main/java/be/seeseemelk/mockbukkit/entity/ArmorStandMock.java
```diff
@@ -65,9 +68,9 @@ public void setBoots(ItemStack item)
 
 	@Override
 	@Deprecated
-	public ItemStack getLeggings()
+	public @NotNull ItemStack getLeggings()
 	{
-		return getEquipment().getLeggings();
+		return getEquipment().getItem(EquipmentSlot.LEGS);
```
TheBusyBiscuit, line 73: Why?

src/main/java/be/seeseemelk/mockbukkit/entity/ArmorStandMock.java
```diff
@@ -79,9 +82,9 @@ public void setLeggings(ItemStack item)
 
 	@Override
 	@Deprecated
-	public ItemStack getChestplate()
+	public @NotNull ItemStack getChestplate()
 	{
-		return getEquipment().getChestplate();
+		return getEquipment().getItem(EquipmentSlot.CHEST);
```
TheBusyBiscuit, line 87: Why?

src/main/java/be/seeseemelk/mockbukkit/entity/ArmorStandMock.java
```diff
@@ -93,9 +96,9 @@ public void setChestplate(ItemStack item)
 
 	@Override
 	@Deprecated
-	public ItemStack getHelmet()
+	public @NotNull ItemStack getHelmet()
 	{
-		return getEquipment().getHelmet();
+		return getEquipment().getItem(EquipmentSlot.HEAD);
```
TheBusyBiscuit, line 101: Why?

TheBusyBiscuit (changes requested):
I didn't review every single file yet.
But I already found of things which will need to be fixed before I'd feel comfortable merging this.
Also I didn't check but I'll trust you that the annotations are all inline with the method contracts of the Bukkit API.

Another very important thing to note is that Jetbrains annotations do not validate parameters, so you may need to make sure that the appropriate exception `IllegalArgumentException` or `NullPointerException` is thrown for null values.

0ddlyoko:
> Another very important thing to note is that Jetbrains annotations do not validate parameters, so you may need to make sure that the appropriate exception `IllegalArgumentException` or `NullPointerException` is thrown for null values.

Yep, I know for that.
This is why we should use Validate.notNull(...) everywhere

src/main/java/be/seeseemelk/mockbukkit/entity/ArmorStandMock.java
0ddlyoko: getBoots() could return null, getItem(...) not. Another way is to check if getBoots() returns null. If true, return Air

src/main/java/be/seeseemelk/mockbukkit/entity/EntityEquipmentMock.java
0ddlyoko: We don't know what to do if we give a different array size. Should we add the first 4 items ? Should we do nothing ? Should we throw an Exception ?

src/main/java/be/seeseemelk/mockbukkit/entity/EntityEquipmentMock.java
0ddlyoko: We don't know what to do if we give a different array size. Should we add the first 4 items ? Should we do nothing ? Should we throw an Exception ?

src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
0ddlyoko: Yes sorry, I'll disable this feature in Intellij

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/EnchantedBookMetaMock.java
0ddlyoko: Why ?
https://stackoverflow.com/a/11490442

src/main/java/be/seeseemelk/mockbukkit/entity/EntityEquipmentMock.java
TheBusyBiscuit: We certainly know what to do.
MockBukkit is a **mock** of CraftBukkit after all, so all methods should always reflect the implementations found in CraftBukkit, for this method the standard implementation is as follows:

```java
    @Override
    public void setArmorContents(final ItemStack[] items) {
        this.setEquipment(EnumItemSlot.FEET, (items.length >= 1) ? items[0] : null);
        this.setEquipment(EnumItemSlot.LEGS, (items.length >= 2) ? items[1] : null);
        this.setEquipment(EnumItemSlot.CHEST, (items.length >= 3) ? items[2] : null);
        this.setEquipment(EnumItemSlot.HEAD, (items.length >= 4) ? items[3] : null);
    }
```

Not only did you get the order completely reversed, but smaller arrays are completely ignored, whereas they should be allowed.

TheBusyBiscuit:
> Yep, I know for that.
> This is why we should use Validate.notNull(...) everywhere

We shouldn't.
I would agree if this was standalone software, but this is a **mock**, so we should only throw an `IllegalArgumentException` where CraftBukkit does.
There are a lot of methods in CB that will also just result in a `NullPointerException`.

0ddlyoko:
Rip

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/EnchantedBookMetaMock.java
TheBusyBiscuit: You should perhaps read the thread you posted.
Because it precisely tells you why this is wrong...

The `==` operator is **not** an equality check, it is a memory address comparison.
Only for primitives can it be used to check for equality but not for Objects. The `Integer` class is what we call a "boxed primitive", it's a wrapper for a primitive but itself is still an `Object` and will follow the object's rules.
Two objects will only ever be equal if they are stored under the same memory location.
Now there are some exceptions due to JVM optimizations which will for example put literals and constants under the same memory location sometimes. But for almost all production code the usage of `==` for comparing Objects should be avoided **at all cost**. It is a really bad practice. And that also applies for boxed integers.
The JVM may still do an implict type transformation to make this work sometimes but it is inherently flawed and should not be relied upon.
That's why this change needs to reverted immediately, we need to make sure that this is correctly converted into a primitive.

seeseemelk:
I'm afraid I cannot merge this pull request for a couple of reasons. The PR description simply says "Some minor changes", which is not very descriptive.

If it was a small PR that only changed a couple of files, this would perhaps be okay. However, this PR changes 975 lines across 71 files. This is in no way 'minor', and it is not really reviewable at all.

The commit history does not make it any easier to try and understand it, and it also does not allow me to easily split it up into smaller (manageable) PRs. 

If the description was clearer about why it was so large, this *might* have been passable, but in it's current state I don't even want to begin reviewing this.

I'm closing this PR. If you still want to see this getting merged, you can split this PR up into several smaller pull requests which can be reviewed and merged individually. This also means that if a small change blocks a certain PR, other unrelated changes are not blocked (as they should be different PRs).

PR #172, opened 2021-02-28
Use command map instead of command array and other minor improvements

nowtryz:
# Description

## Use `CommandMap`
I know a lot of command frameworks work with the `CommandMap` to dynamically introduce commands to bukkit. Though MockBukkit was only using its command array.
Hence this PR changes calls to `PluginManagerMock.getCommands` to `MockCommandMap.getCommand`.

## Player list's name
As easy as it seems, I add the implementation for getter and setter of the player list's name

## Console sender
As the console must always be considered op, I updated the getter for the console as well as its name.

## `getKiller` and `setHealth`
I added an implementation for the `getKiller` method in `LivingEntity` and refactor the death logic so every entity can override the death behavior.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,5 @@
 gradle.ext.apiVersion = '1.16'
 gradle.ext.apiVersionFull = '1.16.5-R0.1-SNAPSHOT'
-gradle.ext.version = '0.27.0'
+gradle.ext.version = '0.27.1'
 
 rootProject.name = 'MockBukkit-' + gradle.ext.apiVersion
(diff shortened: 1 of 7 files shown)


TheBusyBiscuit (commented):
Please update the version in `settings.gradle` too, should be a minor bump (+ 0.0.1)

nowtryz:
> Please update the version in `settings.gradle` too, should be a minor bump (+ 0.0.1)

Yep I forgot on the previous commit, sorry. But with the new commit, It would be a minor bump instead of a patch right?

TheBusyBiscuit:
You broadened out the scope a lot here.
I dont think this counts as a patch (`0.0.1`) anymore, the version should probably be bumped by a minor release (`0.1.0`) instead.

Though I suggest you look into working with branches and splitting up these features into multiple branches and pull requests.
It would also make it a lot easier for us to review changes in small batches rather than multiple different things at once.

nowtryz:
No problem, I'will split it into multiple PRs

nowtryz:
Closing in favor to #174, #175, #176 and #177.

Issue #162, opened 2021-02-03
AsyncTask list can orphan the  the queue 


Narimm:
```
[executorService] executorService = {BukkitSchedulerMock@5945} 
   currentTick = 5645936
   id = 15
   tasks = {LinkedList@6053}  size = 0
   pool = {ThreadPoolExecutor@5948} "java.util.concurrent.ThreadPoolExecutor@1cd2143b[Running, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 6]"
   asyncTasksRunning = {AtomicInteger@5949} "0"
   value = 0
   asyncException = {AtomicReference@5950} "be.seeseemelk.mockbukkit.UnimplementedOperationException: Not implemented"
   value = {UnimplementedOperationException@6026} "be.seeseemelk.mockbukkit.UnimplementedOperationException: Not implemented"
   asyncTasksQueued = 2
```
 
See above that there are 2 tasks queued but the task list is empty....clearly this will never complete.
I think this state exists if a scheduledasynctask is cancelled before it ever runs.  This method needs a safety check if this occurs

```java
public void performOneTick()
	{
		currentTick++;
		List<ScheduledTask> oldTasks = tasks;
		tasks = new LinkedList<>();

		for (ScheduledTask task : oldTasks)
		{
			if (task.getScheduledTick() == currentTick && !task.isCancelled())
			{
				if (task.isSync())
				{
					task.run();
				}
				else
				{
					asyncTasksRunning.incrementAndGet();
					pool.execute(task.getRunnable());
					asyncTasksQueued--;
				}

				if (task instanceof RepeatingTask && !task.isCancelled())
				{
					((RepeatingTask) task).updateScheduledTick();
					tasks.add(task);
				}
			}
			else if (!task.isCancelled())
			{
				tasks.add(task);
			}
		}
	}
```

Narimm:
Just to be clear this results in the  `BukkitSchedulerMock.waitAsyncTasksFinished()` never returning.

Narimm:
I came up with a few options to handle this it primarily blocks ` MockBukkit.unmock() `
1. Force a shutdown.
2. Reset the integer.
3. Limit the time you wait for async tasks to run

adding this to the tick would probably be enough

```java
		if (oldTasks.size() == 0 && asyncTasksQueued > 0) {
			Logger.getLogger(LOGGER_NAME).warning("Asynchronous Tasks may have been lost -> " + asyncTasksQueued);
[...]

PR #173, opened 2021-03-01
Fix for Async task orphaned by the scheduler causing a lockup on shutdown during unit testing.

Narimm:
# Description
This PR was designed to go some way to fix the issues in #162.   In this PR we still wait for the queue but if the async queue count exceeds the actual list of tasks the plugin will cancel them and eject them logging an error to console.    

It will not fail the tests.  In Spigot and Paper, the scheduler doesn't wait but it will force shutdown of the queue if there was a long-running task doing something like:-

```java
while(true) {
   Thread.sleep(1000)
}
```

A fully fleshed out scheduler would be ideal but beyond the scope of this PR.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/PlayerList.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/PlayerList.java
@@ -22,10 +22,10 @@ import be.seeseemelk.mockbukkit.entity.PlayerMock;
 
 /**
  * Replica of the Bukkit internal PlayerList and CraftPlayerList implementation
- * 
+ *
  * @author seeseemelk
  * @author TheBusyBiscuit
- * 
+ *
  */
 public class PlayerList
 {
@@ -75,7 +75,7 @@ public class PlayerList
 	public Set<OfflinePlayer> getOperators()
 	{
 		return Stream.concat(onlinePlayers.stream(), offlinePlayers.stream()).filter(OfflinePlayer::isOp)
-				.collect(Collectors.toSet());
+		       .collect(Collectors.toSet());
 	}
 
 	@NotNull
@@ -99,16 +99,16 @@ public class PlayerList
 	public List<Player> matchPlayer(@NotNull String name)
 	{
 		return onlinePlayers.stream().filter(
-				player -> player.getName().toLowerCase(Locale.ENGLISH).startsWith(name.toLowerCase(Locale.ENGLISH)))
-				.collect(Collectors.toList());
+		           player -> player.getName().toLowerCase(Locale.ENGLISH).startsWith(name.toLowerCase(Locale.ENGLISH)))
+		       .collect(Collectors.toList());
 	}
 
 	@Nullable
 	public Player getPlayerExact(@NotNull String name)
 	{
 		return onlinePlayers.stream()
-				.filter(player -> player.getName().toLowerCase(Locale.ENGLISH).equals(name.toLowerCase(Locale.ENGLISH)))
-				.findFirst().orElse(null);
+		       .filter(player -> player.getName().toLowerCase(Locale.ENGLISH).equals(name.toLowerCase(Locale.ENGLISH)))
+		       .findFirst().orElse(null);
 	}
 
 	@Nullable
--- a/src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
@@ -260,11 +260,13 @@ public class BukkitSchedulerMock implements BukkitScheduler
 		}
 	}
 
-	private void cancelTask(@NotNull ScheduledTask task){
+	private void cancelTask(@NotNull ScheduledTask task)
+	{
 
-		if(!task.isCancelled()) {
+		if (!task.isCancelled())
+		{
 			task.cancel();
-			if(asyncTasksQueued>0)
+			if (asyncTasksQueued > 0)
 			{
 				asyncTasksQueued--;
 			}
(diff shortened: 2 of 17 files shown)


src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -281,67 +304,91 @@ public boolean isQueued(int taskId)
...
+	public @NotNull BukkitTask runTaskTimerAsynchronously(@NotNull Plugin plugin, @NotNull Runnable task, long delay, long period)
 	{
 		RepeatingTask scheduledTask = new RepeatingTask(id++, plugin, false, currentTick + delay, period,
 		        new AsyncRunnable(task));
+		checkForShutDown(scheduledTask);
 		tasks.add(scheduledTask);
 		return scheduledTask;
 	}
 
+	private void checkForShutDown(ScheduledTask task)
```
TheBusyBiscuit, line 370: I reckon we could add an annotation here as well.
```suggestion
	private void checkForShutDown(@NotNull ScheduledTask task)
```

TheBusyBiscuit (approved):
LGTM.

There is one parameter annotation that I think should be added but it's not a big deal.
I will hold off on merging this for now though as I'd like to have @seeseemelk see this first.

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
Narimm: Yeah I guess because it was private I didn't really get overly concerned

TheBusyBiscuit: Like I said, not a big deal.
With private methods we have full control over what goes into them anyway, it was just something I wanted to note, even if it is nitpicky.

src/main/java/be/seeseemelk/mockbukkit/PlayerList.java
```diff
@@ -99,16 +99,16 @@ public boolean isSomeoneOnline()
 	public List<Player> matchPlayer(@NotNull String name)
 	{
 		return onlinePlayers.stream().filter(
-				player -> player.getName().toLowerCase(Locale.ENGLISH).startsWith(name.toLowerCase(Locale.ENGLISH)))
-				.collect(Collectors.toList());
+		           player -> player.getName().toLowerCase(Locale.ENGLISH).startsWith(name.toLowerCase(Locale.ENGLISH)))
+		       .collect(Collectors.toList());
```
seeseemelk, line 103: Perhaps this one can be formatted in the same way as `getPlayerExact` right underneath here.

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -88,6 +91,11 @@ else if (!task.isCancelled())
 				tasks.add(task);
 			}
 		}
+		if (oldTasks.size() == 0 && asyncTasksQueued > 0)
+		{
+			Logger.getLogger(LOGGER_NAME).warning("Asynchronous Tasks may have been lost -> " + asyncTasksQueued);
```
seeseemelk, line 96: I'm not a big fan of this warning message. I can imagine that if you write a unit test, and see this in your logs, many would either never see it or be confused why this cryptic error message appears.

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -281,67 +304,91 @@ public boolean isQueued(int taskId)
...
+			{
+				name = task.getOwner().getName();
+			}
+			else
+			{
+				name = "";
+			}
+			Logger.getLogger(LOGGER_NAME).warning(
+			    name + "(" + task.getRunnable().getClass().getSimpleName()
+			    +  "): Scheduler is shutting down - why are you scheduling tasks - CANCELLED");
```
seeseemelk, line 385: I would prefer to make this an exception. A unit test should be correct. We shouldn't attempt to run code that is _somewhat_ correct, given that we are testing if code _is_ correct..

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
seeseemelk: I personally think `check*` is **always** a bad name for a method. A `check*` method presumbly does some calculation, and then what? Does it throw away the result? Does it return it? Does it perform some task? Execute some callbacks? Something else?

This method would perhaps be called more appropriately `assertSchedulable(Task)`.

src/test/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMockTest.java
```diff
@@ -112,6 +116,51 @@ public void runTaskTimer_ZeroDelay_DoesntExecuteTaskImmediately()
...
 	}
 
+	@Test
+	public void runScheduledTask_WhileScheduler_ShutDown()
+	{
+		AtomicInteger count = new AtomicInteger(0);
+		Runnable callback = () -> count.incrementAndGet();
+		scheduler.shutdown();
+		scheduler.runTaskTimer(null, callback, 0, 2L);
+		scheduler.performTicks(1L);
```
seeseemelk, line 126: Here you are shutting down the scheduler and then running your tasks. However, `shutdown` is normally only called by `MockBukkit.unmock()`. So I'm not entirely certain what the point is of this test.

seeseemelk (changes requested):
While this might work to fix some issues, it does introduces other bugs.

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
Narimm: Basically - the fact was that this is an implementation issue with the currently sheduler it should never happen in a real server.    There is no separate queue and task count they are the same.    So I can remove the log message.

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
Narimm: Bukkit does exactly the same thing if you cause a task to run after shutdown - its bad practice but not breaking code

Narimm: ie it doesnt fail but it doesnt schedule

src/test/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMockTest.java
Narimm: proving task dont run

Narimm: I will look later today

Narimm:
Closed as #183 will cover this until the scheduler gets a make over

PR #174, opened 2021-03-01
Use command map instead of command array

nowtryz:
# Description

I know a lot of command frameworks work with the `CommandMap` to dynamically introduce commands to bukkit. Though MockBukkit was only using its command array.
Hence this PR changes calls to `PluginManagerMock.getCommands` to `MockCommandMap.getCommand`.

# Checklist
- [x] Unit tests added.


TheBusyBiscuit (approved):
LGTM, also inline with the CB implementation.

Btw we may want to relocate the commandmap mock to the command package in the future

PR #178, opened 2021-03-02
fix simulate block break type is not correct

nowtryz:
# Description
When calling `simulateBlockPlace` on PlayerMock, the type of the broken was set after the event. Even though It is semantically correct, the event receives a block with a type different from the one provided to `simulateBlockPlace` causing code using this information to fail during tests.

The PR change is the behavior of this method by setting the type of the block before this event and then resetting it the previous one if the event got canceled.

Nevertheless, I have some hesitations if it is the best solution as `World#getBlock` would return the wrong type during the event propagation.

# Checklist
- [x] Unit tests added.


TheBusyBiscuit:
Any reason for why this is still a draft?

nowtryz:
This is a draft because of the following:
> Nevertheless, I have some hesitations if it is the best solution as World#getBlock would return the wrong type during the event propagation.

I didn't take time to work on it. You can either merge it as is, or maybe somebody can find a better option.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -230,10 +230,21 @@ public boolean simulateBlockPlace(Material material, Location location)
 		if (gamemode == GameMode.ADVENTURE || gamemode == GameMode.SPECTATOR)
 			return false;
 		Block block = location.getBlock();
-		BlockPlaceEvent event = new BlockPlaceEvent(block, null, null, null, this, true, null);
+		Material previousMaterial = block.getType();
+
+		block.setType(material); // swap material so event has the correct block information
+
+		BlockPlaceEvent event = new BlockPlaceEvent(block, null, null, getItemInHand(), this, true, null) {
```
```suggestion
		BlockPlaceEvent event = new BlockPlaceEvent(block, null, null, getItemInHand(), this, true, null)
		{
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -230,10 +230,21 @@ public boolean simulateBlockPlace(Material material, Location location)
...
+			@NotNull
+			@Override
+			public EquipmentSlot getHand() {
```
```suggestion
			public EquipmentSlot getHand()
			{
```

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1038,6 +1039,27 @@ public void testSetMaximumAir()
 		assertEquals(-10, player.getMaximumAir());
 	}
 
+	@Test
+	public void simulateBlockPlace_ValidBlock_EventShouldBeCalledWithCorrectType() {
```
seeseemelk:
```suggestion
	public void simulateBlockPlace_ValidBlock_EventShouldBeCalledWithCorrectType()
	{
```

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1038,6 +1039,27 @@ public void testSetMaximumAir()
...
+		Material newMaterial = Material.BEDROCK;
+		AtomicBoolean eventDispatched = new AtomicBoolean();
+		AtomicBoolean isTypeCorrect = new AtomicBoolean();
+		TestPlugin plugin = MockBukkit.load(TestPlugin.class);
+
+		server.getPluginManager().registerEvents(new Listener() {
```
seeseemelk:
```suggestion
		server.getPluginManager().registerEvents(new Listener()
		{
```

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1038,6 +1039,27 @@ public void testSetMaximumAir()
...
+			@EventHandler
+			public void onPlace(BlockPlaceEvent event) {
```
seeseemelk:
```suggestion
			public void onPlace(BlockPlaceEvent event)
			{
```

seeseemelk (approved):
Other than the markup things, it looks good to me. The small remaining issue you explained can always be fixed at a later point

Insprill:
Implemented in #290.

PR #180, opened 2021-03-04
Implement World#getChunk for Location and Block

Jikoo:
# Description
* Implement `World#getChunkAt(Location)` and `World#getChunkAt(Block)`
  * Fix `Block#getChunk`
* Fix `Chunk#getBlock` always returning a block in the column of the chunk's lowest block coordinate instead of proper local block

I threw the test in BlockMockTest because I discovered this trying to use `Block#getChunk`, but the argument could be made that it would fit better in `WorldMockTest` or `ChunkMockTest`.

# Checklist
- [x] Unit tests added.


TheBusyBiscuit (approved):
LGTM 👍🏻

PR #184, opened 2021-03-04
github: delete formatting workflow

seeseemelk:
# Description
This PR deletes the code formatting action. It didn't work as it couldn't push to `v1.16` anyway.

# Checklist
- [ ] Unit tests added.


Jikoo:
I'm not sure what settings you guys have here, but the action worked fine just now in 62611b6e6eaf3d6011a4a4721bc3e42c40159afd when I accidentally pushed the wrong branch to my fork.

TheBusyBiscuit:
> I'm not sure what settings you guys have here, but the action worked fine just now in [62611b6](https://github.com/seeseemelk/MockBukkit/commit/62611b6e6eaf3d6011a4a4721bc3e42c40159afd) when I accidentally pushed the wrong branch to my fork.

The problem is that this was configured to run **only** on the main branch.
Yet the main branch is a *protected* branch with limited push access, so the action will basically insta-fail on every single run.
For forks it will only run on *their main branch* which... it's not a good practice to be working on anyway and rather discouraged.

That's why this entire workflow is inherently flawed.
I wouldn't exactly delete it but rather re-purpose the configuration for it, so I'm gonna request some changes on this at a later point in time.

TheBusyBiscuit:
@seeseemelk I updated this branch with a **new workflow** I would like to propose.
This will make it run on forks-only and on any branch.
This way, it won't interefere with our branches and I think we can just trust each other to not mess up the formatting ;D
Well, you that is, I mostly work on my fork anyway.

Narimm:
I am just going to leave this here but I generally find its far better to include a checkstyle.xml if your not using a standard format eg Sun/Google that people working on the project can use to setup an autoformatter.   That way you don't rely on a automated code reformatted.  Or at least it wouldn't make major changes.

TheBusyBiscuit (approved):

PR #183, opened 2021-03-04
scheduler: properly cancel async tasks

seeseemelk:
# Description
This PR makes it so that when an asynchronous task gets cancelled, the number of queued asynchronous tasks get decrement accordingly.

# Checklist
The following items should be checked before the pull request can be merged.
- [x] Version number updated in `settings.gradle`.
- [x] Unit tests added.
- [x] Code follows existing code format.

PR #186, opened 2021-03-09
Feature/properly cancel async tasks 2

Narimm:
# Description
This extends the #183 Pull request in that it creats locks for the task queue to avoid CME's.   It also removes the integers that had been used to track queue size and uses the actual queue size filtered by the specifics of the call.

# Checklist
- [X] Version number updated in `settings.gradle`.
- [X] Unit tests added.


src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -1,34 +1,59 @@
...
-	private int asyncTasksQueued = 0;
+	private long holdExecutor = 60000;
+
+	private static Runnable runTask(ScheduledTask task) {
+		return () -> {
+			task.setRunning(true);
+			task.run();
+			task.setRunning(false);
+		};
+	}
```
nowtryz:
```suggestion
	private static Runnable wrapTask(ScheduledTask task) {
		return () -> {
			task.setRunning(true);
			task.run();
			task.setRunning(false);
		};
	}
```
I would consider `wrap` instead of `run` as it does not actually run the task

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -103,135 +121,166 @@ public void performTicks(long ticks)
...
+			if (System.currentTimeMillis() > (systemTime + holdExecutor)) {
+				// If a plugin has left a a runnable going and not cancelled it we could call this bad practice.
+				// we should now force interrupt all these runnables forcing them to throw Interrupted Exceptions.
+				// if they handle that
+				for (ScheduledTask task : scheduledTasks.getCurrentTaskList()) {
+					if (task.isRunning()) {
+						task.cancel();
+						cancelTask(task.getTaskId());
+						throw new RuntimeException("Forced Cancellation of task owned by "
+								+ task.getOwner().getName());
```
nowtryz: Personally, I use this method to wait until all tasks  are finished, throwing an exception here have a strange meaning, I would create a `forceShutdown` instead that could be called along with `unmock` which have more sense to me

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -366,50 +401,157 @@ public void run()
...
+	}
+
+	private static class TaskList {
+
+		private final Map<Integer, ScheduledTask> tasks;
+		private final ReadWriteLock taskLock;
+
+		private TaskList()
+		{
+			tasks = new HashMap<>();
```
nowtryz:
```suggestion
			tasks = new ConcurrentHashMap<>();
```
Consider using `ConcurrentHashMap` instead of `ReadWriteLock`s, in this case locks may slow down the process compared to `java.util.concurrent` API. Always avoid using low level API if possible

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
TheBusyBiscuit: ^ I agree

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -1,34 +1,59 @@
...
+	private static Runnable runTask(ScheduledTask task) {
+		return () -> {
+			task.setRunning(true);
+			task.run();
+			task.setRunning(false);
+		};
+	}
+
+
+	public void setHoldExecutor(long holdExecutor) {
```
Narimm:
```suggestion
	public void setShutdownTimeout(long timeout) {
```

Narimm (commented):
This method if set suitable large would mean the shutdown changes are as there were before.

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
Narimm: The timeout I added when set large enough means this Exception should not be thrown - if it is - even with a long timeout - then the user would want to adjust the plugin code to ensure any long running tasks are properly managed and closed before the Server/Scheduler is shutdown.

Narimm: This just means that the test will exit at some point rather than locking up for 8hrs, and the test will fail if it does do that as it should

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -1,34 +1,59 @@
...
+	private final TaskList scheduledTasks = new TaskList();
+	private final AtomicReference<Exception> asyncException = new AtomicReference<>();
 	private long currentTick = 0;
 	private int id = 0;
-	private List<ScheduledTask> tasks = new LinkedList<>();
-	private ExecutorService pool = Executors.newCachedThreadPool();
-	private AtomicInteger asyncTasksRunning = new AtomicInteger();
-	private AtomicReference<Exception> asyncException = new AtomicReference<>();
-	private int asyncTasksQueued = 0;
+	private long holdExecutor = 60000;
```
Narimm:
```suggestion
	private long executorTimeout = 60000;
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -366,50 +401,157 @@ public void run()
...
+
+	protected int getActiveRunningCount()
+	{
+		return pool.getActiveCount();
+	}
+
+	private static class TaskList {
+
+		private final Map<Integer, ScheduledTask> tasks;
+		private final ReadWriteLock taskLock;
```
Narimm:
```suggestion
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -366,50 +401,157 @@ public void run()
...
+		 * @return true on success.
+		 */
+		private boolean addTask(ScheduledTask task)
+		{
+			if (task == null) {
+				return false;
+			}
+			taskLock.writeLock().lock();
+			tasks.put(task.getTaskId(), task);
+			taskLock.writeLock().unlock();
```
Narimm:
```suggestion
			tasks.put(task.getTaskId(), task);
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -366,50 +401,157 @@ public void run()
...
+			if (tasks.size() == 0) {
+				return out;
+			}
+			taskLock.readLock().lock();
+			if (out.addAll(tasks.values())) {
+				taskLock.readLock().unlock();
+				return out;
+			} else {
+				taskLock.readLock().unlock();
+				throw new RuntimeException("Could not get current task list.");
```
Narimm:
```suggestion
			if (out.addAll(tasks.values())) {
				return out;
			} else {
				throw new RuntimeException("Could not get current task list.");
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -59,34 +89,22 @@ public long getCurrentTick()
...
-					pool.execute(task.getRunnable());
-					asyncTasksQueued--;
+		List<ScheduledTask> oldTasks = scheduledTasks.getCurrentTaskList();
+
+		for (ScheduledTask task : oldTasks) {
+			if (task.getScheduledTick() == currentTick && !task.isCancelled()) {
+				if (task.isSync()) {
+					runTask(task).run();
+				} else {
+					pool.submit(runTask(task));
```
Narimm:
```suggestion
					wrapTask(task).run();
				} else {
					pool.submit(wrapTask(task));
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -281,68 +323,61 @@ public boolean isQueued(int taskId)
...
 	}
 
 	@Override
-	public BukkitTask runTaskAsynchronously(Plugin plugin, Runnable task)
+	public @NotNull BukkitTask runTaskAsynchronously(@NotNull Plugin plugin, @NotNull Runnable task)
 	{
 		ScheduledTask scheduledTask = new ScheduledTask(id++, plugin, false, currentTick, new AsyncRunnable(task));
-		asyncTasksRunning.incrementAndGet();
-		pool.execute(scheduledTask.getRunnable());
+		pool.execute(runTask(scheduledTask));
```
Narimm:
```suggestion
		pool.execute(wrapTask(scheduledTask));
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -1,34 +1,59 @@
 package be.seeseemelk.mockbukkit.scheduler;
 
-import java.util.LinkedList;
+import java.util.ArrayList;
+import java.util.HashMap;
```
Narimm:
```suggestion
import java.util.concurrent.ConcurrentHashMap;
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -1,34 +1,59 @@
...
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
```
Narimm:
```suggestion

```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -1,34 +1,57 @@
...
+		return () -> {
+			task.setRunning(true);
+			task.run();
+			task.setRunning(false);
+		};
+	}
+
+
+	public void setShutdownTimeout(long timeout) {
+		this.holdExecutor = holdExecutor;
```
Narimm:
```suggestion
		this.executorTimeout = timeout;
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -103,135 +119,166 @@ public void performTicks(long ticks)
...
+		while (pool.getActiveCount() > 0) {
+			try {
 				Thread.sleep(10L);
 			}
 			catch (InterruptedException e)
 			{
 				Thread.currentThread().interrupt();
 				return;
 			}
+			if (System.currentTimeMillis() > (systemTime + holdExecutor)) {
```
Narimm:
```suggestion
			if (System.currentTimeMillis() > (systemTime + executorTimeout)) {
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -366,50 +399,152 @@ public void run()
...
+
+	private static class TaskList {
+
+		private final Map<Integer, ScheduledTask> tasks;
+		private final ReadWriteLock taskLock;
+
+		private TaskList()
+		{
+			tasks = new ConcurrentHashMap<>();
+			taskLock = new ReentrantReadWriteLock();
```
Narimm:
```suggestion
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -38,9 +61,14 @@ public void shutdown()
...
 		pool.shutdown();
-
 		if (asyncException.get() != null)
 			throw new AsyncTaskException(asyncException.get());
+		chatExecutor.shutdownNow();
+	}
+
+	public void executeAsyncChatEvent(AsyncPlayerChatEvent event)
+	{
+		chatExecutor.submit(() -> MockBukkit.getMock().getPluginManager().callEvent(event));
```
Narimm: I would note that I added this event so that Chat is not executed using the standard scheduler pool and cause incrementing of the task count. 

However, it can be reverted in lieu of #187

TheBusyBiscuit: Oh my, I didn't even see this change.
I think running this on the executor's thread pool directly sounds like a good idea since this is just supposed to run on a different Thread directly, not necessarily going to have to go through all the scheduler logic first.
Especially when we consider that the scheduling of that event is only supposed to mock the multi-threaded nature of the chat protocol.

The only reason I even touched that part in #187 was to rectify the behaviour when events get triggered on the wrong Thread, since an exception should be thrown in that case.
We recently had this being the case, the unit test passed but when the project was put on a server, it spit out tons of exceptions due to events being scheduled on the wrong Thread.
So this fix could have saved us some trouble.

I think running this on the Executor directly is a good idea but I would generalize this to accept **any async event**.
Then we could have the `PluginManagerMock` run this in the background.
If you want, you could make a PR to my fork.
But I was gonna split that behemoth up into multiple pull requests anyway, so I could also just make a branch on this repo dedicated for this which you could then PR to.

Narimm: So the scheduler as it stands has 2 executors 
1. The Task Executor which handles all async task - the pool is exposed in this scheduler to allow monitoring of the actively running tasks.
2. The Chat executor - which is really fire and forget its purely internal for testing - and not representative of anything that runs on a live server.  Chat on live CAN be executed async but also sometimes isn't - in this case we artificially execute it both ways ie from the PlayerMock.class 
```java
@Override
	public void chat(String msg)
	{
		AsyncPlayerChatEvent eventAsync = new AsyncPlayerChatEvent(false, this, msg,
		        new HashSet<>(Bukkit.getOnlinePlayers()));
		PlayerChatEvent eventSync = new PlayerChatEvent(this, msg);
		MockBukkit.getMock().getScheduler().executeAsyncChatEvent(eventAsync);
		Bukkit.getPluginManager().callEvent(eventSync);
	}
```

Narimm: The only reason I put the Chat Executor in the scheduler mock is I could ensure both pools were shutdown at the same time.

TheBusyBiscuit: Ohh I see, forgive me I haven't had a chance to actually look at this code in detail yet.
Yeah that makes a lot of sense, personally I would probably have a dedicated `AsyncEventSchedulerMock` here though which we could just have as yet another final field in the Scheduler. But that may just be my personal preference.

Also note that any Event can be asynchronous, not just the `AsyncPlayerChatEvent`, even though it is the most prominent implementation (perhaps even the only one in vanilla Bukkit). So the naming should definitely reflect that as "Chat" could be misleading then.

I should probably properly look into this pull request next week (busy week right now).
But currently I am thinking it would be best to seperate out the async event handling from both of our pull requests and work on one dedicated PR for that together, since we are both addressing different problems that are related to this and a combined solution would probably be for the better.

Narimm:
Yep fair enough - I will revert the changes but I would note that due to
the chat events currently using the common async task pool they
artificially inflate the active tasks.

Ben Charlton

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
seeseemelk: I'm not sure that it is a good idea to execute the event twice. That means any event handlers would also get executed twice, which I don't think is something we want to happen.

Narimm: That was the existing behaviour

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -68,6 +68,7 @@
 import be.seeseemelk.mockbukkit.entity.ItemEntityMock;
 import be.seeseemelk.mockbukkit.entity.ZombieMock;
 import be.seeseemelk.mockbukkit.metadata.MetadataTable;
+import org.jetbrains.annotations.Nullable;
```
seeseemelk: This import was already imported. No need to import it twice
```suggestion
```

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -426,4 +472,115 @@ public void runTaskTimerAsynchronously(@NotNull Plugin plugin, @NotNull Consumer
...
+		 */
+		/*@Deprecated
+		protected ScheduledTask removeTask(int taskID)
+		{
+			taskLock.writeLock().lock();
+			ScheduledTask removed = tasks.remove(taskID);
+			taskLock.writeLock().unlock();
+			return removed;
+
+		}*/
```
seeseemelk: Remove commented out code

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -426,4 +472,115 @@ public void runTaskTimerAsynchronously(@NotNull Plugin plugin, @NotNull Consumer
...
+				return out;
+			}
+			if (out.addAll(tasks.values()))
+			{
+				return out;
+			}
+			else
+			{
+				throw new RuntimeException("Could not get current task list.");
+			}
```
seeseemelk: Is this ever reached? The only reason I can imagine `addAll` would ever return `false` is if tasks is empty, but you're already checking for this condition earlier in this function.

seeseemelk (changes requested):
Some small changes, but overall it looks pretty good

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
Narimm: Should be resolved by 69b612b7f159be6b166906cef5479244eff729b4

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
Narimm: Should be resolved by 69b612b7f159be6b166906cef5479244eff729b4

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
Narimm: Should be resolved by 69b612b7f159be6b166906cef5479244eff729b4

seeseemelk (approved):

PR #187, opened 2021-03-09
Some fixes and refactoring, also updated to the latest Spigot-API

TheBusyBiscuit:
# Description
* Added overloaded `World#dropItem` method with `Consumer<Item>` parameter
* Added overloaded `World#dropItem` method with `Consumer<Item>` parameter
* Added parameter validation for item drop methods
* Added parameter validation for Inventory mock
* `ServerMock` not extends `Server.Spigot` (as we discussed earlier)
* ~~AsyncChatEvent is now properly marked as asynchronous~~
* ~~Added `PluginManagerMock#callEventAsynchronously` to help run events on a different Thread easily with this helper method.~~
* ~~Added proper Thread validations when you try to schedule events on the wrong Thread (like Bukkit does)~~
* Added `isRespawnAnchor` parameter to `PlayerRespawnEvent`.
* Relocated some suppressed warnings annotations to only the affected methods, they were hiding a lot of deprecated stuff we should have fixed earlier. Downside: I had to remove the imports and reference the class by its full classpath.
* Added support for Sounds to be played via their String reference.
* Seperated sound assertions into a `SoundReceiver` interface
* Removed redundant `Spigot` calls for entity implementations
* Fixed `EntityEquipmentMock#getArmorContents()` order
* Implemented `EntityEquipmentMock#setArmorContents(...)`
* Also removed some wildcard imports

# Checklist
- [ ] Unit tests added.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,5 @@
 gradle.ext.apiVersion = '1.16'
 gradle.ext.apiVersionFull = '1.16.5-R0.1-SNAPSHOT'
-gradle.ext.version = '0.31.1'
+gradle.ext.version = '0.32.0'
 
 rootProject.name = 'MockBukkit-' + gradle.ext.apiVersion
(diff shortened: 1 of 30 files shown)


TheBusyBiscuit:
Yeah... I may need to split this colossus up into multiple PRs...

PR #188, opened 2021-03-11
Fixed asynchronous event scheduling

TheBusyBiscuit:
# Description
This pull requests addresses some inconsistent or wrong behaviours within the way the `AsyncPlayerChatEvent` is handled.
This PR is a combined effort from #187 and #186.

@Narimm In `PluginManagerMock` (line 419) you can find a reference to the method from `BukkitSchedulerMock` that will be your part. Feel free to rename, refactor or restructure this to your liking, it is just a skeleton placeholder from my part for now.
Simply make a new pull request and select the `fix/async-events` branch as the target and I will merge the changes into this PR.

Also note that the compilation will fail until the newly added Spigot methods are added (see #187)

# Checklist
The following items should be checked before the pull request can be merged.
- [ ] Version number updated in `settings.gradle`.
- [ ] Unit tests added.
- [x] Code follows existing code format.

PR #189, opened 2021-03-11
Add new aysnc event executor

Narimm:
# Description
Fix Nullability notations
Fix RespawnEvent deprecation
Move from Array.asList() to Collections.singleton() - Immutable and not backed by unnecessary array.

# Checklist
- [ ] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
@@ -11,6 +11,8 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.logging.Logger;
 
+import be.seeseemelk.mockbukkit.MockBukkit;
+import org.bukkit.event.Event;
 import org.bukkit.plugin.Plugin;
 import org.bukkit.scheduler.BukkitRunnable;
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -38,11 +43,16 @@ public void shutdown()
...
 		waitAsyncTasksFinished();
 		pool.shutdown();
-
+		asyncEventExecutor.shutdownNow();
 		if (asyncException.get() != null)
 			throw new AsyncTaskException(asyncException.get());
 	}
 
+	public Future<?> scheduleAsyncEventCall(Event event)
+	{
```
TheBusyBiscuit: I would suggest adding annotations here as well and even validating the parameter.

```suggestion
	public @NotNull Future<?> scheduleAsyncEventCall(@NotNull Event event)
	{
	    Validate.notNull(event, "Cannot schedule an Event that is null!");
```

TheBusyBiscuit (changes requested):
LGTM!

Just a small suggestion I'd like to make.

TheBusyBiscuit (approved):
Since I haven't heard back from you I went ahead and committed my requested change myself and merged this into the branch of #188

PR #191, opened 2021-03-15
[FIX] getVersion()

0ddlyoko:
# Description
Testing with MockBukkit & ProtocolLib doesn't work because `Server#getVersion()` always returns 0.1.0 and ProtocolLib use this method to retrieve the [version of Minecraft used](https://github.com/dmulloy2/ProtocolLib/blob/97972acee84ec73c57866ed1d7edbda20f273a40/src/main/java/com/comphenix/protocol/utility/MinecraftVersion.java#L138)
This fix fixes the issue

# Checklist
- [ ] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -426,7 +426,7 @@ public class ServerMock implements Server
 	@Override
 	public String getVersion()
 	{
-		return "0.1.0";
+		return BUKKIT_VERSION;
 	}
 
 	@Override
@@ -715,7 +715,7 @@ public class ServerMock implements Server
 
 		for (Player player : players)
 		{
-			if (player.hasPermission(permission)) 
+			if (player.hasPermission(permission))
 			{
 				player.sendMessage(message);
 				count++;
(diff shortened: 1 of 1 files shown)


src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -426,7 +426,7 @@ public String getName()
 	@Override
 	public String getVersion()
 	{
-		return "0.1.0";
+		return BUKKIT_VERSION;
```
TheBusyBiscuit: That's a hard pass from me.
This version does not follow Bukkit's convention, see my comment here:

https://github.com/seeseemelk/MockBukkit/pull/161/files/ddeab6d986b9976cf9a2c465a74499bac213a140#diff-763d283381d2bc7fcc200b07f44349baea512d37c0bb809b7c8b8e087d7eb1b9

TheBusyBiscuit (changes requested):
Also please be truthful about what you check on the todo list.
I see you have checked that you added Unit Tests but there were no Unit tests to be found.

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
0ddlyoko: Like now ?

TheBusyBiscuit (approved):
If we were worried about micro-optimizations we could just make it a field and return that.
But I am more than fine with this.

The version number could have been bumped too but we are currently blocked by updating to the latest Spigot API anyway.

TheBusyBiscuit:
Also, I cannot merge this due to the build check not completing yet.
GH Actions servers are currently experiencing some issues, so it may resolve itself by tomorrow.

PR #192, opened 2021-03-26
Add move player instantly function.

### LINKED ISSUE #110: Implement player movement
MockBukkit currently has no way of simulating a player moving around the world, even though many plugins listen to movement related events.

A PR fixing this issue should:
 - Add methods to `PlayerMock` to allow a test script to move the player around.
 - When the player is moving, the right events should be fired.
 - If the event gets cancelled, the player should not move.

Collision detection with the world and with other entities would be nice, but it is not required at this point in time.

wuza8:
# Description
Added method moveInstantly simply by moving player from place to place by teleporting him and executing move event. If the event will be canceled, it will not move player.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -54,6 +54,7 @@ import org.bukkit.event.inventory.InventoryCloseEvent;
 import org.bukkit.event.inventory.InventoryType;
 import org.bukkit.event.player.AsyncPlayerChatEvent;
 import org.bukkit.event.player.PlayerChatEvent;
+import org.bukkit.event.player.PlayerMoveEvent;
 import org.bukkit.event.player.PlayerLevelChangeEvent;
 import org.bukkit.event.player.PlayerRespawnEvent;
 import org.bukkit.inventory.EntityEquipment;
@@ -264,6 +265,13 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 		setHealth(getAttribute(Attribute.GENERIC_MAX_HEALTH).getValue());
 		setLocation(event.getRespawnLocation().clone());
 		alive = true;
+	}
+
+	public void moveInstantly(Location moveLocation) {
+		PlayerMoveEvent event = new PlayerMoveEvent(this, this.getLocation(), moveLocation);
+		Bukkit.getPluginManager().callEvent(event);
+		if(event.isCancelled() == false)
+			this.setLocation(moveLocation.clone());
 	}
 
 	@Override
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -267,7 +267,12 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 		alive = true;
 	}
 
-	public void simulatePlayerMove(@NotNull Location moveLocation)
+	/**
+	 * This method moves player instantly with respect to PlayerMoveEvent
+	 *
+	 * @param moveLocation Location to move player to
+	 */
+	public void fakeMove(@NotNull Location moveLocation)
 	{
 		PlayerMoveEvent event = new PlayerMoveEvent(this, this.getLocation(), moveLocation);
 		Bukkit.getPluginManager().callEvent(event);
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -272,7 +272,7 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 		PlayerMoveEvent event = new PlayerMoveEvent(this, this.getLocation(), moveLocation);
 		Bukkit.getPluginManager().callEvent(event);
 		if(event.isCancelled() == false)
-			this.setLocation(moveLocation.clone());
+			this.setLocation(event.getTo());
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -266,6 +267,13 @@ public void respawn()
 		alive = true;
 	}
 
+	public void moveInstantly(Location moveLocation) {
```
TheBusyBiscuit: Please follow the existing convention of method names that do these sort of things.
I would suggest `simulatePlayerMove` for this.
Also please follow the `Allman` brackets style and add annotations which indicate nullability where possible.

```suggestion
	public void simulatePlayerMove(@NotNull Location moveLocation)
	{
```

TheBusyBiscuit: Javadocs for this method would be a nice thing to have too.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -266,6 +267,13 @@ public void respawn()
...
+		PlayerMoveEvent event = new PlayerMoveEvent(this, this.getLocation(), moveLocation);
+		Bukkit.getPluginManager().callEvent(event);
+		if(event.isCancelled() == false)
```
TheBusyBiscuit: Comparing a boolean to a boolean to get a boolean is redundant.

```suggestion
		if (!event.isCancelled())
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -266,6 +267,13 @@ public void respawn()
...
+			this.setLocation(moveLocation.clone());
```
TheBusyBiscuit: The `PlayerMoveEvent` can manipulate the destination, so we absolutely need to respect this modification.

```suggestion
			this.setLocation(event.getTo());
```

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1059,4 +1060,38 @@ public void testSimulateBlockPlaceInvalid()
 		player.setGameMode(originalGM);
 		assertFalse(worked);
 	}
+	@Test
```
TheBusyBiscuit: Minor styling remark
```suggestion
	}

	@Test
```

TheBusyBiscuit (changes requested):

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
wuza8: I really didn't think about it, big brain!

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
wuza8: I'm not sure about this name now, because this method actually doesn't simulate anything. I will name it for now "fakeMove", because it's not actually that every tick it moves player to position, it just teleports him to destination instantly. Javadocks will be added.

wuza8: I've added JavaDocs but idk if that's enough, about that "fakeMove" name, maybe there's some better name... I don't want to name it "move" for the same reason why I think it shouldn't prefix with simulate. What do you think?

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -267,7 +267,12 @@ public void respawn()
 		alive = true;
 	}
 
-	public void simulatePlayerMove(@NotNull Location moveLocation)
+	/**
+	 * This method moves player instantly with respect to PlayerMoveEvent
+	 *
+	 * @param moveLocation Location to move player to
+	 */
+	public void fakeMove(@NotNull Location moveLocation)
```
wuza8: New JavaDoc and method name

TheBusyBiscuit: 👍🏻 for the docs.
But the name is still not quite right. `fakeMove` is very much misleading since that method **does** move the player, the movement is not faked since the Player is actually moved, just virtually, hence why `simulatePlayerMove(...)` would be the most appropriate and also inline with the other `simulate...()` methods naming convention.

wuza8: Ok, I will change that to simulatePlayerMove()

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -272,7 +272,7 @@ public void respawn()
 	 *
 	 * @param moveLocation Location to move player to
 	 */
-	public void fakeMove(@NotNull Location moveLocation)
+	public void simulatePlayerMove(@NotNull Location moveLocation)
```
wuza8, line 275: It's changed as suggested.

TheBusyBiscuit (approved):
Looks good to me,

perhaps the Unit Tests could still be improved somehow but I don't have anything particular in mind at this moment.
So: 👍🏻

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1059,4 +1060,31 @@ public void testSimulateBlockPlaceInvalid()
...
+			{
+				event.setCancelled(true);
+			}
+		}, plugin);
+		World world = server.addSimpleWorld("world");
+		player.setLocation(new Location(world, 0, 0, 0));
+		player.simulatePlayerMove(new Location(world, 10,0,0));
+		server.getPluginManager().assertEventFired(PlayerMoveEvent.class);
+		assertTrue(player.getLocation().getX() == 0.0);
+	}
```
wuza8: Made those tests a little cleaner now

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1059,4 +1060,31 @@ public void testSimulateBlockPlaceInvalid()
...
 }
```
wuza8, line 1090: Cleaned the tests and changed their names

TheBusyBiscuit (approved):
Still looking good to me.

Thanks for going the extra step to update the unit tests too!

PR #194, opened 2021-03-31
Implemented server settings getters

ReinWD:
# Description

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -113,7 +113,6 @@ public class ServerMock extends Server.Spigot implements Server
 	private static final String BUKKIT_VERSION = "1.16.5";
 	private static final String JOIN_MESSAGE = "%s has joined the server.";
 	private static final String MOTD = "A Minecraft Server";
-	private static final String SHUTDOWN_MESSAGE = "Shutting down server...";
 
 	private final Logger logger = Logger.getLogger("ServerMock");
 	private final Thread mainThread = Thread.currentThread();
@@ -136,6 +135,34 @@ public class ServerMock extends Server.Spigot implements Server
 	private WarningState warningState = WarningState.DEFAULT;
 	private GameMode defaultGameMode = GameMode.SURVIVAL;
 	private ConsoleCommandSender consoleSender;
+
+	//===BEGIN bukkit.yml fields getters support
+	//maybe configurable in the future.
+	//todo: make this section configurable
+	//default values from https://bukkit.fandom.com/wiki/Bukkit.yml
+
+	//settings
+	private boolean allowEnd = true;
+	private String updateFolder = "update";
+	//default worldContainer is current working directory (according to the wiki)
+	private File worldContainer = new File("");
+	private long connectionThrottle = 4000;
+	private String shutdownMessage = "Server closed";
+
+	//spawn-limits
+	private int monsterSpawnLimit = 70;
+	private int animalSpawnLimit = 10;
+	private int waterAnimalSpawnLimit = 15;
+	private int waterAmbientSpawnLimit = 20;
+	private int ambientSpawnLimit = 15	;
+
+	//ticks-per
+	private int ticksPerAnimalSpawns = 400;
+	private int ticksPerMonsterSpawns = 1;
+	private int ticksPerWaterSpawns = 1;
+	private int ticksPerWaterAmbientSpawns = 1;
+	private int ticksPerAmbientSpawns = 1;
+	//===END bukkit.yml fields getters support
 
 	public ServerMock()
 	{
@@ -881,8 +908,7 @@ public class ServerMock extends Server.Spigot implements Server
 	@Override
 	public boolean getAllowEnd()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return allowEnd;
 	}
 
 	@Override
@@ -923,36 +949,31 @@ public class ServerMock extends Server.Spigot implements Server
 	@Override
 	public String getUpdateFolder()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return updateFolder;
 	}
 
 	@Override
 	public File getUpdateFolderFile()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return new File(getUpdateFolder());
 	}
 
 	@Override
 	public long getConnectionThrottle()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return connectionThrottle;
 	}
 
 	@Override
 	public int getTicksPerAnimalSpawns()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return ticksPerAnimalSpawns;
 	}
 
 	@Override
 	public int getTicksPerMonsterSpawns()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return ticksPerMonsterSpawns;
 	}
 
 	@Override
@@ -1097,8 +1118,7 @@ public class ServerMock extends Server.Spigot implements Server
 	@Override
 	public File getWorldContainer()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return worldContainer;
 	}
 
 	@Override
@@ -1118,29 +1138,25 @@ public class ServerMock extends Server.Spigot implements Server
 	@Override
 	public int getMonsterSpawnLimit()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return monsterSpawnLimit;
 	}
 
 	@Override
 	public int getAnimalSpawnLimit()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return animalSpawnLimit;
 	}
 
 	@Override
 	public int getWaterAnimalSpawnLimit()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return waterAnimalSpawnLimit;
 	}
 
 	@Override
 	public int getAmbientSpawnLimit()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return ambientSpawnLimit;
 	}
 
 	@Override
@@ -1158,7 +1174,7 @@ public class ServerMock extends Server.Spigot implements Server
 	@Override
 	public String getShutdownMessage()
 	{
-		return SHUTDOWN_MESSAGE;
+		return shutdownMessage;
 	}
 
 	@Override
@@ -1460,15 +1476,13 @@ public class ServerMock extends Server.Spigot implements Server
 	@Override
 	public int getTicksPerWaterSpawns()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return ticksPerWaterSpawns;
 	}
 
 	@Override
 	public int getTicksPerAmbientSpawns()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return ticksPerAmbientSpawns;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


ReinWD:
may implement fields getters in bukkit.yml later.
this may worth a version bump.

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -544,6 +545,14 @@ public void testDefaultPotionEffects()
...
 	}
 
+	@Test
+	public void testWarningStateAndShutdownMessage(){
+		final String SHUTDOWN_MESSAGE = "Shutting down server...";
+		Warning.WarningState defaultWarningState = Warning.WarningState.DEFAULT;
+
+		assertEquals(defaultWarningState, server.getWarningState());
+		assertEquals(SHUTDOWN_MESSAGE, server.getShutdownMessage());
+	}
```
TheBusyBiscuit: I don't think this Unit Test is really needed here.
In general Getter-testing is sometimes frowned upon, especially when they are constants since messing these up is not really likely to happen. And if the values should diverge one day, then that was probably intentional.

So to me this Unit Test is more of a maintenance burden than really helping to ensure that MockBukkit runs flawlessly.

TheBusyBiscuit (commented):
Looking good to me, just the Unit Test seems a bit unneeded here since it compares a constant with a constant and introduces a bit of a maintenance hurdle when changing these values.

TheBusyBiscuit:
Also, it looks like you may have forgotten to bump the version number

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
ReinWD: thanks for your suggestions. i'll leave it blank until new features come in.

ReinWD:
> Also, it looks like you may have forgotten to bump the version number

I think it's not worthy to bump a minor version for several lines 
so I made more changes & bumped it :p

ReinWD:
I implemented configuration getters in bukkit.yml.
default values are from [official bukkit wiki](https://bukkit.fandom.com/wiki/Bukkit.yml).

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -19,6 +19,7 @@
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.bukkit.Material;
+import org.bukkit.Warning;
```
TheBusyBiscuit: unused import

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -132,9 +132,38 @@
...
+	private int waterAmbientSpawnLimit = 20;
+	private int ambientSpawnLimit = 15	;
+
+	//ticks-per
+	private int ticksPerAnimalSpawns = 400;
+	private int ticksPerMonsterSpawns = 1;
+	private int ticksPerWaterSpawns = 1;
+	private int ticksPerWaterAmbientSpawns = 1;
+	private int ticksPerAmbientSpawns = 1;
+	//===END bukkit.yml fields getters support
```
TheBusyBiscuit: I think having all these fields here pollutes the `ServerMock` class a bit too much.
Ideally these should all have corresponding setters too, in case someone needs to test their plugin under different environments.

With only two fields added I wouldn't have bat an eye but this amount of extra settings needs a cleaner solution in my opinion.
So if you are gonna add all of these settings, then I'd reckon a dedicated `ServerSettingsMock` class would be most appropriate, this would clean up the `ServerMock` class a bit, we could add setters in there and it could also benefit performance since this class is initialized on almost every test, so we could only load the settings into memory as necessary.
The getters in `ServerMock` could then just redirect like this:

```java
public Something getSomething()
{
    return getServerSettings().getSomething();
}
```

And then we would have a  method like this:
```java
public @NotNull ServerSettingsMock getServerSettings()
{
    if (serverSettings = null)
    {
        serverSettings = new ServerSettingsMock(this);
    }

    return serverSettings;
}
```


We could also add existing constants in there, like the MOTD and stuff.

TheBusyBiscuit (changes requested):

TheBusyBiscuit:
> I think it's not worthy to bump a minor version for several lines
> so I made more changes & bumped it :p

It is.
MockBukkit follows the semver (semantic versioning) convention, any new release should have a uniquely identifying version number since we no longer follow a SNAPSHOT versioning scheme.
Small additions should be a minor bump to indicate that there was a minor update.

Also, for future reference:
It isn generally not a good idea to expand your scope this drastically mid-way through a PR.
Plan ahead what you want to contribute and stick to that, unless changes were explicitly requested.
If you want to add additional stuff, it is most often best to make a seperate PR for it. This also makes it a lot easier for reviewers to go through these changes.

ReinWD:
Thanks a lot.
I extracted those fields from ServerMock into ServerSettingsMock with the same level of package.
Feel free to rename or move package.

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -157,6 +160,14 @@ public ServerMock()
 		logger.setLevel(Level.ALL);
 	}
 
+	public ServerSettingsMock getServerSettings(){
```
TheBusyBiscuit, line 170: Please follow MockBukkit's bracket style.
It's also a good practice to add nullabity annotations (NotNull / Nullable) where possible to indicate whether a method can or cannot return null values.

```suggestion
	public @NotNull ServerSettingsMock getServerSettings()
	{
```

src/main/java/be/seeseemelk/mockbukkit/ServerSettingsMock.java
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * configuration sections for {@link ServerMock}
+ * default values from https://bukkit.fandom.com/wiki/Bukkit.yml
+ * @author ReinWD
+ *
+ * @see ServerMock
+ */
+public class ServerSettingsMock {
+	//maybe configurable in the future.
+	//todo: make this section configurable
```
TheBusyBiscuit: Isn't this todo resolved now?

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
TheBusyBiscuit: Unused import.

src/main/java/be/seeseemelk/mockbukkit/ServerSettingsMock.java
```diff
@@ -0,0 +1,157 @@
...
+		this.ticksPerWaterAmbientSpawns = ticksPerWaterAmbientSpawns;
+	}
+
+	public int getTicksPerAmbientSpawns() {
+		return ticksPerAmbientSpawns;
+	}
+
+	public void setTicksPerAmbientSpawns(int ticksPerAmbientSpawns) {
+		this.ticksPerAmbientSpawns = ticksPerAmbientSpawns;
+	}
```
TheBusyBiscuit: Can you please update these methods to follow the bracket style of MockBukkit?

Also some parameter validation would be very much welcome.
or any non-primitives the parameters should have a `@NotNull` annotation and best would also be an additional
`Validate.notNull(x, "... must not be null");` check in the first line of the method.
For primitives, such as ints a simple `Validate.isTrue(x > 0), "... cannot be negative");` would be awesome too.

But we can always add these later on ourselves, just wanted to make you aware of these practices.
Those are also some examples, we would likely need to double-check with the bukkit.yml documentation to find out which values are valid and which are not.

TheBusyBiscuit (changes requested):

src/main/java/be/seeseemelk/mockbukkit/ServerSettingsMock.java
ReinWD: changed braket style
added @NotNull check for non-primitive getters.

src/main/java/be/seeseemelk/mockbukkit/ServerSettingsMock.java
ReinWD: implemented in latest commit df0682e

src/main/java/be/seeseemelk/mockbukkit/ServerSettingsMock.java
```diff
@@ -0,0 +1,313 @@
...
+import org.bukkit.configuration.InvalidConfigurationException;
+import org.bukkit.configuration.file.YamlConfiguration;
+import org.jetbrains.annotations.NotNull;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * configuration sections for {@link ServerMock}
+ * default values from https://bukkit.fandom.com/wiki/Bukkit.yml
```
seeseemelk, line 15:
```suggestion
 * Configuration sections for {@link ServerMock}
 * Default values from https://bukkit.fandom.com/wiki/Bukkit.yml
```

src/main/java/be/seeseemelk/mockbukkit/ServerSettingsMock.java
```diff
@@ -0,0 +1,313 @@
...
+				.append(ticksPerMonsterSpawns)
+				.append(ticksPerWaterSpawns)
+				.append(ticksPerWaterAmbientSpawns)
+				.append(ticksPerAmbientSpawns).toHashCode();
+	}
+
+	public boolean isAllowEnd()
+    {
+		return allowEnd;
+	}
```
seeseemelk, line 156: The indentation here and in the rest of the file is a bit all over the place. Mixing tabs and spaces perhaps?

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -544,6 +543,72 @@ public void testDefaultPotionEffects()
...
+			serverSettings.load(file);
+			assertEquals(serverSettings, dummy2);
+		}
+		catch (IOException | InvalidConfigurationException e)
+		{
+			fail(e.getMessage());
+		}
+	}
+
+	//fill another value for server setting for serialization test
```
seeseemelk, line 592: Why is this a normal comment and not a Javadoc comment?

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -544,6 +543,72 @@ public void testDefaultPotionEffects()
...
+		//test on clean start
+		try
+		{
+			serverSettings = server.getServerSettings();
+		}
+		catch (Exception e)
+		{
+			fail(e.getMessage());
+		}
+		assert serverSettings != null;
```
seeseemelk, line 573: This should probably be an `assertNotNull` or something similar.

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -544,6 +543,72 @@ public void testDefaultPotionEffects()
...
+		assertEquals(serverSettings, dummy);
+
+		//test on changing props
+		ServerSettingsMock dummy2 = new ServerSettingsMock();
```
seeseemelk, line 577: Can this not be done in a separate test? Testing both these states in a single test makes this test a little busy.

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -544,6 +543,72 @@ public void testDefaultPotionEffects()
...
+		File file = null;
+		try {
+			Path tempFile = Files.createTempFile(SERVER_SETTING_FILE, String.valueOf(Math.random()));
+			file = new File(tempFile.toUri());
+			file.deleteOnExit();
+		}
+		catch (Exception e)
+		{
+			fail(e.getMessage());
+		}
```
seeseemelk, line 559: You don't need a try-catch here. You can simply define the test method with `throws Exception`. JUnit will detect an exception as a failed test, and it will pretty print the exception. It also makes the test much cleaner.

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -157,6 +166,42 @@ public ServerMock()
...
+				else
+				{
+					file = new File(BUKKIT_SETTING_FILE);
+				}
+
+				if (!file.exists())
+				{
+					Path tempFile = Files.createTempFile(file.getName(), String.valueOf(Math.random()));
+					file = new File(tempFile.toUri());
+					file.deleteOnExit();
```
seeseemelk, line 191: Could we just use an in-memory configuration instead of creating a new file every time?

seeseemelk (changes requested):

thelooter:
As there was no Action on this for over a year, we're gonna close this

PR #195, opened 2021-04-01
Implement flight, sprint, and added events

Seggan:
# Description
Added player sprinting and flying, added sprinting, sneaking, and flying simulation methods

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -949,7 +949,8 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 	{
 		PlayerToggleSneakEvent event = new PlayerToggleSneakEvent(this, sneaking);
 		Bukkit.getPluginManager().callEvent(event);
-		if (!event.isCancelled()) {
+		if (!event.isCancelled())
+		{
 			this.sneaking = event.isSneaking();
 		}
 	}
@@ -965,7 +966,8 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 	{
 		PlayerToggleSprintEvent event = new PlayerToggleSprintEvent(this, sprinting);
 		Bukkit.getPluginManager().callEvent(event);
-		if (!event.isCancelled()) {
+		if (!event.isCancelled())
+		{
 			this.sprinting = event.isSprinting();
 		}
 	}
@@ -1499,7 +1501,8 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 	{
 		PlayerToggleFlightEvent event = new PlayerToggleFlightEvent(this, value);
 		Bukkit.getPluginManager().callEvent(event);
-		if (!event.isCancelled()) {
+		if (!event.isCancelled())
+		{
 			this.flying = event.isFlying();
 		}
 	}
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -961,12 +961,13 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 	@Override
 	public void setSneaking(boolean sneaking)
 	{
-		PlayerToggleSneakEvent event = new PlayerToggleSneakEvent(this, sneaking);
+		this.sneaking = sneaking;
+	}
+
+	public void simulateSneak(boolean sneak)
+	{
+		PlayerToggleSneakEvent event = new PlayerToggleSneakEvent(this, sneak);
 		Bukkit.getPluginManager().callEvent(event);
-		if (!event.isCancelled())
-		{
-			this.sneaking = event.isSneaking();
-		}
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1494,15 +1507,18 @@ public boolean canSee(Player player)
...
 	public void setFlying(boolean value)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		PlayerToggleFlightEvent event = new PlayerToggleFlightEvent(this, value);
+		Bukkit.getPluginManager().callEvent(event);
+		if (!event.isCancelled())
+		{
+			this.flying = event.isFlying();
+		}
```
TheBusyBiscuit: This is an erroneous implementation.
CraftBukkit does not call such an event when invoking this method.
Setters only modify the state whereas any PlayerEvent is the result of the actions from an actual Player, you would need a simulation method for this. The setter itself doesn't call an event.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -956,21 +961,29 @@ public boolean isSneaking()
...
 	public void setSprinting(boolean sprinting)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		PlayerToggleSprintEvent event = new PlayerToggleSprintEvent(this, sprinting);
+		Bukkit.getPluginManager().callEvent(event);
+		if (!event.isCancelled())
+		{
+			this.sprinting = event.isSprinting();
+		}
```
TheBusyBiscuit: This is an erroneous implementation.
CraftBukkit does not call such an event when invoking this method.
Setters only modify the state whereas any PlayerEvent is the result of the actions from an actual Player, you would need a simulation method for this. The setter itself doesn't call an event.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -956,21 +961,29 @@ public boolean isSneaking()
 	@Override
 	public void setSneaking(boolean sneaking)
 	{
-		this.sneaking = sneaking;
+		PlayerToggleSneakEvent event = new PlayerToggleSneakEvent(this, sneaking);
+		Bukkit.getPluginManager().callEvent(event);
+		if (!event.isCancelled())
+		{
+			this.sneaking = event.isSneaking();
+		}
```
TheBusyBiscuit: This is an erroneous implementation.
CraftBukkit does not call such an event when invoking this method.
Setters only modify the state whereas any PlayerEvent is the result of the actions from an actual Player, you would need a simulation method for this. The setter itself doesn't call an event.

TheBusyBiscuit (changes requested):
These implementations are completely wrong and not inline with CraftBukkit.
Calling events here will lead to unintended results and possible infinite feedback loops.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1494,15 +1509,19 @@ public boolean canSee(Player player)
...
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		this.flying = value;
+	}
+
+	public void simulateToggleFlight(boolean fly)
+	{
+		PlayerToggleFlightEvent event = new PlayerToggleFlightEvent(this, fly);
+		Bukkit.getPluginManager().callEvent(event);
```
TheBusyBiscuit, line 1532: The method should respect the Event outcome and set the `flying` state accordingly.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -959,18 +964,28 @@ public void setSneaking(boolean sneaking)
...
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		this.sprinting = sprinting;
+	}
+
+	public void simulateSprint(boolean sprint)
+	{
+		PlayerToggleSprintEvent event = new PlayerToggleSprintEvent(this, sprint);
+		Bukkit.getPluginManager().callEvent(event);
```
TheBusyBiscuit, line 992: The method should respect the Event outcome and set the `sprinting` state accordingly.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -959,18 +964,28 @@ public void setSneaking(boolean sneaking)
 		this.sneaking = sneaking;
 	}
 
+	public void simulateSneak(boolean sneak)
+	{
+		PlayerToggleSneakEvent event = new PlayerToggleSneakEvent(this, sneak);
+		Bukkit.getPluginManager().callEvent(event);
```
TheBusyBiscuit, line 970: The method should respect the Event outcome and set the `sneaking` state accordingly.

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):
Some javadocs on these methods would be awesome but it looks good as is for now!

PR #198, opened 2021-04-03
Refactor LivingEntityMock

lluiscamino:
# Description
I saw https://github.com/seeseemelk/MockBukkit/issues/112 and thought that I could contribute by refactoring the `LivingEntityMock` class. This is the list of the changes I've done:

1. Overwritten the setHealth method in the `PlayerMock` class to avoid having to check if `(this instanceof Player)` inside `LivingEntityMock`.
2. Merged `damage(double amount)` and `damage(double amount, Entity source)` into one.
3. Simplied `hasPotionEffect(PotionEffectType type)` by making use `getPotionEffect(PotionEffectType type)`
3. Simplied `removePotionEffect(PotionEffectType type)`

# Checklist
- [ ] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -5,6 +5,7 @@ import static org.junit.Assert.fail;
 
 import java.net.InetSocketAddress;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1600,6 +1602,38 @@ public void setScoreboard(Scoreboard scoreboard)
 		throw new UnimplementedOperationException();
 	}
 
+	@Override
+	public void setHealth(double health) {
```
```suggestion
	public void setHealth(double health)
	{
```

seeseemelk (approved):

PR #201, opened 2021-04-19
Playermock can see implemented

wuza8:
# Description
Added implementation of functions PlayerMock:canSee/showPlayer/hidePlayer

# Checklist
- [x] Unit tests added.


src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1128,4 +1128,14 @@ public void testFlightEventFired()
 		assertTrue(player.isFlying());
 		server.getPluginManager().assertEventFired(PlayerToggleFlightEvent.class);
 	}
+
+	@Test
+	public void testPlayerHide(){
+		PlayerMock player2 = server.addPlayer();
+		player.hidePlayer(null, player2);
+		assertFalse(player.canSee(player2));
+		player.showPlayer(null, player2);
```
TheBusyBiscuit: Plugin is marked as `NotNull`, this right here is a direct violation of that constraint and should result in a test failure.

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1128,4 +1128,14 @@ public void testFlightEventFired()
 		assertTrue(player.isFlying());
 		server.getPluginManager().assertEventFired(PlayerToggleFlightEvent.class);
 	}
+
+	@Test
+	public void testPlayerHide(){
```
TheBusyBiscuit: formatting 
```suggestion
	public void testPlayerHide() {
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1479,41 +1480,37 @@ public void setAllowFlight(boolean flight)
...
 	@Deprecated
 	public void showPlayer(@NotNull Player player)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		hiddenPlayerNicks.remove(player.getName());
 	}
 
 	@Override
 	public void showPlayer(@NotNull Plugin plugin, @NotNull Player player)
```
TheBusyBiscuit, line 1511: `plugin` parameter is ignored.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1479,41 +1480,37 @@ public void setAllowFlight(boolean flight)
...
 	@Deprecated
 	public void hidePlayer(@NotNull Player player)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		hiddenPlayerNicks.add(player.getName());
 	}
 
 	@Override
 	public void hidePlayer(@NotNull Plugin plugin, @NotNull Player player)
```
TheBusyBiscuit, line 1496: `plugin` parameter is ignored.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -117,6 +117,7 @@
 
 	private final PlayerSpigotMock playerSpigotMock = new PlayerSpigotMock();
 	private final List<AudioExperience> heardSounds = new LinkedList<>();
+	private final List<String> hiddenPlayerNicks = new ArrayList<String>();
```
```suggestion
	private final List<String> hiddenPlayerNicks = new ArrayList<>();
```

TheBusyBiscuit: Also, UUIDs exist... player names are not guaranteed to be unique in Minecraft.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1479,41 +1480,37 @@ public void setAllowFlight(boolean flight)
...
-		throw new UnimplementedOperationException();
+		hiddenPlayerNicks.remove(player.getName());
 	}
 
 	@Override
 	public void showPlayer(@NotNull Plugin plugin, @NotNull Player player)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		showPlayer(player);
```
TheBusyBiscuit: This would yield an unwanted outcome.
To quote from the Spigot docs:

`If another another plugin had hidden the player too, then the player will remain hidden until the other plugin calls this method too.`

TheBusyBiscuit (changes requested):
This implementation is somewhat faulty, the plugin parameter should not be ignored.

wuza8:
Sorry for Pull Request duplication, i'm git noob :face_with_head_bandage:  I've changed implementation. I think the logic is here now, but I would need to make unit tests to prove it.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -117,6 +118,7 @@
 
 	private final PlayerSpigotMock playerSpigotMock = new PlayerSpigotMock();
 	private final List<AudioExperience> heardSounds = new LinkedList<>();
+	private final HashMap<UUID,List<String>> hiddenPlayerNicks = new HashMap<>();
```
TheBusyBiscuit: Collections, like Maps or Lists should always be declared with their base type to abstract away from the concrete implementation (makes it easier to change later to a different type of Map in case we need to.

Also the method name is a bit misleading, with `hiddenPlayerNicks` I would expect it to store nicknames of hidden players? But it just stores the UUIDs of hidden Players with a set of plugin names. So we should probably simplify that name.

```suggestion
	private final Map<UUID, List<String>> hiddenPlayers = new HashMap<>();
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1483,37 +1485,45 @@ public void setAllowFlight(boolean flight)
...
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return !hiddenPlayerNicks.containsKey(player.getUniqueId());
+	}
+
+	private void createRecordOfHiddenPlayerIfNotExists(Player playerToHide)
+	{
+		if(!hiddenPlayerNicks.containsKey(playerToHide.getUniqueId()))
+			hiddenPlayerNicks.put(playerToHide.getUniqueId(), new ArrayList<>());
```
```suggestion
		hiddenPlayerNicks.putIfAbsent(playerToHide.getUniqueId(), new ArrayList<>());
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1483,37 +1485,45 @@ public void setAllowFlight(boolean flight)
...
 
 	@Override
 	public boolean canSee(@NotNull Player player)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return !hiddenPlayerNicks.containsKey(player.getUniqueId());
+	}
+
+	private void createRecordOfHiddenPlayerIfNotExists(Player playerToHide)
```
```suggestion
	private void createRecordOfHiddenPlayerIfNotExists(@NotNull Player playerToHide)
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1483,37 +1485,45 @@ public void setAllowFlight(boolean flight)
...
 	@Override
 	public void showPlayer(@NotNull Plugin plugin, @NotNull Player player)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		if(hiddenPlayerNicks.containsKey(player.getUniqueId()))
+		{
+			List<String> blockingPlugins = hiddenPlayerNicks.get(player.getUniqueId());
+			blockingPlugins.remove(plugin.getName());
+			if(blockingPlugins.size() == 0)
```
TheBusyBiscuit: `.size() == 0` can be simplified.
```suggestion
			if (blockingPlugins.size().isEmpty())
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1483,37 +1485,45 @@ public void setAllowFlight(boolean flight)
...
 	}
 
 	@Override
 	public void showPlayer(@NotNull Plugin plugin, @NotNull Player player)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		if(hiddenPlayerNicks.containsKey(player.getUniqueId()))
+		{
+			List<String> blockingPlugins = hiddenPlayerNicks.get(player.getUniqueId());
```
TheBusyBiscuit: `.containsKey()` simply does a null check on `.get()`.
So you are essentially doing `.get()` twice which is wasteful from a performance point of view.
You can just do the null check yourself instead.

TheBusyBiscuit (changes requested):
Not a fan of this approach with the seperate `createRecordOfHiddenPlayerIfNotExists` method I think.
But regardless of that there are still some things that need to be fixed first.

TheBusyBiscuit:
> Sorry for Pull Request duplication, i'm git noob 🤕 I've changed implementation. I think the logic is here now, but I would need to make unit tests to prove it.

It's fine.
Thanks for making this pull request though! New features and contributions are always welcome!

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
seeseemelk: Surely not `.size().isEmpty()`?
```suggestion
			if(blockingPlugins.isEmpty())
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1486,37 +1488,44 @@ public void setAllowFlight(boolean flight)
...
-		throw new UnimplementedOperationException();
+		hiddenPlayerNicks.remove(player.getUniqueId());
 	}
 
 	@Override
 	public void showPlayer(@NotNull Plugin plugin, @NotNull Player player)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		if(hiddenPlayerNicks.containsKey(player.getUniqueId()))
```
```suggestion
		if (hiddenPlayerNicks.containsKey(player.getUniqueId()))
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1486,37 +1488,44 @@ public void setAllowFlight(boolean flight)
...
+		{
+			List<String> blockingPlugins = hiddenPlayerNicks.get(player.getUniqueId());
+			blockingPlugins.remove(plugin.getName());
+			if(blockingPlugins.isEmpty())
```
```suggestion
			if (blockingPlugins.isEmpty())
```

seeseemelk (approved):
Looks good to me, now you just have to convince TheBusyBiscuit.

TheBusyBiscuit:
> Looks good to me, now you just have to convince TheBusyBiscuit.

It doesn't compile though.

seeseemelk:
> > Looks good to me, now you just have to convince TheBusyBiscuit.
> 
> It doesn't compile though.

Wow, I didn't notice that. The `getPing` error is fixable by simply updating the branch though.

wuza8:
I've finally done something here, but there's a problem, I think bukkit changed API on PluginClassLoader, so I added null to compile project, but now every test spams 
"maj 15, 2021 1:08:46 PM org.bukkit.plugin.java.JavaPluginLoader <init>
WARNING: Could not initialize LibraryLoader (missing dependencies?)". I don't know what to do about it...

AyBiCi:
This new argument is Nullable so I was thinking null will fix everything...

AyBiCi:
I don't see this error on CI so maybe it's only my local copy and dependencies but I don't know.

wuza8:
Ok, I see now xd waiting for #213

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
```diff
@@ -327,11 +327,22 @@ public static void unmock()
 	 *
 	 * @return An instance of a mock plugin.
 	 */
-	public static MockPlugin createMockPlugin()
+	public static MockPlugin createMockPlugin(){
```
seeseemelk:
```suggestion
	public static MockPlugin createMockPlugin()
	{
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1486,37 +1488,49 @@ public void setAllowFlight(boolean flight)
...
-		throw new UnimplementedOperationException();
+		hidePlayer(plugin.getName(), player);
+	}
+
+	private void hidePlayer(@NotNull String pluginName, @NotNull Player player)
+	{
+		hiddenPlayers.putIfAbsent(player.getUniqueId(), new ArrayList<>());
+		List<String> blockingPlugins = hiddenPlayers.get(player.getUniqueId());
+		if(blockingPlugins.contains(pluginName)) return;
+		blockingPlugins.add(pluginName);
```
```suggestion
		if (!blockingPlugins.contains(pluginName))
			blockingPlugins.add(pluginName);
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1486,37 +1488,49 @@ public void setAllowFlight(boolean flight)
...
+		showPlayer(plugin.getName(), player);
+	}
+
+	private void showPlayer(@NotNull String pluginName, @NotNull Player player)
+	{
+		if(!hiddenPlayers.containsKey(player.getUniqueId())) return;
+		List<String> blockingPlugins = hiddenPlayers.get(player.getUniqueId());
+		blockingPlugins.remove(pluginName);
+		if(!blockingPlugins.isEmpty()) return;
+		hiddenPlayers.remove(player.getUniqueId());
```
seeseemelk: Is it a good idea to use `pluginName` here? Wouldn't it be better to simply use the `Plugin` itself as the key?

```suggestion
		if (hiddenPlayers.containsKey(player.getUniqueId()))
		{
			List<String> blockingPlugins = hiddenPlayers.get(player.getUniqueId());
			blockingPlugins.remove(pluginName);
			if (blockingPlugins.isEmpty())
				hiddenPlayers.remove(player.getUniqueId());
		}
```

seeseemelk (approved):

seeseemelk (changes requested):
Didn't mean to approve it

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
```diff
@@ -328,10 +328,22 @@ public static void unmock()
...
+	}
+
+	/**
+	 * Creates a mock instance of a {@link JavaPlugin} implementation and gives you a chance to name the plugin. This plugin offers no functionality, but it does
+	 * allow a plugin that might enable and disable other plugins to be tested.
+	 *
+	 * @param pluginName A name of a new plugin.
+	 * @return An instance of a mock plugin.
+	 */
+	public static MockPlugin createMockPlugin(String pluginName)
```
TheBusyBiscuit: This sounds like a sensible addition, though some annotations can be added here (nitpicky I know)

```suggestion
	public static @NotNull MockPlugin createMockPlugin(@NotNull String pluginName)
```

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
```diff
@@ -328,10 +328,22 @@ public static void unmock()
 	 * @return An instance of a mock plugin.
 	 */
 	public static MockPlugin createMockPlugin()
```
TheBusyBiscuit:
```suggestion
	public static @NotNull MockPlugin createMockPlugin()
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -118,6 +119,8 @@
 
 	private final PlayerSpigotMock playerSpigotMock = new PlayerSpigotMock();
 	private final List<AudioExperience> heardSounds = new LinkedList<>();
+	private final Map<UUID, List<Plugin>> hiddenPlayers = new HashMap<>();
+	private final List<UUID> hiddenPlayersDeprecated = new ArrayList<>();
```
TheBusyBiscuit: To be honest I am not the biggest fan of this being split into two here but I guess it is fine for now...

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1486,37 +1489,43 @@ public void setAllowFlight(boolean flight)
 	@Deprecated
 	public void hidePlayer(@NotNull Player player)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		if(!hiddenPlayersDeprecated.contains(player.getUniqueId()))
+			hiddenPlayersDeprecated.add(player.getUniqueId());
```
TheBusyBiscuit: If you change `hiddenPlayersDeprecated` to a `HashSet` here (since you are already preventing duplicates), you can simplify this.

```suggestion
		hiddenPlayersDeprecated.add(player.getUniqueId());
```

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1136,4 +1137,55 @@ void testFlightEventFired()
...
+		assertFalse(player.canSee(player2));
+		player.hidePlayer(player2);
+		assertFalse(player.canSee(player2));
+		player.showPlayer(player2);
+		assertFalse(player.canSee(player2));
+		player.showPlayer(plugin2, player2);
+		assertFalse(player.canSee(player2));
+		player.showPlayer(plugin1, player2);
+		assertTrue(player.canSee(player2));
+	}
```
TheBusyBiscuit: I feel like these should **definitely** be split into multiple Unit Tests O.o

TheBusyBiscuit (changes requested):

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
wuza8: Another magic class that pushes me to read the whole Java util reference

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
wuza8: It's now separated

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
wuza8: It can be done by making a fake plugin `private static final Plugin fakeDeprecatedPlugin = new ???()` and then just using this plugin on this deprecated functions like that `hidePlayer(fakeDeprecatedPlugin, player);` `showPlayer(fakeDeprecatedPlugin, player);` but I don't know how to get easy and safe this fake plugin object without just implementing like 20 functions from this interface

TheBusyBiscuit: We could also just have an "empty list" represent this but that would probably be a bit more work than necessary. It is fine for now

TheBusyBiscuit (approved):
lgtm

seeseemelk (approved):

PR #205, opened 2021-04-29
Implement WorldMock#getLoadedChunks and isChunkLoaded

seeseemelk:
# Description
This PR implements `WorldMock#getLoadedChunks()` and `WorldMock#isChunkLoaded()`.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
@@ -5,9 +5,9 @@ package be.seeseemelk.mockbukkit;
  */
 public class Coordinate
 {
-	public int x;
-	public int y;
-	public int z;
+	public final int x;
+	public final int y;
+	public final int z;
 
 	/**
 	 * Creates a new coordinate object with a specified (x, y, z).
--- a/src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
@@ -1,5 +1,7 @@
 package be.seeseemelk.mockbukkit;
 
+import org.jetbrains.annotations.NotNull;
+
 /**
  * A simple class that contains an x, y and z coordinate as integers.
  */
@@ -36,6 +38,7 @@ public class Coordinate
 	 *
 	 * @return The chunk coordinate of this coordinate.
 	 */
+	@NotNull
 	public ChunkCoordinate toChunkCoordinate()
 	{
 		return new ChunkCoordinate(x >> 4, z >> 4);
--- a/src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
@@ -38,7 +38,7 @@ public class Coordinate
 	 */
 	public ChunkCoordinate toChunkCoordinate()
 	{
-		return new ChunkCoordinate(x / 16, z / 16);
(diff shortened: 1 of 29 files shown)


src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
```diff
@@ -31,6 +31,16 @@ public Coordinate()
...
 	}
 
+	/**
+	 * Gets the chunk coordinate this coordinate is in.
+	 *
+	 * @return The chunk coordinate of this coordinate.
+	 */
+	public ChunkCoordinate toChunkCoordinate()
+	{
+		return new ChunkCoordinate(x / 16, z / 16);
```
TheBusyBiscuit: This should rather use bit-shifting.
```suggestion
		return new ChunkCoordinate(x >> 4, z >> 4);
```

src/main/java/be/seeseemelk/mockbukkit/Coordinate.java
```diff
@@ -31,6 +31,16 @@ public Coordinate()
 		this(0, 0, 0);
 	}
 
+	/**
+	 * Gets the chunk coordinate this coordinate is in.
+	 *
+	 * @return The chunk coordinate of this coordinate.
+	 */
+	public ChunkCoordinate toChunkCoordinate()
```
TheBusyBiscuit, line 42: We may also want to annotate this with `@NotNull`

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -268,8 +269,17 @@ public boolean setSpawnLocation(int x, int y, int z)
 	@Override
 	public ChunkMock getChunkAt(int x, int z)
 	{
-		ChunkMock chunk = new ChunkMock(this, x, z);
-		return chunk;
+		return getChunkAt(new ChunkCoordinate(x, z));
+	}
+
+	public ChunkMock getChunkAt(ChunkCoordinate coordinate)
```
TheBusyBiscuit: docs?
And `@NotNull` for the parameter and return type?

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -268,8 +269,17 @@ public boolean setSpawnLocation(int x, int y, int z)
...
+	{
+		if (!loadedChunks.containsKey(coordinate))
+		{
+			ChunkMock chunk = new ChunkMock(this, coordinate.getX(), coordinate.getZ());
+			loadedChunks.put(coordinate, chunk);
+		}
+		return loadedChunks.get(coordinate);
```
TheBusyBiscuit: Using `HashMap#containsKey()` like that is a bad idea.
Internally `containsKey()` does a null check on `.get()`. Since you are getting the value here anyway, this results in the same operation being run twice, causing redundancy.
Moreso, you are running `.get()` on a value you just `.put()` into the Map yourself.

This can all be heavily improved by storing it as a value and doing a manual null check:

```suggestion
		ChunkMock chunk = loadedChunks.get(coordinate);
		if (chunk == null)
		{
			chunk = new ChunkMock(this, coordinate.getX(), coordinate.getZ());
			loadedChunks.put(coordinate, chunk);
		}
		return chunk;
```

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):
LGTM

PR #208, opened 2021-05-04
MockBukkit 1.0.0 🚀 - Migrating to JUnit 5

TheBusyBiscuit:
# Description
As discussed with @seeseemelk the time for the first MockBukkit 1.x release has come!
This update will bring a couple of drastic changes.

1. Everything has been migrated from JUnit4 to JUnit5, including unit tests and our own assertion methods
2. The deprecated method `MockBukkit.unload()` has finally been removed

When you update your dependencies, things *should* probably continue to work just fine.
But I recommend you to double-check that just to be sure!

# Checklist
- [ ] Unit tests added.


TheBusyBiscuit:
Nice, this has merge conflicts now ._.

timkalkus:
I think the [README.md](https://github.com/MockBukkit/MockBukkit#readme) should also be updated (haven't found it under the changed files).

If I understand the changes correctly the dependency path for gradle/maven from `com.github.seeseemelk` has to be changed to `com.github.mockbukkit`. Also simultaneously updating the version in the gradle/maven examples to v1.0.0 may be nice.

TheBusyBiscuit:
> If I understand the changes correctly the dependency path for gradle/maven from `com.github.seeseemelk` has to be changed to `com.github.mockbukkit`. Also simultaneously updating the version in the gradle/maven examples to v1.0.0 may be nice.

No, the `groupId` hasn't been changed, it still remains `com.github.seeseemelk`.
The version could be updated though I guess, will do so.

TheBusyBiscuit:
Updated example in README to 1.0.0, thanks

seeseemelk:
> > If I understand the changes correctly the dependency path for gradle/maven from `com.github.seeseemelk` has to be changed to `com.github.mockbukkit`. Also simultaneously updating the version in the gradle/maven examples to v1.0.0 may be nice.
> 
> No, the `groupId` hasn't been changed, it still remains `com.github.seeseemelk`.
> The version could be updated though I guess, will do so.

This indeed hasn't changed yet. I still have to open up a support ticket in order to get a second groupId approved so that I can release artifacts under that groupId.

src/test/java/be/seeseemelk/mockbukkit/metadata/MetadataTest.java
```diff
@@ -1,100 +0,0 @@
...
-	@Test
-	public void testRemoveMetadataNotSet()
-	{
-		MockPlugin plugin = MockBukkit.createMockPlugin();
-		assertFalse(testSubject.hasMetadata("MyMetadata3"));
-		testSubject.removeMetadata("MyMetadata3", plugin);
-		assertFalse(testSubject.hasMetadata("MyMetadata3"));
-	}
-
-}
```
seeseemelk, line 100: Why was this file removed?

src/test/java/be/seeseemelk/mockbukkit/scoreboard/TeamMockTest.java
```diff
@@ -40,58 +42,60 @@ public void setUp() throws Exception
...
 		assertFalse(team.canSeeFriendlyInvisibles());
 	}
 
+	@SuppressWarnings("deprecation")
 	@Test
-	public void getNameTagVisibility()
+	void getNameTagVisibility()
 	{
-		assertEquals(NameTagVisibility.ALWAYS, team.getNameTagVisibility());
+		assertEquals(org.bukkit.scoreboard.NameTagVisibility.ALWAYS, team.getNameTagVisibility());
```
seeseemelk, line 93: Why is the FQN used here instead of an import?

seeseemelk (changes requested):

src/test/java/be/seeseemelk/mockbukkit/metadata/MetadataTest.java
TheBusyBiscuit: This test is not possible that way in JUnit5 and it is completely redundant anyway,
we already have a MetadataTable test, testing the exact same methods for each class that has a MetadataTable is just plain redundant.

src/test/java/be/seeseemelk/mockbukkit/scoreboard/TeamMockTest.java
TheBusyBiscuit: Because `NameTagVisibility` is a deprecated enum.
I rather suppress deprecation warnings for a test instead of for a whole class because of an import.
Can change that if you want of course.

seeseemelk (approved):

PR #212, opened 2021-05-13
Bump to gradle 7

md5sha256:
# Description
- Bump the gradle wrapper to 7.0.2. 
- Removed the `jcenter()` repository declaration, replacing it with `mavenCentral()`
- Added [toolchains](https://docs.gradle.org/current/userguide/toolchains.html) for the `JavaCompile` task
- Added a `gradle.properties` file; explicitly disable auto toolchains provisioning
- Converted plugin declarations to use the newer `plugins` block
- Converted deprecated method assignments to use the setters instead

# Checklist
- [ ] Unit tests added.


TheBusyBiscuit (approved):
LGTM but I have too few experience with gradle to make a proper judgement here.
@seeseemelk may do a better job here

md5sha256:
I've just read that `options.release` is broken on Java 9 (this is a javac bug apparently) so I've fixed that by bumping the version check in the if-block to `> 9`. 
I've also realized that `options.release` superceed `sourceCompatibility` and `targetCompatibility` as compiler flags to specify the bytecode output. That has now also been fixed.

TheBusyBiscuit (approved):

seeseemelk (approved):

PR #217, opened 2021-05-31
Improved generation of random playernames

Mmaarten23:
[+] Allowed for more than 100 random names to be requested
[*] All generated random names now follow the minecraft.net specifications for usernames

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
@@ -30,7 +30,7 @@ public final class PlayerMockFactory
 	{
 		String firstName = FIRST_NAMES[random.nextInt(FIRST_NAMES.length)];
 		String lastName = LAST_NAMES[random.nextInt(LAST_NAMES.length)];
-		return firstName + " " + lastName;
+		return firstName + "_" + lastName;
 	}
 
 	/**
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
@@ -46,6 +46,7 @@ public final class PlayerMockFactory
 			name = getRandomName();
 		}
 		while (usedNames.contains(name));
+		usedNames.add(name);
 		return name;
 	}
 
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
@@ -38,7 +38,8 @@ public final class PlayerMockFactory {
 	 *
 	 * @return A newly created player mock object.
 	 */
-	public PlayerMock createRandomPlayer() {
+	public @NotNull PlayerMock createRandomPlayer()
+	{
 		String name = getUniqueRandomName();
 		UUID uuid = new UUID(random.nextLong(), random.nextLong());
 		return new PlayerMock(server, name, uuid);
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
@@ -30,7 +30,7 @@ public final class PlayerMockFactory
 	{
 		String firstName = FIRST_NAMES[random.nextInt(FIRST_NAMES.length)];
 		String lastName = LAST_NAMES[random.nextInt(LAST_NAMES.length)];
-		return firstName + "_" + lastName;
+		return firstName + "_" + lastName + random.nextInt(9999);
 	}
 
 	/**
@@ -40,12 +40,7 @@ public final class PlayerMockFactory
 	 */
 	private String getUniqueRandomName()
 	{
-		if (usedNames.size() >= 100)
-		{
-			throw new RuntimeException("Out of names");
-		}
-
-		while (true)
+		for (int i = 0; i < 10; i++)
 		{
 			String name = getRandomName();
 			if (!usedNames.contains(name))
@@ -54,6 +49,7 @@ public final class PlayerMockFactory
 				return name;
 			}
 		}
+		throw new RuntimeException("Failed to find a unique name");
 	}
 
 	/**
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
@@ -37,6 +37,7 @@ public final class PlayerMockFactory
 	 * Gets a unique random name. Any name that is generated will never be generated again.
 	 *
 	 * @return A unique random name.
+	 * @throws RuntimeException if no unique name is found after 10 tries
 	 */
 	private String getUniqueRandomName()
 	{
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
@@ -20,9 +20,15 @@ public final class PlayerMockFactory {
 	 *
 	 * @return A unique random name.
 	 */
-	private String getUniqueRandomName() {
-		String name = "name_" + currentNameIndex++;
-		if (name.length() > 16) throw new RuntimeException("Maximum number of player names reached");
+	private @NotNull String getUniqueRandomName()
+	{
+		String name = "Player" + currentNameIndex++;
+
+		if (name.length() > 16)
+		{
+			throw new IllegalStateException("Maximum number of player names reached");
+		}
+
 		return name;
 	}
 
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
@@ -9,8 +9,8 @@ import be.seeseemelk.mockbukkit.ServerMock;
 
 public final class PlayerMockFactory
 {
-	private static final String[] FIRST_NAMES = { "James", "Mary", "John", "Particia", "Robert", "Jennifer", "Michael", "Elizabeth", "William", "Linda" };
-	private static final String[] LAST_NAMES = { "Smith", "Johnson", "Jones", "Brown", "Davis", "Miller", "Wilson", "Moore", "Taylor", "Anderson" };
+	private static final String[] FIRST_NAMES = { "James", "Mary", "John", "Amy", "Cora", "Edie", "Abe", "Alex", "Steve", "Linda" };
+	private static final String[] LAST_NAMES = { "Smith", "Abby", "Jones", "Brown", "Davis", "Miller", "Wilson", "Moore", "Taylor", "Bean" };
 
 	private final ServerMock server;
 	private Random random = new Random();
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
@@ -1,6 +1,7 @@
 package be.seeseemelk.mockbukkit.entity;
 
 import be.seeseemelk.mockbukkit.ServerMock;
+import org.jetbrains.annotations.NotNull;
 
 import java.util.Random;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -30,22 +30,18 @@ private String getRandomName()
...
 	 */
 	private String getUniqueRandomName()
 	{
-		if (usedNames.size() >= 100)
-		{
-			throw new RuntimeException("Out of names");
-		}
-
-		while (true)
+		for (int i = 0; i < 10; i++)
```
TheBusyBiscuit: Imposing a hard limit of 10 tries here is a terrible idea.

While the current limitation of 100 names is not the best solution either, it ensured a reliable and consistent name generation. By imposing this limit of 10 tries, you run into the chance of generating 10 names that already existed.
This will now throw a `RuntimeException`, making the unit test fail.

In other words: There is now a probality for randomly failing a unit test, which defeats the point behind a unit test. Never make your unit tests dependent on randomness being in your favour.

TheBusyBiscuit (changes requested):

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
timkalkus: You could change the name generation to simply use an counter which is added to each name, making any doubling impossible: `<RandomPlayerName>0, <OtherRandomPlayerName>1, ...`

This would also remove the need to check, whether a name was already generated.

But if you don't want consecutive numbers, you could also generate a (short) hash value based on the counter and add this to the player string (then the check for doubling names may be needed in rare circumstances).

Mmaarten23:
Updated the code to always provide a random username (unless more than 100.000 players are added to the server)
I just want to ask something though, is it really needed that they have "real" names? 
Because I have never seen someone creating random names like this.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -40,20 +40,13 @@ private String getRandomName()
...
-		{
-			String name = getRandomName();
-			if (!usedNames.contains(name))
-			{
-				usedNames.add(name);
-				return name;
-			}
+			name = getRandomName();
 		}
+		while (usedNames.contains(name));
```
wuza8: It should add this new name to already used names here
```suggestion
		while (usedNames.contains(name));
		usedNames.add(name);
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -1,68 +1,37 @@
...
 	private final ServerMock server;
-	private Random random = new Random();
-	private Set<String> usedNames = new HashSet<>();
+	private int currentNameIndex;
+	private final Random random = new Random();
 
-	public PlayerMockFactory(ServerMock server)
-	{
+	public PlayerMockFactory(ServerMock server) {
+		this.currentNameIndex = 0;
```
TheBusyBiscuit:
```suggestion
	public PlayerMockFactory(@NotNull ServerMock server)
	{
		this.currentNameIndex = 0;
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -1,68 +1,37 @@
...
-			{
-				usedNames.add(name);
-				return name;
-			}
-		}
+	private String getUniqueRandomName() {
+		String name = "name_" + currentNameIndex++;
+		if (name.length() > 16) throw new RuntimeException("Maximum number of player names reached");
+		return name;
 	}
```
TheBusyBiscuit: We should also annotate this, also formatting.

I also feel like `name_` is not descriptive enough. I think "Player" + number is a bit more reasonable.
Then the use of `RuntimeException` here doesn't really seem verbose enough to me. Perhaps an `IllegalStateException`?
The docs also need to be updated too.

```suggestion
	private @NotNull String getUniqueRandomName()
	{
		String name = "Player" + currentNameIndex++;

		if (name.length() > 16)
		{
			throw new IllegalStateException("Maximum number of player names reached");
		}

		return name;
	}
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -1,68 +1,37 @@
...
 
 	/**
 	 * Create a random player mock object with a unique name.
 	 *
 	 * @return A newly created player mock object.
 	 */
-	public PlayerMock createRandomPlayer()
-	{
+	public PlayerMock createRandomPlayer() {
```
TheBusyBiscuit: Formatting.
```suggestion
	public @NotNull PlayerMock createRandomPlayer()
	{
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -73,8 +42,7 @@ public PlayerMock createRandomPlayer()
 	 *
 	 * @return A newly created player mock object.
 	 */
-	public OfflinePlayerMock createRandomOfflinePlayer()
-	{
+	public OfflinePlayerMock createRandomOfflinePlayer() {
```
TheBusyBiscuit: Formatting
```suggestion
	public @NotNull OfflinePlayerMock createRandomOfflinePlayer()
	{
```

TheBusyBiscuit (changes requested):
I am fine with a name scheme like this, the names themselves do not really matter that much anyway.
But there are some formatting issues that need to be fixed.

TheBusyBiscuit (changes requested):
Missing import for the annotations

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -1,67 +1,45 @@
...
-	private static final String[] LAST_NAMES = { "Smith", "Johnson", "Jones", "Brown", "Davis", "Miller", "Wilson", "Moore", "Taylor", "Anderson" };
+import java.util.Random;
+import java.util.UUID;
 
+public final class PlayerMockFactory {
 	private final ServerMock server;
-	private Random random = new Random();
-	private Set<String> usedNames = new HashSet<>();
+	private int currentNameIndex;
+	private final Random random = new Random();
```
TheBusyBiscuit: personal preference
```suggestion
	private final Random random = new Random();
	private int currentNameIndex;
```

TheBusyBiscuit (approved):
Looks good from my point of view. Leaving this open though in case Seesee has any sufficient reasons as to why the old name-based generation would be better.

Mmaarten23:
image
pretty sure he approves

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -1,67 +1,45 @@
...
-	private static final String[] FIRST_NAMES = { "James", "Mary", "John", "Particia", "Robert", "Jennifer", "Michael", "Elizabeth", "William", "Linda" };
-	private static final String[] LAST_NAMES = { "Smith", "Johnson", "Jones", "Brown", "Davis", "Miller", "Wilson", "Moore", "Taylor", "Anderson" };
+import java.util.Random;
+import java.util.UUID;
 
+public final class PlayerMockFactory {
 	private final ServerMock server;
-	private Random random = new Random();
-	private Set<String> usedNames = new HashSet<>();
+	private final Random random = new Random();
```
wuza8, line 12:
```suggestion
```
This isn't used anymore

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -1,67 +1,45 @@
...
+	private int currentNameIndex;
```
wuza8, line 13:
```suggestion
	private long currentNameIndex;
```
int type would limit us to use only 2147483647 players, because after this number currentNameIndex would overflow

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -1,67 +1,45 @@
...
-		return firstName + " " + lastName;
-	}
-
 	/**
 	 * Gets a unique random name. Any name that is generated will never be generated again.
 	 *
 	 * @return A unique random name.
 	 */
-	private String getUniqueRandomName()
+	private @NotNull String getUniqueRandomName()
```
wuza8, line 26:
```suggestion
	private @NotNull String getUniqueName()
```
It's not random anymore

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -1,67 +1,45 @@
...
-import java.util.UUID;
-
 import be.seeseemelk.mockbukkit.ServerMock;
+import org.jetbrains.annotations.NotNull;
 
-public final class PlayerMockFactory
-{
-	private static final String[] FIRST_NAMES = { "James", "Mary", "John", "Particia", "Robert", "Jennifer", "Michael", "Elizabeth", "William", "Linda" };
-	private static final String[] LAST_NAMES = { "Smith", "Johnson", "Jones", "Brown", "Davis", "Miller", "Wilson", "Moore", "Taylor", "Anderson" };
+import java.util.Random;
```
wuza8, line 6:
```suggestion
```
Import not used

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -1,67 +1,45 @@
...
 
 	/**
 	 * Create a random player mock object with a unique name.
 	 *
 	 * @return A newly created player mock object.
 	 */
-	public PlayerMock createRandomPlayer()
+	public @NotNull PlayerMock createRandomPlayer()
 	{
 		String name = getUniqueRandomName();
```
wuza8, line 45:
```suggestion
		String name = getUniqueName();
```

wuza8 (changes requested):

okx-code:
+1 to this. Spaces in names really mess with tests that rely on commands.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
seeseemelk: But this import is used at line 11

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -1,67 +1,45 @@
...
+import org.jetbrains.annotations.NotNull;
 
-public final class PlayerMockFactory
-{
-	private static final String[] FIRST_NAMES = { "James", "Mary", "John", "Particia", "Robert", "Jennifer", "Michael", "Elizabeth", "William", "Linda" };
-	private static final String[] LAST_NAMES = { "Smith", "Johnson", "Jones", "Brown", "Davis", "Miller", "Wilson", "Moore", "Taylor", "Anderson" };
+import java.util.Random;
+import java.util.UUID;
 
+public final class PlayerMockFactory {
```
seeseemelk:
```suggestion
public final class PlayerMockFactory
{
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
seeseemelk: Indeed, this would make it so that the exception at line 31 could never be thrown, because Player2147483647 is exactly 16 characters long. That said, I don't think it is realistic anyone would ever create that many players.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
seeseemelk: It is still used at line 45, so it can't be deleted.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
seeseemelk: It is still used by the Random property at line 11, which is itself still used.

seeseemelk (approved):

seeseemelk (approved):

PR #219, opened 2021-06-01
Change simulate methods in PlayerMock to return the event they fired. Change build.gradle to run tests.

ohowe1:
# Description
This pull request changed the simulate methods in `PlayerMock` to return the event that they fired to allow for better testing. This pull request also added the lines:
```
test {
    useJUnitPlatform()
}
```
to the build.gradle file so that the junit tests run when the test task is run. If this was intentional, let me know so that I can roll back the file.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -133,6 +133,10 @@ publishing {
     }
 }
 
+test {
+    useJUnitPlatform()
+}
+
 eclipse.classpath.downloadJavadoc = true
 eclipse.classpath.downloadSources = true
 
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -186,7 +186,7 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 	 * @return {@code true} if the block was damaged, {@code false} if the event was cancelled or the player was not in
 	 *         survival gamemode.
 	 */
-	public boolean simulateBlockDamage(Block block)
+	public @Nullable BlockDamageEvent simulateBlockDamage(Block block)
 	{
 		if (gamemode == GameMode.SURVIVAL)
 		{
@@ -199,11 +199,11 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 					block.setType(Material.AIR);
 			}
 
-			return !event.isCancelled();
+			return event;
 		}
 		else
 		{
-			return false;
+			return null;
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -283,12 +283,13 @@ public void respawn()
 	 *
 	 * @param moveLocation Location to move player to
 	 */
-	public void simulatePlayerMove(@NotNull Location moveLocation)
+	public PlayerMoveEvent simulatePlayerMove(@NotNull Location moveLocation)
```
```suggestion
	public @NotNull PlayerMoveEvent simulatePlayerMove(@NotNull Location moveLocation)
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -972,14 +973,15 @@ public void setSneaking(boolean sneaking)
 		this.sneaking = sneaking;
 	}
 
-	public void simulateSneak(boolean sneak)
+	public PlayerToggleSneakEvent simulateSneak(boolean sneak)
```
```suggestion
	public @NotNull PlayerToggleSneakEvent simulateSneak(boolean sneak)
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -994,14 +996,15 @@ public void setSprinting(boolean sprinting)
 		this.sprinting = sprinting;
 	}
 
-	public void simulateSprint(boolean sprint)
+	public PlayerToggleSprintEvent simulateSprint(boolean sprint)
```
```suggestion
	public @NotNull PlayerToggleSprintEvent simulateSprint(boolean sprint)
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1538,14 +1541,15 @@ public void setFlying(boolean value)
 		this.flying = value;
 	}
 
-	public void simulateToggleFlight(boolean fly)
+	public PlayerToggleFlightEvent simulateToggleFlight(boolean fly)
```
```suggestion
	public @NotNull PlayerToggleFlightEvent simulateToggleFlight(boolean fly)
```

build.gradle
```diff
@@ -133,6 +133,10 @@ publishing {
     }
 }
 
+test {
+    useJUnitPlatform()
```
TheBusyBiscuit, line 137: Afaik the `Junit` *should* be the default, so this shouldn't be necessary.
But it cannot hurt to declare this explicitly. 👍🏻

TheBusyBiscuit (changes requested):
Looks good to me, some missing annotations can be added though

ohowe1:
Those requests should be implemented now. As for Junit, it must have been a local preference that made it so tests didn't run for some reason.

TheBusyBiscuit (approved):
LGTM.

Thanks for changing that one unit test to a ParameterizedTest and fixing that typo!

seeseemelk:
I'm a little bit considered that the return values of the simulate methods were changed. This introduces a breaking change which could break projects when they attempt to upgrade from `v1.1.0` to `v1.2.0`. I would have preferred to see these being introduced under a slightly different name, and have the old methods be annotated with `@deprecated`. But it has already been merged

TheBusyBiscuit:
@seeseemelk Oh, sorry about that.
We can always change that in retrospect, afaik you haven't pushed it to maven-central yet right?

PR #222, opened 2021-06-02
Implement WorldBorderMock and WorldMock#getWorldBorder

ohowe1:
# Description
Relatively straight forward pull request. Adds `WorldBorderMock` and implements it into `WorldMock`

# Checklist
- [x] Unit tests added.


src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -1158,8 +1161,7 @@ public boolean isGameRule(String rule)
 	@Override
 	public WorldBorder getWorldBorder()
```
TheBusyBiscuit: Wrong return type and missing annotation

```suggestion
	public @NotNull WorldBorderMock getWorldBorder()
```

src/test/java/be/seeseemelk/mockbukkit/WorldBorderMockTest.java
```diff
@@ -0,0 +1,133 @@
+package be.seeseemelk.mockbukkit;
+
+import org.bukkit.Location;
+import org.bukkit.World;
+import org.bukkit.WorldBorder;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
```
TheBusyBiscuit: wildcard imports *cries*

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,136 @@
...
+{
+
+	World world;
+	double size;
+	double damageAmount = 0.2;
+	double damageBuffer = 5.0;
+	int warningDistance = 5;
+	int warningTime = 15;
+	double centerX = 0;
+	double centerZ = 0;
```
TheBusyBiscuit: Missing access modifiers

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,136 @@
+package be.seeseemelk.mockbukkit;
+
+import org.apache.commons.lang.Validate;
+import org.bukkit.Location;
+import org.bukkit.World;
+import org.bukkit.WorldBorder;
+import org.bukkit.util.BoundingBox;
+import org.jetbrains.annotations.NotNull;
+
+public class WorldBorderMock implements WorldBorder
```
TheBusyBiscuit, line 18: Some docs would be nice, no necessity though

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
TheBusyBiscuit: Also are these the correct default values here?

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,136 @@
...
+{
+
+	World world;
```
TheBusyBiscuit: World should be final.

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,136 @@
...
+	@Override
+	public void setSize(double newSize)
+	{
+		setSize(newSize, 0);
+	}
+
+	@Override
+	public void setSize(double newSize, long seconds)
+	{
+		// I dont think that implementing seconds here is very important
```
TheBusyBiscuit: why.

TheBusyBiscuit (changes requested):
Not a full review but I spotted some things that should be addressed before continuing.

Thanks for the pull request though!

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
ohowe1: > Also are these the correct default values here?

Yep

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,166 @@
...
+	public WorldBorderMock(World world, Server server)
+	{
+		this.world = world;
+		this.server = server;
+	}
+
+	@Override
+	public void reset()
+	{
+		setSize(6.0E7);
```
TheBusyBiscuit: `6.0E7` is used twice here, I think there should be a constant for this at the top of the class.
`private static final double DEFAULT_BORDER_SIZE = 6.0E7`

Avoiding magic numbers like these makes it easier to tweak them if this default size changes in future versions of Minecraft.

*Same theoretically goes for the other values.*

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,166 @@
...
+	public void setCenter(@NotNull Location location)
+	{
+		setCenter(location.getX(), location.getZ());
+	}
+
+	@Override
+	public void setCenter(double x, double z)
+	{
+		this.centerX = x;
+		this.centerZ = z;
```
TheBusyBiscuit: CraftBukkit does a minmax check here before setting these values:
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/src/main/java/org/bukkit/craftbukkit/CraftWorldBorder.java#63-64

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
```diff
@@ -23,7 +10,10 @@
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
-import be.seeseemelk.mockbukkit.block.BlockMock;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
```
TheBusyBiscuit: Wildcard imports D:

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
```diff
@@ -1,20 +1,7 @@
...
-
-import java.util.List;
-
-import org.bukkit.Chunk;
-import org.bukkit.GameRule;
-import org.bukkit.Location;
-import org.bukkit.Material;
-import org.bukkit.World;
+import be.seeseemelk.mockbukkit.block.BlockMock;
+import org.bukkit.*;
```
TheBusyBiscuit: Do not use a wildcard import here :monkaS:

TheBusyBiscuit (changes requested):

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,188 @@
...
+	private double centerX;
+	private double centerZ;
+
+	/**
+	 * Creates a new world border mock
+	 *
+	 * @param world The world it is the border of
+	 * @param server The server it is in
+	 */
+	public WorldBorderMock(World world, Server server)
```
TheBusyBiscuit:
```suggestion
	public WorldBorderMock(@NotNull World world,@NotNull Server server)
```

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,188 @@
...
+{
+
+	private static final double DEFAULT_BORDER_SIZE = 6.0E7;
+	private static final double DEFAULT_DAMAGE_AMOUNT = 0.2;
+	private static final double DEFAULT_DAMAGE_BUFFER = 5.0;
+	private static final int DEFAULT_WARNING_DISTANCE = 5;
+	private static final int DEFAULT_WARNING_TIME = 15;
+	private static final double DEFAULT_CENTER_X = 0;
+	private static final double DEFAULT_CENTER_Z = 0;
+	private static final double MAX_CENTER_VALUE = 3.0E7D;
```
TheBusyBiscuit: Why is `MAX_CENTER_VALUE` the only one with a type literal suffix?

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,188 @@
...
+					{
+						size += distancePerTick;
+					}
+					else
+					{
+						size = newSize;
+						this.cancel();
+					}
+				}
+			}, 1, 1);
```
TheBusyBiscuit: We can use a lambda-expression here.
Also is `null` valid to use here?

```suggestion
			server.getScheduler().runTaskTimer(null, () ->
			{
				if ((size < newSize && distance > 0) || (size > newSize && distance < 0))
				{
					size += distancePerTick;
				}
				else
				{
					size = newSize;
					this.cancel();
				}
			}, 1, 1);
```

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,188 @@
...
+		}
+		else
+		{
+			setSize(newSize);
+		}
+	}
+
+	@NotNull
+	@Override
+	public Location getCenter()
```
TheBusyBiscuit:
```suggestion
	@Override
	public @NotNull Location getCenter()
```

TheBusyBiscuit (dismissed):

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
ohowe1: > Also is `null` valid to use here?

Unless there is an system for internally scheduling repeating tasks inside mockbukkit, then I think that this is the only way to do it. The null should not be a problem as all plugin is used for is canceling tasks that were created by the plugin in `BukkitSchedulerMock#cancelTasks`.

ohowe1: > We can use a lambda-expression here.

It appears that a lambada cannot be used as the cancel method is called.

seeseemelk:
@TheBusyBiscuit Are you still requesting changes here?

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
```diff
@@ -0,0 +1,187 @@
...
+			double ticksToTake = seconds * 20;
+
+			double distancePerTick = distance / ticksToTake;
+
+			server.getScheduler().runTaskTimer(null, new BukkitRunnable()
+			{
+				@Override
+				public void run()
+				{
+					if ((size < newSize && distance > 0) || (size > newSize && distance < 0))
```
TheBusyBiscuit: Floating point values may not always converge to exactly 0.0

TheBusyBiscuit (commented):
Sorry for the late response!

Most of my previous requests have been addressed.
Only things I am pondering about are param validation and floating point issues.
Other than that, it looks mergeable @seeseemelk

src/main/java/be/seeseemelk/mockbukkit/WorldBorderMock.java
ohowe1: I think you are referring to the size comparison to newSize, correct? So if I'm not completely misunderstanding then something like `newSize - size < delta` would work? As in if size is bigger than newSize within this margin.

Insprill (approved):

PR #223, opened 2021-06-03
Implements function that removes player from server and fixes player duplication on server

### LINKED ISSUE #220: Removing/Disconnecting player from ServerMock
Is there a way now to disconnect player from ServerMock? I wasn't able to found it so if there's none, what about something like `server.removePlayer(player)`? Name with respect to `server.addPlayer(player)`.

wuza8:
# Description
This PR:
- adds `server.removePlayer(PlayerMock)` function
- doesn't allow for duplicated players on server (the same name or uuid)
- fixes `server.setOfflinePlayers(int)` that was duplicating players on server
- fixes one random test, that also was duplicating player on server

There's also a problem with naming functions `server.setOfflinePlayers()` - it actually adds new offline players, doesn't set a number of them.

# Checklist
- [x] Unit tests added.


wuza8:
I didn't realized that disconnecting player from a server causes some data loss inside Player class because of converting Player to PlayerData and then saving that PlayerData to file, so when player connects back again, it's loaded from this file and because of that Player class should have some data loss like losing data of players that this certain player was hiding etc. I will implement it asap.

wuza8:
Problem mentioned before fixed.

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -111,6 +112,7 @@
 {
 	private static final String BUKKIT_VERSION = "1.16.5";
 	private static final String JOIN_MESSAGE = "%s has joined the server.";
+	private static final String QUIT_MESSAGE = "%s has exit the server.";
```
ohowe1:
```suggestion
	private static final String QUIT_MESSAGE = "%s has exited the server.";
```
Small typo I noticed while looking through this.

src/main/java/be/seeseemelk/mockbukkit/MockPlayerList.java
```diff
@@ -62,10 +62,21 @@ public BanList getProfileBans()
...
 		onlinePlayers.add(player);
 		offlinePlayers.add(player);
 	}
 
+	public void removePlayer(@NotNull PlayerMock player) throws IllegalStateException
+	{
+		if(onlinePlayers.contains(player))
+			onlinePlayers.remove(player);
+		else
+			throw new IllegalStateException("Player is not connected!");
```
TheBusyBiscuit: This can be simplified, no need to perform two operations here, we get the lookup for free.

```suggestion
		if (!onlinePlayers.remove(player)) {
			throw new IllegalStateException("Player is not connected!");
		}
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1017,6 +1027,18 @@ public void saveData()
...
+	 * This function simulates data loss from consequences of
+	 * qutting the server by player. It should be used only in ServerMock.
+	 */
+	public void simulateDataLossAfterSave()
+	{
+		hiddenPlayers.clear();
+		hiddenPlayersDeprecated.clear();
+		heardSounds.clear();
+		sneaking = sprinting = false;
+	}
```
TheBusyBiscuit, line 1407: hmmmmmmmm, not too sure what to think of this method here

src/main/java/be/seeseemelk/mockbukkit/MockPlayerList.java
```diff
@@ -95,6 +106,13 @@ public boolean isSomeoneOnline()
 		return !onlinePlayers.isEmpty();
 	}
 
+	public boolean isPlayerAlreadyOnServer(@NotNull PlayerMock player)
```
TheBusyBiscuit: I think this is a more reasonable name.
```suggestion
	public boolean isOnline(@NotNull PlayerMock player)
```

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -236,14 +246,36 @@ public PlayerMock addPlayer()
...
 	}
 
+	/**
+	 * Removes previously added player, calls PlayerQuitEvent on server
+	 *
+	 * @param player player to remove.
+	 * @return PlayerQuitEvent that was called on a server.
+	 */
+	@NotNull
+	public PlayerQuitEvent removePlayer(@NotNull PlayerMock player)
```
TheBusyBiscuit: Not sure if a `remove` operation should return an event here...

TheBusyBiscuit (changes requested):

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
wuza8: If someone would want to getQuitMessage he would need to get this event I think

src/main/java/be/seeseemelk/mockbukkit/MockPlayerList.java
wuza8: xd

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
wuza8: Because this removePlayer function doesn't emulate saving player data in any state (all the data stays in PlayerMock object), instead of saving and reading again needed variables, it just deletes the ones that aren't saved by a server after player quits, so it's faster and it takes much less code

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
wuza8: So I'm not sure what ServerMock.removePlayer should return here, so I would change this to void and make another function called ServerMock.getLastPlayerQuitMessage() if someone would need it.

wuza8: I've actually made ServerMock.getLastPlayerQuitEvent() in case when Bukkit devs would add some other data to PlayerQuitEvent that someone would want to retrieve

seeseemelk:
@TheBusyBiscuit Do you still request changes to this?

Benlewis9000:
Is this ready to be merged or abandoned? I'm also interested in this feature.

wuza8:
@TheBusyBiscuit it was some time ago... take some time, but I would say implementation wouldn't never be perfect... :H

src/main/java/be/seeseemelk/mockbukkit/MockPlayerList.java
```diff
@@ -62,10 +62,19 @@ public BanList getProfileBans()
 
 	public void addPlayer(@NotNull PlayerMock player)
 	{
+		if(isOnline(player))
```
seeseemelk:
```suggestion
		if (isOnline(player))
```

src/main/java/be/seeseemelk/mockbukkit/MockPlayerList.java
```diff
@@ -95,6 +104,13 @@ public boolean isSomeoneOnline()
 		return !onlinePlayers.isEmpty();
 	}
 
+	public boolean isOnline(@NotNull PlayerMock player)
+	{
+		return 	onlinePlayers.contains(player)
```
seeseemelk:
```suggestion
		return onlinePlayers.contains(player)
```

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -84,6 +89,17 @@ void addPlayer_TwoPlayers_SizeIsTwo()
 		assertTrue(entities.contains(player2), "Player 2 was not registered");
 	}
 
+	@Test
+	void addPlayer_PlayersWithTheSameName()
+	{
+		server.addPlayer("Player");
+		assertThrows(IllegalStateException.class, ()->{
```
seeseemelk:
```suggestion
		assertThrows(IllegalStateException.class, () ->
		{
```

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -534,7 +550,104 @@ void testDefaultPotionEffects()
...
+		server.addPlayer(player1);
+		assertTrue(player1.canSee(player2));
+	}
+
+	@Test
+	void removePlayer_PlayerDataLoss_HeardSounds()
+	{
+		PlayerMock player = server.addPlayer();
+		player.playSound(new Location(null, 0, 0, 0),
+				Sound.AMBIENT_BASALT_DELTAS_ADDITIONS,1f,1f);
```
seeseemelk:
```suggestion
				Sound.AMBIENT_BASALT_DELTAS_ADDITIONS, 1f, 1f);
```

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -534,7 +550,104 @@ void testDefaultPotionEffects()
...
+		assertFalse(player.isOnline());
+		server.addPlayer(player);
+		assertTrue(player.isOnline());
+	}
 
+	@Test
+	void removePlayer_GetLastPlayerQuitEvent()
+	{
+		TestPlugin plugin = MockBukkit.load(TestPlugin.class);
+		server.getPluginManager().registerEvents(new Listener(){
```
seeseemelk:
```suggestion
		server.getPluginManager().registerEvents(new Listener()
		{
```

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -534,7 +550,104 @@ void testDefaultPotionEffects()
...
+			@EventHandler
+			public void onPlayerQuit(PlayerQuitEvent event){
```
seeseemelk:
```suggestion
			public void onPlayerQuit(PlayerQuitEvent event)
			{
```

seeseemelk (approved):
Other than some minor formatting inconsistencies it looks pretty good to me. I've DM'ed @TheBusyBiscuit to ask if he just forgot about it or if he still wants to see some changes.

seeseemelk (approved):

seeseemelk:
@thelooter or @Insprill, could I get an additional review on this one?

thelooter:
> @thelooter or @Insprill, could I get an additional review on this one?

Seems like this is a duplicate of #352

thelooter:
Closed as some of it's functionality is already fixed and the remaining Functionality should be added in a new PR

PR #224, opened 2021-06-03
Fix PlayerTeleportEvent not firing when player is teleported.

ohowe1:
# Description
Fixes the player teleport event not firing when a player is teleported

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -2181,18 +2181,6 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 	}
 
 	@Override
-	public boolean teleport(@NotNull Location location)
-	{
-		PlayerTeleportEvent playerTeleportEvent = new PlayerTeleportEvent(this, getLocation(), location);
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -2178,6 +2180,24 @@ public int getPing()
 		return 0;
 	}
 
+	@Override
+	public boolean teleport(@NotNull Location location, PlayerTeleportEvent.@NotNull TeleportCause cause)
```
```suggestion
	public boolean teleport(@NotNull Location location, @NotNull PlayerTeleportEvent.TeleportCause cause)
```

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):
Looks good to me thus far.
Bukkit does quite a few more things on teleporting though, such as closing the inventory, preventing the teleportation of dead players etc..
But I doubt these will be relevant in this simplified environment, so it should be good for now. Perhaps Seeseemelk wants to have a look at it before-hand though, so I'll wait on merging it until he had a look at it. But it is an approval from me.

seeseemelk (approved):
Looks pretty good. The only thing to change is perhaps the version number. It was bumped from 1.2.0 to 1.2.1, but it's not a bug fix so it should have been 1.3.0. I've gone ahead a fixed that though.

PR #227, opened 2021-06-13
[1.8-branch] Update BlockMock.java

ghost:
Implemented getRelative in MockBlock

# Description

# Checklist
- [ ] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
@@ -106,22 +106,22 @@ public class BlockMock implements org.bukkit.block.Block
 	@Override
 	public Block getRelative(int modX, int modY, int modZ)
 	{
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
I was confused at first since these methods are already implemented in MockBukkit, see:
https://github.com/MockBukkit/MockBukkit/blob/v1.16/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java#L118-L137

Then I realized that this PR is targeted towards the `v1.8` branch which we don't really support anymore to my knowledge.

ghost:
My plugin is targetting minecraft 1.8 so i need that version, when i use v1.16 it generates a package sealing violation on org.bukkit
pls continue developing for that version of minecraft

seeseemelk:
We're not actively maintaining this anymore, but I'll merge. I'm also going to required disable status checks for older branches as I think they're only functional for v1.16 anyway.

PR #228, opened 2021-06-13
Fixed a null file argument provided to the plugin constructor

zodiia:
# Description
Added a non-null `file` argument in the `JavaPlugin` constructor used by MockBukkit.

In plain Java, having a null argument works fine, even if it is annotated as `@NotNull`, because annotations are not checks, only annotations.
However, it will not work when plugins are developed using Kotlin. It has very strict null-checks in place, and especially with these kind of annotations. Using MockBukkit with Kotlin results in a NullPointerException thrown by the constructor, which only accepts non-null parameters.

This PR aims at adding a non-null `file` argument, so that this error will not be thrown.

Note: I cannot unit-test this, as the getter method to retrieve this file is protected.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
@@ -92,7 +92,7 @@ public class PluginManagerMock implements PluginManager
 		{
 			try
 			{
-				FileUtils.deleteDirectory(file);
+				FileUtils.forceDelete(file);
 			}
 			catch (IOException e)
 			{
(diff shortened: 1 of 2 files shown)


TheBusyBiscuit:
> Note: I cannot unit-test this, as the getter method to retrieve this file is protected.

As a side note: Yes you can.
`protected` means `package-protected`. If you use the same package for your unit test class as for the class being tested, you should have access to that method.

Not sure if such a "hacky" approach is worth that though.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -92,7 +92,7 @@ public void unload()
 		{
 			try
 			{
-				FileUtils.deleteDirectory(file);
+				FileUtils.forceDelete(file);
```
TheBusyBiscuit, line 95: Can you please elaborate on why this was changed?

TheBusyBiscuit (commented):
Looks fine to me I guess, just one clarification needed.

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
zodiia: The previous method only deleted directories, and would fail if provided with a file. The thing is, now we are deleting both directories AND files. Instead of putting an if-else if, this method will delete both directories and files (in fact, if provided a directory, it will call the previously used `deleteDirectory`). If needed, I can convert this to an if-else if, but I would find this not necessary, as `forceDelete` is already doing exactly what we need here.

zodiia:
@TheBusyBiscuit any update? I saw there was new merges and the version string has been updated, do I need to update it as well? If so I'll do it asap

TheBusyBiscuit (approved):
Looks good to me thus far

PR #229, opened 2021-06-13
Update ItemFactoryMock.java

Riley31415:
# Description
Corrected the functionality of `ItemFactoryMock#equals(ItemMeta, ItemMeta)` so that it returns true when both values are null.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemFactoryMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemFactoryMock.java
@@ -94,13 +94,13 @@ public class ItemFactoryMock implements ItemFactory
 	@Override
 	public boolean equals(ItemMeta meta1, ItemMeta meta2)
 	{
-		if (meta1 != null && meta2 != null)
+		if (meta1 == null)
 		{
-			return meta1.equals(meta2);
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/inventory/ItemFactoryMock.java
```diff
@@ -94,13 +94,13 @@ public boolean isApplicable(ItemMeta meta, Material material)
...
+		if (meta1 == null)
 		{
-			return meta1.equals(meta2);
+			return meta2 == null;
 		}
 		else
 		{
-			return false;
+			return meta1.equals(meta2);
 		}
```
TheBusyBiscuit: This can probably be simplified to
`return Objects.equals(meta1, meta2);`
which does null checks too and returns true if both are null.

TheBusyBiscuit: Also a comment explaining this behaviour would be nice :)

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):
LGTM

PR #230, opened 2021-06-23
[1.8-branch] Implemented most of the methods in 1.8-spigot

ghost:
# Description
I implemented most of the methods for 1.8
# Checklist
- [ 1.1.4 ] Version number updated in `settings.gradle`.
- [ yes ] Unit tests added.
- [ yes ] Code follows existing code format.


repositoryPassword.txt
```diff
@@ -0,0 +1 @@
+tiagodinis123
```
wuza8: What's that?

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -1,22 +1,10 @@
...
-import org.bukkit.Bukkit;
-import org.bukkit.EntityEffect;
-import org.bukkit.Location;
-import org.bukkit.Server;
-import org.bukkit.World;
+import be.seeseemelk.mockbukkit.MockBukkit;
+import be.seeseemelk.mockbukkit.UnimplementedOperationException;
+import be.seeseemelk.mockbukkit.command.MessageTarget;
+import be.seeseemelk.mockbukkit.metadata.MetadataTable;
+import org.bukkit.*;
```
wuza8: Avoid wildcard imports

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -14,16 +14,9 @@
 import be.seeseemelk.mockbukkit.scheduler.BukkitSchedulerMock;
 import be.seeseemelk.mockbukkit.scoreboard.ScoreboardManagerMock;
 import com.avaje.ebean.config.ServerConfig;
-import org.bukkit.BanEntry;
-import org.bukkit.BanList;
+import org.bukkit.*;
```
wuza8: Avoid wildcard imports

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -28,552 +16,488 @@
 import org.bukkit.plugin.Plugin;
 import org.bukkit.util.Vector;
 
-import be.seeseemelk.mockbukkit.UnimplementedOperationException;
-import be.seeseemelk.mockbukkit.command.MessageTarget;
-import be.seeseemelk.mockbukkit.metadata.MetadataTable;
+import java.util.*;
```
wuza8: Avoid wildcard imports

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -28,552 +16,488 @@
...
+import java.util.concurrent.LinkedTransferQueue;
+
+import static org.junit.Assert.*;
```
wuza8: Avoid wildcard imports

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -28,552 +16,488 @@
...
+     * @param name The new name of the entity.
+     */
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    @Override
+    public void sendMessage(String message) {
+        messages.add(message);
+        System.out.println(this.name + ": " + message);
```
wuza8: Feels like testing leftover

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -28,552 +16,488 @@
...
+    @Override
+    public boolean isPermissionSet(Permission perm) {
+
+        for (PermissionAttachment attachment : permissionAttachments) {
+            Map<String, Boolean> permissions = attachment.getPermissions();
+            if (permissions.containsKey(perm.getName()))
+                return true;
+        }
+        return false;
+    }
```
wuza8: The same code, it's fixed in v1.16

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -28,552 +16,488 @@
...
+
+    @Override
+    public boolean isPermissionSet(Permission perm) {
+
+        for (PermissionAttachment attachment : permissionAttachments) {
+            Map<String, Boolean> permissions = attachment.getPermissions();
+            if (permissions.containsKey(perm.getName()))
+                return true;
+        }
+        return false;
```
wuza8:
```suggestion
        return isPermissionSet(perm.getName().toLowerCase(Locale.ENGLISH));
```

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -28,552 +16,488 @@
...
+
+    @Override
+    public List<Entity> getNearbyEntities(double x, double y, double z) {
+        // TODO Auto-generated constructor stub
+        throw new UnimplementedOperationException();
+    }
+
+    @Override
+    public int getEntityId() {
+        return new ArrayList<>(MockBukkit.getMock().getEntities()).indexOf(this);
```
wuza8: This would not give entityId, because if an entity would be removed from this set, indexes would be shifted, You also don't really know if entities in this set are sorted or not, so for example we have an entity that has hashcode "Zombie", you use getEntityId and it gives you 1, then you add entity with hashcode "Player" and now if you would use getEntityId on "Zombie", it would give you 2

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -51,19 +51,14 @@
...
 import java.net.InetSocketAddress;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.EnumMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
wuza8: Avoid wildcard imports

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -28,552 +16,488 @@
...
+import java.util.concurrent.LinkedTransferQueue;
+
+import static org.junit.Assert.*;
+
+public abstract class EntityMock implements Entity, MessageTarget {
+    private final UUID uuid;
+    private final Queue<String> messages = new LinkedTransferQueue<>();
+    private final Set<PermissionAttachment> permissionAttachments = new HashSet<>();
+    String customName = null;
+    EntityDamageEvent e = null;
```
wuza8:
```suggestion
    EntityDamageEvent lastEntityDamageEvent = null;
```
I would say it's better name for that.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -240,12 +235,13 @@ public void damage(double amount)
 		EntityDamageEvent event = new EntityDamageEvent(this, DamageCause.CUSTOM, modifiers, modifierFunctions);
 		event.setDamage(amount);
 		Bukkit.getPluginManager().callEvent(event);
+		e = event;
```
wuza8: This line should be moved to if(!event.isCancelled()), because if this event would be cancelled, the methods EntityMock::getLastDamageCause and PlayerMock::getLastDamage would output false data

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -240,12 +235,13 @@ public void damage(double amount)
...
+
 		if (!event.isCancelled())
 		{
 			setHealth(health - amount);
 		}
 	}
-	
 	@Override
 	public void damage(double amount, Entity source)
 	{
```
wuza8: Those two functions do nearly the same, it would be nice to somehow fix this redundancy.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -368,26 +365,23 @@ public void setItemInHand(ItemStack item)
...
 		getInventory().setItemInHand(item);
 	}
-	
+	ItemStack itemOnCursor;
```
wuza8: There are many errors like this in the entire PlayerMock class - all global variables should be on top of a class and they should be PRIVATE, because if not, any other class would be able to change and use them as their own.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -414,22 +408,19 @@ public int getExpToLevel()
 	@Override
 	public double getEyeHeight()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return 1.6;
 	}
 	
 	@Override
 	public double getEyeHeight(boolean ignorePose)
```
wuza8: Argument ignorePose is ignored

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -498,78 +489,86 @@ public void setMaximumNoDamageTicks(int ticks)
...
 	@Override
 	public void setNoDamageTicks(int ticks)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		lastTimeDamage = System.currentTimeMillis()-ticks;
 	}
 	
 	@Override
 	public Player getKiller()
```
wuza8: getKiller() method would give last damage causer even if player wasn't killed

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -610,15 +609,13 @@ public void setCanPickupItems(boolean pickup)
...
-		throw new UnimplementedOperationException();
+		return true;
 	}
 	
 	@Override
 	public boolean isLeashed()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return false;
```
wuza8:
```suggestion
		// Players can not be leashed
		return false;
```
Appropriate comment

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -701,8 +698,7 @@ public long getLastPlayed()
 	@Override
 	public boolean hasPlayedBefore()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return false;
```
wuza8: That wouldn't be true used with removePlayer and addPlayer methods

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1144,40 +1135,39 @@ public void giveExp(int amount)
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
-	
+	int level;
+	float xp;
 	@Override
 	public void giveExpLevels(int amount)
```
wuza8: This implementation doesn't call PlayerLevelChangeEvent and I don't know if this function is the same in 1.16 but in that version it also makes xp = 0.0f.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
wuza8: Keep global variables private.

src/main/java/be/seeseemelk/mockbukkit/inventory/ChestStateMock.java
```diff
@@ -0,0 +1,26 @@
+package be.seeseemelk.mockbukkit.inventory;
+
+import be.seeseemelk.mockbukkit.block.BlockStateMock;
+import org.bukkit.block.Chest;
+import org.bukkit.inventory.Inventory;
+import org.bukkit.material.MaterialData;
+
+public class ChestStateMock extends BlockStateMock implements Chest {
+    ChestInventoryMock inventory = new ChestInventoryMock(this, "Baú", 9*3);
```
wuza8: Should be private plus is this "Baú" valid?

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
```diff
@@ -251,161 +253,214 @@ public InventoryType getType()
...
 	}
 	
 	@Override
-	public void setMaxStackSize(int size)
-	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
-		
-	}
+	public void setMaxStackSize(int size) {}
```
wuza8: Doesn't feel like a valid implementation

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
```diff
@@ -251,161 +253,214 @@ public InventoryType getType()
...
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		for (int i = 0; i < this.items.length; i++) {
+			if(items[i].equals( this.items[i] )){
+				items[i] = null;
+			}
+		}
+		return new HashMap<>();
+
```
wuza8: This suspicious return statement gives me impression, that this doesn't work as it should.

src/test/java/be/seeseemelk/mockbukkit/block/BlockMockTest.java
```diff
@@ -1,16 +1,14 @@
...
-import static org.junit.Assert.assertNull;
-
 import org.bukkit.Location;
 import org.bukkit.Material;
 import org.junit.Before;
 import org.junit.Test;
 
 import be.seeseemelk.mockbukkit.WorldMock;
 
+import static org.junit.Assert.*;
```
wuza8, line 10: Avoid wildcard imports

wuza8 (changes requested):
Many implementations have no unit tests, some of them are faulty and the changes touches 800+ lines of code, so there are many things to be fixed before merging.

src/main/java/be/seeseemelk/mockbukkit/inventory/ChestStateMock.java
ghost: oh sorry im portuguese i had put chest in portuguese which is baú

repositoryPassword.txt
ghost: oh god is my repository's password frick

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
ghost: i think every inventory has a 64 max stack size so i've assumed that

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
ghost: well, it worked on my projects on PatzMC

ghost: but i'll fix that return statement

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
ghost: intellij puts a wildcard import automatically when i use many classes from a package

repositoryPassword.txt
seeseemelk: You might be interested in knowing that there is a file called `.git/info/exclude`. It works exactly like `.gitignore`, but it's local to a single clone of a repository. It never gets committed and pushed.
Adding the line `repositoryPassword.txt` to your `.git/info/exclude` is an easy way to prevent you accidentally committing it without having to modify the repo's `.gitignore`.

ghost: hmm thx

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
ghost: i addedthis sysout because its strange not logging the messages, but if you dont like, sure

src/main/java/be/seeseemelk/mockbukkit/inventory/ChestStateMock.java
ghost: Fixed

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
ghost: hmm, what if i stored the index when the class is constructed and then i just return that stored index

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: no,it returns the player who killed on the last death event

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: oh yeah, i added removePlayer and then i have forgotten, lemme change that

ghost:
@AyBiCi Now i believe i solved all the problems, can you check it pls?

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
wuza8: This would not fix the problem with removing entities from list:
- we add two entities to a list and they get id 1 and 2 (2 entities in a list, the next one would get id = 3)
- we remove entity number 1 (1 entity in a list with id = 2, the next one would get id = 2)
- we add new entity, but this entity gets index number 2 and we have two entities with id = 2

Better than storing index would be storing the value that would be never the same, like this code would give us:
```java
private static int nextId = 1; //static value has only one instance for all objects in a class

private static int getNextId(){
    return nextId++;  //post-incrementation
}
```

wuza8:
I will do it ASAP

src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
```diff
@@ -251,16 +255,16 @@ public void setType(Material type, boolean applyPhysics)
 	@Deprecated
 	public boolean setTypeId(int type)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Material old = material;
+		this.material = Material.getMaterial(type);
```
wuza8:
```suggestion
		setType(material);
```
BlockMock.setType(Material) and BlockMock.setTypeId(int) should actually do the same thing, but it would not do, when I would use chest ID.

src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
```diff
@@ -344,14 +348,12 @@ public int getBlockPower()
 	@Override
 	public boolean isEmpty()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return this.getTypeId() == 0;
```
wuza8:
```suggestion
		return this.getType().equals(Material.AIR);
```
This smells like a magic value a little

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -28,552 +16,482 @@
...
+    private String name = "entity";
+    private int id;
+    public EntityMock(UUID uuid) {
+        this.uuid = uuid;
+
+        if (Bukkit.getWorlds().size() > 0)
+            location = Bukkit.getWorlds().get(0).getSpawnLocation();
+        else
+            location = new Location(null, 0, 0, 0);
+        this.id = new ArrayList<>(MockBukkit.getMock().getEntities()).indexOf(this);
```
wuza8: This would not fix the problem with removing entities from list:

- we add two entities to a list and they get id 1 and 2 (2 entities in a list, the next one would get id = 3)
- we remove entity number 1 (1 entity in a list with id = 2, the next one would get id = 2)
- we add new entity, but this entity gets index number 2 and we have two entities with id = 2

Better than storing index would be storing the value that would be never the same, like this code would give us:

```java
private static int nextId = 1; //static value has only one instance for all objects in a class

private static int getNextId(){
    return nextId++;  //post-incrementation
}
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1144,40 +1140,39 @@ public void giveExp(int amount)
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
-	
+	int level;
+	float xp;
```
wuza8: These ones are not private

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -498,78 +491,86 @@ public void setMaximumNoDamageTicks(int ticks)
...
 	
 	@Override
 	public void setLastDamage(double damage)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		lastEntityDamageEvent.setDamage(damage);
 	}
-	
+	long lastTimeDamage;
```
wuza8: This is not private

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -498,78 +491,86 @@ public void setMaximumNoDamageTicks(int ticks)
...
 	@Override
 	public int getNoDamageTicks()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return (int) ((System.currentTimeMillis()-lastTimeDamage)/1000)/20;
```
wuza8:
```suggestion
		return (int) ((System.currentTimeMillis()-lastTimeDamage)/(1000/20));
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -498,78 +491,86 @@ public void setMaximumNoDamageTicks(int ticks)
...
 	}
 	
 	@Override
 	public void setNoDamageTicks(int ticks)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		lastTimeDamage = System.currentTimeMillis()-ticks;
```
wuza8:
```suggestion
		lastTimeDamage = System.currentTimeMillis()-ticks*(1000/20);
```
Every tick cost 1000/20 = 50 ms

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -498,78 +491,86 @@ public void setMaximumNoDamageTicks(int ticks)
...
-		throw new UnimplementedOperationException();
+		if(e instanceof EntityDamageByEntityEvent){
+			EntityDamageByEntityEvent ev = ((EntityDamageByEntityEvent) lastEntityDamageEvent);
+
+			if(ev.getDamager() instanceof Player){
+				return ((Player) ev.getDamager());
+			}
+		}
+		return null;
 	}
```
wuza8: I've meant this implementation would give last damage causer (Player class) instead of last killer and damaging player doesn't mean killing him

wuza8: This "e" should have a better name, I don't really know what it is and can't find it

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1260,50 +1253,44 @@ public void setBedSpawnLocation(Location location, boolean force)
 	@Override
 	public boolean getAllowFlight()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return canFly;
 	}
-	
+	boolean canFly;
+	boolean isFlying;
```
wuza8: Not private

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1260,50 +1253,44 @@ public void setBedSpawnLocation(Location location, boolean force)
...
 	
 	@Override
 	public void showPlayer(Player player)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		hiddenPlayers.remove(this,player);
 	}
-	
+	HashMap<Player,Player> hiddenPlayers = new HashMap<>();
```
wuza8: Not private

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1351,15 +1338,13 @@ public void setResourcePack(String url)
 	@Override
 	public Scoreboard getScoreboard()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return scoreboard;
 	}
-	
+	Scoreboard scoreboard;
```
wuza8: Not private

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1260,50 +1253,44 @@ public void setBedSpawnLocation(Location location, boolean force)
...
 	}
-	
+	HashMap<Player,Player> hiddenPlayers = new HashMap<>();
 	@Override
 	public boolean canSee(Player player)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return hiddenPlayers.containsKey(this) && hiddenPlayers.containsValue(player);
 	}
```
wuza8: In implementation of this 3 methods there's no need for a HashMap, because every time the key would be just "this", maybe this would make any sense if this hiddenPlayers HashMap would be static, but the code for canSee(Player) method would not work. It would be much more logical to just create a List<Player> that would store hidden players for each player.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1260,50 +1253,44 @@ public void setBedSpawnLocation(Location location, boolean force)
...
+		return this.isFlying;
 	}
 	
 	@Override
 	public void setFlying(boolean value)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		this.isFlying = true;
 	}
```
wuza8: Argument is ignored :face_with_thermometer:

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
```diff
@@ -251,161 +253,223 @@ public InventoryType getType()
...
 	}
 	
 	@Override
-	public void setMaxStackSize(int size)
-	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
-		
-	}
+	public void setMaxStackSize(int size) {}
```
wuza8: I've meant InventoryMock.setMaxStackSize(int) method. This method sets a limit of how many items can be in a stack, in a certain inventory, but there's no implementation of that (doesn't need to be implemented, but it should have this old UnimplementedOperationException throw and TODO comment instead

src/test/java/be/seeseemelk/mockbukkit/scoreboard/ScoreMockTest.java
```diff
@@ -64,7 +64,6 @@ public void getScore_ObjectiveRegisteredAndScoreSet_ReturnsNumber()
 		score.setScore(5);
 		assertEquals(5, score.getScore());
 	}
-
```
wuza8, line 67: Random backspace xd

wuza8 (commented):
There are some more problems that are here and there:
- wildcard imports are still there (the best way to fix that is to find all occurrences of them, get normal imports back and add missing imports by hand or with turning off options in IDE and fixing them automatically)
- automatic tools broke build.gradle (the easiest thing would be just remove all changes from it go back to the previous build.gradle, also there are two different sources for JUnit in there)
- EntityMock changes are broken and I don't really know how to fix it, it feels like somehow Git just done a bad job or something. for now I don't see any other fix but just go back to old EntityMock file and somehow remake all the changes but probably there's an other, easier way to somehow recalculate changes? 
- In PlayerMock every field you made private, should also be on top of a file

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
ghost: i have to say, this is tricky

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
ghost: done

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
ghost: what i made is similar to that, i made a nextID and in the contructor, I increment by 1 and sets it as the entityID
```java
private static int nextID = 0;
public EntityMock(...){
    //stuff
    nextID++;
    this.id = nextID;
}
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: i dont think this works

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: Fixed, now i created a field that is set to the damageEvent when health is less or equal 0

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: Fixed

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: hmm, thx i'll fix now

ghost: Done

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: Fixed

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: Fixed

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: bruh im dumb lol

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: fixed

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: fixed

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
ghost: I'll make a set, because i dont want repeated elements in the list

ghost: Done

tiagodinis33:
Lol, intellij has a context action to split imports

ghost:
ok, now its done @AyBiCi

wuza8:
Something is wrong with PlayerMock and EntityMock files. If you would go to files changed section, every line is marked as changed in those files. Because of that there's no way of reviewing some changes, because it would get to the point of reviewing 2000 lines of code that was reviewed and tested before. There needs to be some way to fix that, but I don't have idea how. You can post that error in MockBukkit discord, maybe someone would know the answer, or someone would post it here.

ghost:
that's reasonable, because there was many implementations missing

PR #235, opened 2021-07-11
PlayerMock.simulatePlayerMove fix for teleportation in PlayerMoveEvent

wuza8:
# Description
This small rearrange of code fixes event execution. 

# Checklist
- [x] Unit tests added.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -14,7 +14,7 @@ gradle.ext.apiVersionFull = '1.16.5-R0.1-SNAPSHOT'
   Every new update should bump the version below according to
   the conventions.
 */
-gradle.ext.version = '1.5.0'
+gradle.ext.version = '1.5.1'
 
 /*
   This is the name of our MockBukkit artifact, it includes
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -292,9 +292,10 @@ public class PlayerMock extends LivingEntityMock implements Player, SoundReceive
 	public @NotNull PlayerMoveEvent simulatePlayerMove(@NotNull Location moveLocation)
 	{
 		PlayerMoveEvent event = new PlayerMoveEvent(this, this.getLocation(), moveLocation);
+		this.setLocation(event.getTo());
 		Bukkit.getPluginManager().callEvent(event);
-		if (!event.isCancelled())
-			this.setLocation(event.getTo());
+		if (event.isCancelled())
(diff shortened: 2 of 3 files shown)


TheBusyBiscuit:
You didn't actually explain what this fixes though.

wuza8:
Before when PlayerMoveEvent listener would teleport a Player, Player would not be teleported when event would not be cancelled and if event would be cancelled, he would be teleported. This fixes it how it should be - when event would be not canceled Player would be teleported, if event would be canceled player would not be teleported.

wuza8:
So when we have a PlayerMoveEvent listener in our plugin, that does a teleportation inside like this one
```java @EventHandler
			public void onPlayerMove(PlayerMoveEvent event){
                                ...
                                if( blockPlayerSteppedOn.getType().equals(Material.DIRT))
				    event.getPlayer().teleport(teleportLocation);
                                ...
			} 
```
Now what we want to accomplish with a code like this is for example is when a player steps on a dirt block we want to teleport player somewhere and it works in normal bukkit like this.

But because of a false arrangement of player teleportation in simulatePlayerMove method we get to this kind of a code execution:

- player.simulatePlayerMove(simulatingLocation) is invoked
- simulatingLocation gets into event.getTo()
- simulatePlayerMove invokes PlayerMoveEvent
- it looks like player stands on a DIRT block, so player gets teleported to our special location
- simulatePlayerMove gets executed event with setCancelled = false
- !!!because of setCancelled == false simulatePlayerMove SETS PLAYER LOCATION TO event.getTo() THAT IS simulatingLocation SO TELEPORTATION GETS LOST ;C!!!

And that's not how bukkit works...

After this fix the execution of simulatePlayerMove goes like this:

- player.simulatePlayerMove(simulatingLocation) is invoked
- simulatingLocation gets into event.getTo()
- simulatePlayerMove SETS LOCATION OF PLAYER - player.setLocation(simulatingLocation)
- simulatePlayerMove invokes PlayerMoveEvent
- it looks like player stands on a DIRT block, so player gets teleported to our special location
- simulatePlayerMove gets executed event with setCancelled = false
- because of setCancelled == false simulatePlayerMove DOES NOT SET LOCATION OF PLAYER to simulatingLocation

After some tests made with real Bukkit I've come to a conclusion that's how it works and this fix makes that small change

If we would be going with event.setCancelled(true) scenario, the teleportation invoked in PlayerMoveEvent listener would be cancelled by invoking player.setLocation(event.getFrom()) and that also goes like that in Bukkit.

wuza8:
@TheBusyBiscuit Fixed lack of explanation.

seeseemelk (approved):

PR #238, opened 2021-07-21
Add chunk data mocks

Seggan:
# Description
Added `MockChunkData` to MockBukkit. The implementation isn't really what I'd like it to be but it works

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
@@ -91,7 +91,7 @@ public class MockChunkData implements ChunkGenerator.ChunkData
 		        z >= 0 && z < CHUNK_SIZE
 		   )
 		{
-			BlockData data = blocks[x][y][z];
+			BlockData data = blocks[x][y - this.minHeight][z];
 			// shortcut to return air directly instead of creating air block data then unpacking material
 			return data == null ? Material.AIR : data.getMaterial();
 		}
@@ -115,7 +115,7 @@ public class MockChunkData implements ChunkGenerator.ChunkData
 		        z >= 0 && z < CHUNK_SIZE
 		   )
 		{
-			BlockData data = blocks[x][y][z];
+			BlockData data = blocks[x][y - this.minHeight][z];
 			return data == null ? new BlockDataMock(Material.AIR) : data;
 		}
 
--- a/src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
@@ -8,57 +8,72 @@ import org.bukkit.generator.ChunkGenerator;
 import org.bukkit.material.MaterialData;
 import org.jetbrains.annotations.NotNull;
 
-public class MockChunkData implements ChunkGenerator.ChunkData {
+public class MockChunkData implements ChunkGenerator.ChunkData
+{
 
 	private final BlockData[][][] blocks;
 
 	private final int maxHeight;
 	private final int minHeight; // unused but will be useful in 1.17/1.18
 
-	public MockChunkData(World world) {
+	public MockChunkData(World world)
+	{
 		this.maxHeight = world.getMaxHeight();
 		this.minHeight = world.getMinHeight();
 		blocks = new BlockData[16][this.maxHeight][16];
 	}
 
 	@Override
-	public int getMaxHeight() {
+	public int getMaxHeight()
+	{
 		return this.maxHeight;
 	}
 
 	@Override
-	public void setBlock(int x, int y, int z, @NotNull Material material) {
+	public void setBlock(int x, int y, int z, @NotNull Material material)
+	{
 		this.setBlock(x, y, z, new BlockDataMock(material));
 	}
 
 	@Override
-	public void setBlock(int x, int y, int z, @NotNull MaterialData material) {
+	public void setBlock(int x, int y, int z, @NotNull MaterialData material)
+	{
 		this.setBlock(x, y, z, new BlockDataMock(material.getItemType()));
 	}
 
 	@Override
-	public void setBlock(int x, int y, int z, @NotNull BlockData blockData) {
-		try {
+	public void setBlock(int x, int y, int z, @NotNull BlockData blockData)
+	{
+		try
+		{
 			blocks[x][y][z] = blockData;
-		} catch (ArrayIndexOutOfBoundsException ignored) {
+		}
+		catch (ArrayIndexOutOfBoundsException ignored)
+		{
 		}
 	}
 
 	@Override
-	public void setRegion(int xMin, int yMin, int zMin, int xMax, int yMax, int zMax, @NotNull Material material) {
+	public void setRegion(int xMin, int yMin, int zMin, int xMax, int yMax, int zMax, @NotNull Material material)
+	{
 
 	}
 
 	@Override
-	public void setRegion(int xMin, int yMin, int zMin, int xMax, int yMax, int zMax, @NotNull MaterialData material) {
+	public void setRegion(int xMin, int yMin, int zMin, int xMax, int yMax, int zMax, @NotNull MaterialData material)
+	{
 
 	}
 
 	@Override
-	public void setRegion(int xMin, int yMin, int zMin, int xMax, int yMax, int zMax, @NotNull BlockData blockData) {
-		for (int x = xMin; x < xMax; x++) {
-			for (int y = yMin; y < yMax; y++) {
-				for (int z = zMin; z < zMax; z++) {
+	public void setRegion(int xMin, int yMin, int zMin, int xMax, int yMax, int zMax, @NotNull BlockData blockData)
+	{
+		for (int x = xMin; x < xMax; x++)
+		{
+			for (int y = yMin; y < yMax; y++)
+			{
+				for (int z = zMin; z < zMax; z++)
+				{
 
 				}
 			}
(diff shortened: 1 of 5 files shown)


src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,126 @@
...
+	@Override
+	public void setBlock(int x, int y, int z, @NotNull BlockData blockData)
+	{
+		try
+		{
+			blocks[x][y][z] = blockData;
+		}
+		catch (ArrayIndexOutOfBoundsException ignored)
+		{
+		}
```
TheBusyBiscuit: Don't use a try/catch like that for god sake.
Java is a language which supports the use of `if`-statements to make code run conditionally, you know.

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,126 @@
...
+		try
+		{
+			BlockData data = blocks[x][y][z];
+			// shortcut to return air directly instead of creating air block data then unpacking material
+			return data == null ? Material.AIR : data.getMaterial();
+		}
+		catch (ArrayIndexOutOfBoundsException e)
+		{
+			return Material.AIR;
+		}
```
TheBusyBiscuit: Just NO...

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,126 @@
...
+	{
+		try
+		{
+			BlockData data = blocks[x][y][z];
+			return data == null ? new BlockDataMock(Material.AIR) : data;
+		}
+		catch (ArrayIndexOutOfBoundsException e)
+		{
+			return new BlockDataMock(Material.AIR);
+		}
```
TheBusyBiscuit: ...

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,126 @@
...
+import org.bukkit.material.MaterialData;
+import org.jetbrains.annotations.NotNull;
+
+public class MockChunkData implements ChunkGenerator.ChunkData
+{
+
+	private final BlockData[][][] blocks;
+
+	private final int maxHeight;
+	private final int minHeight; // unused but will be useful in 1.17/1.18
```
TheBusyBiscuit: Please do not place code and comments on the same line.

TheBusyBiscuit: So... why did you choose to place max before min here...? It would make much more sense the other way.

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,126 @@
...
+
+	public MockChunkData(World world)
```
TheBusyBiscuit: `@NotNull`

TheBusyBiscuit (changes requested):
One of the worst code snippets / PRs I have seen in a long time.

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
Seggan: i added it after

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,129 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public class MockChunkData implements ChunkGenerator.ChunkData
+{
+
+	private final BlockData[][][] blocks;
+
+	// unused but will be useful in 1.17/1.18
+	private final int minHeight;
+	private final int maxHeight;
```
seeseemelk: How are these unused?

seeseemelk: Ah, only the `minHeight` is unused

seeseemelk:
In it's current state it looks pretty good to me. How about you @TheBusyBiscuit?

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,129 @@
...
+	public void setBlock(int x, int y, int z, @NotNull MaterialData material)
+	{
+		this.setBlock(x, y, z, new BlockDataMock(material.getItemType()));
+	}
+
+	@Override
+	public void setBlock(int x, int y, int z, @NotNull BlockData blockData)
+	{
+		if (x >= 0 && x < 16 &&
+		        y >= 0 && y < this.maxHeight &&
```
TheBusyBiscuit: Should be `minHeight` and not `0`

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,129 @@
...
+import org.bukkit.generator.ChunkGenerator;
+import org.bukkit.material.MaterialData;
+import org.jetbrains.annotations.NotNull;
+
+public class MockChunkData implements ChunkGenerator.ChunkData
+{
+
+	private final BlockData[][][] blocks;
+
+	// unused but will be useful in 1.17/1.18
```
TheBusyBiscuit: This comment should be removed really.

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,129 @@
...
+			}
+		}
+	}
+
+	@NotNull
+	@Override
+	public Material getType(int x, int y, int z)
+	{
+		if (x >= 0 && x < 16 &&
+		        y >= 0 && y < this.maxHeight &&
```
TheBusyBiscuit: Should be `minHeight` and not `0`

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,129 @@
...
+	{
+		return new MaterialData(this.getType(x, y, z));
+	}
+
+	@NotNull
+	@Override
+	public BlockData getBlockData(int x, int y, int z)
+	{
+		if (x >= 0 && x < 16 &&
+		        y >= 0 && y < this.maxHeight &&
```
TheBusyBiscuit: Should be `minHeight` and not `0`

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,129 @@
...
+		        z >= 0 && z < 16
```
TheBusyBiscuit: The number `16` is used a **lot** here. While this number is certainly not gonna change anytime soon...
So-called "Magic numbers" are always best to be avoided.
I think a `private static final int CHUNK_SIZE = 16;` field would be better.

TheBusyBiscuit (changes requested):
Much better but I don't really see a reason not to use the min height for y here. The API has it and after all, it will be 0 by default anyway.

TheBusyBiscuit:
> In it's current state it looks pretty good to me. How about you @TheBusyBiscuit?

Certainly looks much better, I left a review with more details.

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,130 @@
...
+	private final BlockData[][][] blocks;
+
+	private final int minHeight;
+	private final int maxHeight;
+
+	public MockChunkData(@NotNull World world)
+	{
+		this.minHeight = world.getMinHeight();
+		this.maxHeight = world.getMaxHeight();
+		blocks = new BlockData[CHUNK_SIZE][this.maxHeight][CHUNK_SIZE];
```
TheBusyBiscuit:
```suggestion
		blocks = new BlockData[CHUNK_SIZE][this.maxHeight - this.minHeight][CHUNK_SIZE];
```

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,130 @@
...
+
+	@Override
+	public void setBlock(int x, int y, int z, @NotNull BlockData blockData)
+	{
+		if (x >= 0 && x < CHUNK_SIZE &&
+		        y >= this.minHeight && y < this.maxHeight &&
+		        z >= 0 && z < CHUNK_SIZE
+		   )
+		{
+			blocks[x][y][z] = blockData;
```
TheBusyBiscuit:
```suggestion
			blocks[x][y - this.minHeight][z] = blockData;
```

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,130 @@
...
+	@NotNull
+	@Override
+	public Material getType(int x, int y, int z)
+	{
+		if (x >= 0 && x < CHUNK_SIZE &&
+		        y >= this.minHeight && y < this.maxHeight &&
+		        z >= 0 && z < CHUNK_SIZE
+		   )
+		{
+			BlockData data = blocks[x][y][z];
```
TheBusyBiscuit:
```suggestion
			BlockData data = blocks[x][y - this.minHeight][z];
```

src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
```diff
@@ -0,0 +1,130 @@
...
+	@NotNull
+	@Override
+	public BlockData getBlockData(int x, int y, int z)
+	{
+		if (x >= 0 && x < CHUNK_SIZE &&
+		        y >= this.minHeight && y < this.maxHeight &&
+		        z >= 0 && z < CHUNK_SIZE
+		   )
+		{
+			BlockData data = blocks[x][y][z];
```
TheBusyBiscuit:
```suggestion
			BlockData data = blocks[x][y - this.minHeight][z];
```

TheBusyBiscuit (changes requested):
I haven't tested this yet but this should add support for minHeights to be properly handled.
Can you perhaps write two Unit Tests for this?
One with a min height of -64 and one with a min height of +64, so we know this is handled correctly?

src/test/java/be/seeseemelk/mockbukkit/MockChunkDataTest.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	@Test
+	void test_set_out_of_bounds()
+	{
+		WorldMock dummy = server.addSimpleWorld("dummy");
+		ChunkGenerator.ChunkData data = server.createChunkData(dummy);
+
+		data.setBlock(33, 1000, 33, Material.STONE);
+	}
```
TheBusyBiscuit: There are no assertions in this test, at least an `assertDoesNotThrow` would be minimum.

TheBusyBiscuit (changes requested):
One test is broken and I would like to have two tests for varying min height, otherwise this looks good

src/test/java/be/seeseemelk/mockbukkit/MockChunkDataTest.java
```diff
@@ -0,0 +1,98 @@
...
+		Assertions.assertEquals(Material.STONE, data.getType(0, 0, 0));
+	}
+
+	@Test
+	void test_set_out_of_bounds()
+	{
+		WorldMock dummy = server.addSimpleWorld("dummy");
+		ChunkGenerator.ChunkData data = server.createChunkData(dummy);
+
+		Assertions.assertDoesNotThrow(() -> data.setBlock(33, 1000, 33, Material.STONE));
```
TheBusyBiscuit, line 42: Perhaps an additional assertion to make sure the type is air would be nice.
But that's about the last improvement I can think of

TheBusyBiscuit (changes requested):

src/test/java/be/seeseemelk/mockbukkit/MockChunkDataTest.java
Seggan: i already check for the air type in `test_get_out_of_bounds`

TheBusyBiscuit: That is a different test case though.

TheBusyBiscuit (approved):
LGTM

TheBusyBiscuit (approved):
Now targeting `v1.17`

TheBusyBiscuit (approved):
👍🏻 Thanks

PR #239, opened 2021-07-27
MC 1.17 basic upgrade

original-codematrix:
# Description
MC 1.17
Java 16
Gradle 7.1.1
Dependency upgrades:
- spigot-api
- junit-jupiter
- hamcrest
- commons-io
- jetbrains annotations

# Checklist
- [ ] Unit tests added.


src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -83,17 +42,13 @@
...
-import be.seeseemelk.mockbukkit.inventory.PlayerInventoryMock;
-import be.seeseemelk.mockbukkit.inventory.PlayerInventoryViewMock;
-import be.seeseemelk.mockbukkit.inventory.SimpleInventoryViewMock;
-import be.seeseemelk.mockbukkit.sound.AudioExperience;
-import be.seeseemelk.mockbukkit.sound.SoundReceiver;
-import net.md_5.bungee.api.ChatMessageType;
-import net.md_5.bungee.api.chat.BaseComponent;
+import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.util.*;
```
TheBusyBiscuit: Please avoid wildcard imports

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -59,22 +31,9 @@
...
-import org.bukkit.event.player.PlayerRespawnEvent;
-import org.bukkit.event.player.PlayerTeleportEvent;
-import org.bukkit.event.player.PlayerToggleFlightEvent;
-import org.bukkit.event.player.PlayerToggleSneakEvent;
-import org.bukkit.event.player.PlayerToggleSprintEvent;
-import org.bukkit.inventory.EntityEquipment;
-import org.bukkit.inventory.Inventory;
-import org.bukkit.inventory.InventoryView;
+import org.bukkit.event.player.*;
+import org.bukkit.inventory.*;
```
TheBusyBiscuit: Please avoid wildcard imports

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -46,11 +22,7 @@
 import org.bukkit.block.data.BlockData;
 import org.bukkit.conversations.Conversation;
 import org.bukkit.conversations.ConversationAbandonedEvent;
-import org.bukkit.entity.Entity;
-import org.bukkit.entity.EntityType;
-import org.bukkit.entity.Player;
-import org.bukkit.entity.Pose;
-import org.bukkit.entity.Villager;
+import org.bukkit.entity.*;
```
TheBusyBiscuit: Please avoid wildcard imports

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1,43 +1,19 @@
...
-import org.bukkit.Material;
-import org.bukkit.NamespacedKey;
-import org.bukkit.Note;
-import org.bukkit.Particle;
-import org.bukkit.Sound;
-import org.bukkit.SoundCategory;
-import org.bukkit.Statistic;
-import org.bukkit.WeatherType;
-import org.bukkit.World;
+import org.bukkit.*;
```
TheBusyBiscuit: Please avoid wildcard imports

build.gradle
```diff
@@ -40,17 +41,18 @@ task sourcesJar(type: Jar) {
...
         // Mark release as 8 if the compiler is Java 10 or newer
-        options.release.set(8)
+        options.release.set(16)
     } else {
         // Only use the legacy compiler options otherwise
-        sourceCompatibility = 1.8
-        targetCompatibility = 1.8
+        sourceCompatibility = 16
+        targetCompatibility = 16
     }
```
TheBusyBiscuit: Do we even need this logic here? Since an older (legacy) compiler won't even be able to compile this into Java16 bytecode, right?

Perhaps @md5sha256 knows better.

build.gradle
```diff
@@ -22,10 +22,11 @@ repositories {
 
 dependencies {
     api "org.spigotmc:spigot-api:${gradle.ext.apiVersionFull}"
-    implementation 'org.junit.jupiter:junit-jupiter:5.7.1'
-    implementation 'org.hamcrest:hamcrest-library:1.3'
-    implementation 'org.apache.commons:commons-io:1.3.2'
-    implementation 'org.jetbrains:annotations:19.0.0'
+    implementation 'org.junit.jupiter:junit-jupiter:5.7.2'
+    implementation 'org.hamcrest:hamcrest-library:2.2'
+    //noinspection GradlePackageUpdate
```
TheBusyBiscuit: Mind elaborating this comment?

settings.gradle
```diff
@@ -2,12 +2,12 @@
...
+gradle.ext.apiVersion = '1.17'
 
 /*
   The Spigot version we are building MockBukkit against.
 */
-gradle.ext.apiVersionFull = '1.16.5-R0.1-SNAPSHOT'
+gradle.ext.apiVersionFull = '1.17.1-R0.1-SNAPSHOT'
 
 /*
   MockBukkit uses semantic-versioning (https://semver.org/).
```
TheBusyBiscuit, line 13: Hmmm, should we bump the MockBukkit version here too?

TheBusyBiscuit (changes requested):
Thanks a lot, looks good to me for the most part

settings.gradle
original-codematrix: Ya, I've already told @seeseemelk that I'm working rn and this adventure here will be continued when I have more free time. (Thats also the reason why I haven't checked it in the checklist in the description)

build.gradle
original-codematrix: Ya that's why I've added a todo for me because it shouldn't be necessary anymore or mostly just somehow adjusted.

build.gradle
original-codematrix: I mean it can be deleted :D was just for ignoring the hint that there is a "newer" version (which isn't the newer version) for that dependency. In my opinion, it shouldn't be a problem but that should the maintainers say somewhere if that kinda comment shouldn't be added.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
original-codematrix: adjusted my config for the project ;)

build.gradle
TheBusyBiscuit: I would say "remove the comment" then, we do wanna update these from time to time, so better have our IDEs hinting at it then us completely forgetting about that

settings.gradle
TheBusyBiscuit: I am aware but I don't think that answers my question.
I am not 100% sure if we need to update the version since `MockBukkit-v1.17` will be a new artifact anyway.
Then again, things were updated, so... I would argue either works but we should definitely think about it.

seeseemelk: Indeed, as it's a new artifact we wouldn't have any conflicts if we keep the current version number. But perhaps it's a good idea to bump it just in case someone thinks `MockBukkit-v1.16:1.5.1` and `MockBukkit-v1.17:1.5.1` are for some reason the same?

build.gradle
md5sha256: We won't need source and target here; release replaces those two legacy options. And yes, as Biscuit mentions, we won't be able to compile Java 16 bytecode if we're using a compiler which wants source and target as opposed to release.

seeseemelk (commented):
Other than TheBusyBiscuit's comments, I think it looks pretty good! I'm gonna wait with approving it though until it is no longer a draft PR.

original-codematrix:
@TheBusyBiscuit @seeseemelk I've adjust the Github actions so the publish don't start for forks

settings.gradle
original-codematrix: So how we should move on now? :D

TheBusyBiscuit: Just bump the version I would say.
Other than that, the redundant function from above's comment needs to get sorted out. Then it seems good to go for me.
Any reason why you still kept it as a draft?

original-codematrix: I wanted to add more tests. But if we want I can resolve that and add a new branch where I add more tests features

build.gradle
original-codematrix: So I would remove the complete method and just add the `options.release.set(...)` thingy. Or do we want to remove this here completely? It would just be there as an indicator that we use Java 16+.

settings.gradle
TheBusyBiscuit: I can't see how any unit test would have to do with the update to 1.17 explicitly.
So I would 100% favour you making a seperate PR for that.
Small, compact PRs are also easier for us to review and therefore get approved and merged faster ;)

build.gradle
TheBusyBiscuit: I think what @md5sha256 tried to say is that we only need `options.release.set(16)` here and can remove the rest, if I understoof correctly.
I am relying on him here as I don't use gradle this often.

seeseemelk: I agree here with @md5sha256  and @TheBusyBiscuit , we should remove the if and simple have an `options.release.set(16)` here

original-codematrix: done

TheBusyBiscuit (approved):
Looks good to me

PR #246, opened 2021-08-16
Bump dependency versions in README.md

bermudalocket:
Change the two dependency examples to reference version 1.7.0 instead of 1.6.0 (which does not exist under MockBukkit-v1.17 as seen here: https://repo1.maven.org/maven2/com/github/seeseemelk/MockBukkit-v1.17/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit (approved):

Narimm:
This will pass when #249 is merged

PR #257, opened 2021-10-19
Fix listener deregistration

### LINKED ISSUE #256: EventHandlers accumulate over multiple Mocks.
[BUG] EventHandlers accumulate over multiple mocking.

Consider this test

```java
public class MockUnMockTest {

	@AfterEach
	void shutdown(){
		MockBukkit.unmock();
	}

	@Test
	void mockTest(){
		MockBukkit.mock();
		TestPlugin plugin = MockBukkit.load(TestPlugin.class);
		MockBukkit.getMock().getPluginManager().registerEvents(plugin,plugin);
		Assertions.assertEquals(3,BlockBreakEvent.getHandlerList().getRegisteredListeners().length);
	}

	@Test
	void mockTest2(){
		MockBukkit.mock();
		TestPlugin plugin = MockBukkit.load(TestPlugin.class);
		MockBukkit.getMock().getPluginManager().registerEvents(plugin,plugin);
		Assertions.assertEquals(3,BlockBreakEvent.getHandlerList().getRegisteredListeners().length); // FAIL
	}
}
```

This test will fail on mockTest2 with the assertion that the number of registered listerners is 6.    Due to the static nature of EventHandlers they are not removed in between mocks.  Over the course of a sequence if tests.  Eventually they will be triggering exponentially.

We need to deregister eventHandlers when a plugin unloads.

Narimm:
# Description
Fixes #256 
No versioning required as this does not change any exposed methods - it corrects a bug 

# Checklist
- [X] Version number updated in `settings.gradle`.
- [X] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
@@ -646,12 +646,11 @@ public class PluginManagerMock implements PluginManager
 	private void addListener(Listener listener, Plugin plugin)
 	{
 		List<Listener> l  = listeners.getOrDefault(plugin.getName(), new ArrayList<>());
-		if (l.contains(listener))
+		if (!l.contains(listener))
 		{
-			return;
+			l.add(listener);
+			listeners.put(plugin.getName(), l);
 		}
-		l.add(listener);
-		listeners.put(plugin.getName(), l);
 	}
 
 	public void unregisterPluginEvents(Plugin plugin)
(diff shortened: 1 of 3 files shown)


seeseemelk:
> No versioning required as this does not change any exposed methods - it corrects a bug
A bug fix does demand a bump of the patch version though

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -634,11 +635,39 @@ public void registerEvents(@NotNull Listener listener, @NotNull Plugin plugin)
...
+
+	private void addListener(Listener listener, Plugin plugin)
+	{
+		List<Listener> l  = listeners.getOrDefault(plugin.getName(), new ArrayList<>());
+		if (l.contains(listener))
+		{
+			return;
+		}
+		l.add(listener);
+		listeners.put(plugin.getName(), l);
```
seeseemelk:
```suggestion
		if (!l.contains(listener))
		{
			l.add(listener);
			listeners.put(plugin.getName(), l);
		}
```
This seems a bit shorter and more readable imo

seeseemelk (changes requested):

seeseemelk (approved):

PR #264, opened 2021-11-04
Switch API Implementation to PaperMC

Silthus:
# Description
All tests will fail if using PaperMC as the API. Most servers are run an paper and there are very good performance and API reasons to switch. Because of that I switched the leading API implementation of MockBukkit to Paper.

Let's discuss if this is useful and should be merged.

# Checklist
- [X] Unit tests added.


Seggan:
There is already a fork of MockBukkit for Paper: https://github.com/electronicboy/MockBukkitPaper. My opinion ¯\_(ツ)_/¯

rona-tombo:
> There is already a fork of MockBukkit for Paper: https://github.com/electronicboy/MockBukkitPaper. My opinion ¯_(ツ)_/¯

The repository seems to be unmaintained.
I think we should discuss again whether we should merge this PR or not?

seeseemelk:
Would it be possible to update this PR also for v1.18? At this point I would also like to get paper support as there is more and more demand for MockBukkit to support it

PR #269, opened 2021-11-22
Support World#spawn

sya-ri:
# Description

- Changed `mockEntity` to use `Class<T extends Entity>` instead of `EntityType`.
- Changed to call `spawn` from `spawnEntity` as in NMS.
  (The internal `spawn` call is a bit different from NMS to simplify the code. No problem.)

# Checklist
- [ ] Unit tests added. (Tests using `spawnEntity` are `spawn` tests.)

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -14,7 +14,7 @@ gradle.ext.apiVersionFull = '1.17.1-R0.1-SNAPSHOT'
   Every new update should bump the version below according to
   the conventions.
 */
-gradle.ext.version = '1.10.4'
+gradle.ext.version = '1.11.0'
 
 /*
   This is the name of our MockBukkit artifact, it includes
--- a/settings.gradle
+++ b/settings.gradle
@@ -14,7 +14,7 @@ gradle.ext.apiVersionFull = '1.17.1-R0.1-SNAPSHOT'
   Every new update should bump the version below according to
   the conventions.
 */
-gradle.ext.version = '1.12.2'
+gradle.ext.version = '1.13.0'
 
 /*
   This is the name of our MockBukkit artifact, it includes
--- a/settings.gradle
+++ b/settings.gradle
@@ -14,7 +14,7 @@ gradle.ext.apiVersionFull = '1.17.1-R0.1-SNAPSHOT'
   Every new update should bump the version below according to
   the conventions.
 */
-gradle.ext.version = '1.11.0'
+gradle.ext.version = '1.12.0'
 
 /*
   This is the name of our MockBukkit artifact, it includes
--- a/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
@@ -827,7 +827,7 @@ public class WorldMock implements World
 	public <T extends Entity> T spawn(@NotNull Location location, @NotNull Class<T> clazz) throws IllegalArgumentException
 	{
 		Validate.notNull(location, "The provided location must not be null.");
-		Validate.notNull(location, "The provided clazz must not be null.");
+		Validate.notNull(clazz, "The provided clazz must not be null.");
 
 		EntityMock entity = mockEntity(clazz);
 		entity.setLocation(location);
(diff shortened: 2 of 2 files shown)


src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -826,18 +824,27 @@ public void save()
 	}
 
 	@Override
-	public <T extends Entity> T spawn(Location location, Class<T> clazz) throws IllegalArgumentException
+	public <T extends Entity> T spawn(@NotNull Location location, @NotNull Class<T> clazz) throws IllegalArgumentException
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Validate.notNull(location, "The provided location must not be null.");
+		Validate.notNull(location, "The provided clazz must not be null.");
```
TheBusyBiscuit:
```suggestion
		Validate.notNull(clazz, "The provided class must not be null.");
```

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -826,18 +824,27 @@ public void save()
...
+		EntityMock entity = mockEntity(clazz);
+		entity.setLocation(location);
+		server.registerEntity(entity);
+
+		return (T) entity;
 	}
 
 	@Override
-	public <T extends Entity> T spawn(Location location, Class<T> clazz, Consumer<T> function)
+	public <T extends Entity> T spawn(@NotNull Location location, @NotNull Class<T> clazz, Consumer<T> function)
```
TheBusyBiscuit: The `Consumer<T>` should be marked as nullable here

settings.gradle
```diff
@@ -14,7 +14,7 @@ gradle.ext.apiVersionFull = '1.17.1-R0.1-SNAPSHOT'
   Every new update should bump the version below according to
   the conventions.
 */
-gradle.ext.version = '1.10.4'
+gradle.ext.version = '1.11.0'
```
TheBusyBiscuit:
```suggestion
gradle.ext.version = '1.12.0'
```

TheBusyBiscuit (changes requested):
Not the most elegant but it should be fine, I marked some things which will need to be changed though.

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
sya-ri: Thank you. After copying, I forgot to change it. I fixed in [`a82c24a` (#269)](https://github.com/MockBukkit/MockBukkit/pull/269/commits/a82c24ae663366a86473a152e0fde069c4e8e836), but I didn't fix the message because I think it is better the same as the argument name.

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
sya-ri: fix [`95beb27` (#269)](https://github.com/MockBukkit/MockBukkit/pull/269/commits/95beb27604d3578102eb7098565c7029b300f999)

TheBusyBiscuit (approved):
LGTM

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -826,18 +824,27 @@ public void save()
 	}
 
 	@Override
-	public <T extends Entity> T spawn(Location location, Class<T> clazz) throws IllegalArgumentException
+	public <T extends Entity> T spawn(@NotNull Location location, @NotNull Class<T> clazz) throws IllegalArgumentException
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Validate.notNull(location, "The provided location must not be null.");
+		Validate.notNull(clazz, "The provided clazz must not be null.");
```
TheBusyBiscuit:
```suggestion
		Validate.notNull(clazz, "The provided class must not be null.");
```

settings.gradle
```diff
@@ -14,7 +14,7 @@ gradle.ext.apiVersionFull = '1.17.1-R0.1-SNAPSHOT'
   Every new update should bump the version below according to
   the conventions.
 */
-gradle.ext.version = '1.12.1'
+gradle.ext.version = '1.12.2'
```
TheBusyBiscuit:
```suggestion
gradle.ext.version = '1.13.0'
```

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):

PR #278, opened 2021-12-06
fix(player): getOfflinePlayer returns random player with different id

Silthus:
# Description
Querying for an `OfflinePlayer` with `ServerMock.getOfflinePlayer(UUID)` returns a random offline player that has a different id. This can lead to an endless loop.

# Checklist
- [X] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
@@ -62,7 +62,7 @@ public final class PlayerMockFactory
 	 * @param uuid the id of the offline player
 	 * @return A newly created offline player mock object.
 	 */
-	public @NotNull OfflinePlayerMock createOfflinePlayer(UUID uuid)
+	public @NotNull OfflinePlayerMock createOfflinePlayer(@NotNull UUID uuid)
 	{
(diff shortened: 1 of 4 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMockFactory.java
```diff
@@ -56,4 +56,14 @@ public PlayerMockFactory(@NotNull ServerMock server)
 	{
 		return new OfflinePlayerMock(getUniqueRandomName());
 	}
+
+	/**
+	 * Create a random {@link OfflinePlayerMock} object with a unique name and the given id.
+	 * @param uuid the id of the offline player
+	 * @return A newly created offline player mock object.
+	 */
+	public @NotNull OfflinePlayerMock createOfflinePlayer(UUID uuid)
```
TheBusyBiscuit:
```suggestion
	public @NotNull OfflinePlayerMock createOfflinePlayer(@NotNull UUID uuid)
```

TheBusyBiscuit (changes requested):
LGTM, small change though.
And the version number needs to be updated.

Silthus:
Ok thank you. I bumped the version and added the `@NotNull`.

TheBusyBiscuit (approved):
LGTM

PR #280, opened 2021-12-13
Implement create world

PeachesMLG:
# Description

# Checklist
- [ ] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -109,8 +109,7 @@ import be.seeseemelk.mockbukkit.tags.TagsMock;
 import net.md_5.bungee.api.chat.BaseComponent;
 import org.jetbrains.annotations.Nullable;
 
-public class ServerMock extends Server.Spigot implements Server
-{
+public class ServerMock extends Server.Spigot implements Server {
 	private static final String BUKKIT_VERSION = "1.17.1";
 	private static final String JOIN_MESSAGE = "%s has joined the server.";
 	private static final String MOTD = "A Minecraft Server";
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -994,8 +994,7 @@ public ServicesManagerMock getServicesManager()
 	@Override
 	public World createWorld(WorldCreator creator)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return addSimpleWorld(creator.name());
```
TheBusyBiscuit: The point of a `WorldCreator` is to create a World that fits certain requirements.
These requirements include:
* Name
* Seed
* Environment
* ChunkGenerator
* BiomeProvider
* WorldType
* GenerateStructures (boolean)
* GeneratorSettings
* HardcoreMode (boolean)

Your implementation only fulfills one of these requirements, if I want to create a Nether world and I get an overworld, clearly something's wrong.

If you are interested in a proper implementation of this, I suggest you to take a test-driven approach in this case.
Write a Test class for the `WorldCreator` and write test cases for each of these requirements.
Set one of those values, create a world and check if the requirement was fulfilled.
Then implement accordingly.

Of course we don't need actually need chunk generation or biomes and stuff here.
But environment, seed, worldtype and such should be the minimum here.
The rest can always be implemented later (if anyone cares for those features)

TheBusyBiscuit (changes requested):

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
PeachesMLG: I can add mocked Environment, but both getSeed and getWorldType return unimplemented exceptions, so does anyone even care about that?

PeachesMLG:
uh... Think I had some issues with versions or something, seems some of the code rolled back, ima make another pr

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
TheBusyBiscuit: In terms of simple get operations, yes. Those should be covered.

PR #287, opened 2021-12-18
Implemented CreateWorld

PeachesMLG:
# Description

# Checklist
- [ ] Unit tests added.


src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -907,6 +907,29 @@ public boolean getAllowNether()
...
+	@NotNull
+	@Override
+	public String getResourcePackPrompt() {
+		return null;
+	}
+
+	@Override
+	public boolean isResourcePackRequired() {
+		return false;
+	}
```
TheBusyBiscuit: Invalid implementations, please follow our specifications and throw an `UnimplementedOperationException` for non-implemented method stubs.

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -14,26 +14,7 @@
...
-import org.bukkit.Particle;
-import org.bukkit.Raid;
-import org.bukkit.Sound;
-import org.bukkit.SoundCategory;
-import org.bukkit.StructureType;
-import org.bukkit.TreeType;
-import org.bukkit.World;
-import org.bukkit.WorldBorder;
-import org.bukkit.WorldType;
+import org.bukkit.*;
```
TheBusyBiscuit: Please avoid wildcard imports

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1613,6 +1613,21 @@ public void setResourcePack(@NotNull String url, byte[] hash)
...
+
+	@Override
+	public void setResourcePack(@NotNull String url, @Nullable byte[] hash, boolean force) {
+
+	}
+
+	@Override
+	public void setResourcePack(@NotNull String url, @Nullable byte[] hash, @Nullable String prompt, boolean force) {
+
+	}
```
TheBusyBiscuit: Invalid implementations, please follow our specifications and throw an `UnimplementedOperationException` for non-implemented method stubs.

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -21,8 +21,7 @@
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicReference;
 
-import org.bukkit.Material;
-import org.bukkit.OfflinePlayer;
+import org.bukkit.*;
```
TheBusyBiscuit: Please avoid wildcard imports

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -67,6 +66,21 @@ void class_NumberOfPlayers_Zero()
...
+				.seed(12345)
+				.type(WorldType.FLAT)
+				.environment(World.Environment.NORMAL);
+		server.createWorld(worldCreator);
+
+		assertEquals(1, server.getWorlds().size());
+		assertEquals("test", server.getWorlds().get(0).getName());
+		assertEquals(12345, server.getWorlds().get(0).getSeed());
+		assertEquals(WorldType.FLAT, server.getWorlds().get(0).getWorldType());
+		assertEquals(World.Environment.NORMAL, server.getWorlds().get(0).getEnvironment());
```
TheBusyBiscuit: `Server#createWorld` returns the created World.
It would be much more sensible to store it in a var and use that for your assertions instead of `server.getWorlds().get(0)`

TheBusyBiscuit (changes requested):

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
PeachesMLG: Thats true, good point

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
PeachesMLG: Not quite sure why you dont like wildcard imports, they look way nicer, but whatever. Took me ages to find out how to make my ide disable them. Your code formatting conventions are annoying af

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -141,6 +144,14 @@ public WorldMock(Material defaultBlock, int height, int grassHeight)
 		gameRules.put(GameRule.SPECTATORS_GENERATE_CHUNKS, true);
 	}
 
+	public WorldMock(WorldCreator creator){
```
TheBusyBiscuit:
```suggestion
	public WorldMock(@NotNull WorldCreator creator)
	{
```

TheBusyBiscuit (changes requested):
One small change left (which you can just commit right from my review), then it is good to go!

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
TheBusyBiscuit: > they look way nicer, but whatever

Wildcard imports are not a style choice, they are a functional utility.
Their purpose is to import entire namespaces which can reduce the file size when you got many imports from the same namespace. However, the 90s are over, so we don't need to worry about every single bit or byte in our file anymore.

Nowadays, wildcard imports are sometimes regarded as a bad practice as they lead to so-called namespace pollution.
By nature, wildcard imports will automatically import any classes from the given namespace into the current class.
Overuse of this can lead to conflicts and collisions ([Wikipedia](https://en.wikipedia.org/wiki/Name_collision)), as the whole point of a namespace is to seperate class names which are non-unique.
By using a wildcard import, you are opening the door to ambiguities and potential collisions.

Check this for more info and a pretty good example:
https://rules.sonarsource.com/java/RSPEC-2208

And as you can guess, the `org.bukkit` namespace is by no means a small one...

PeachesMLG: 99% sure it gets optimized on compilation, but whatever

TheBusyBiscuit: This is not a runtime issue, it's a bad development practice and not happening on the compiler level.
But either way, I gave you a decent explanation, what you do with this information and advice is all your call.

TheBusyBiscuit (approved):
LGTM

Issue #289, opened 2021-12-22
Issues with PlayerMock#simulateBlockPlace


PeachesMLG:
PlayerMock#simulateBlockPlace sets BlockPlaceEvent#getBlock to the block before the block was placed (air), however on spigot, this is set to the block that will be placed.

Insprill:
Fixed with #290.

PR #290, opened 2021-12-22
Fixed Issues with Simulate BlockPlaceEvent

PeachesMLG:
# Description
Fixes Issue: https://github.com/MockBukkit/MockBukkit/issues/289

# Checklist
- [ ] Unit tests added.


src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -243,11 +244,13 @@ protected BlockDamageEvent simulateBlockDamagePure(Block block)
...
 		Block block = location.getBlock();
+		BlockState blockState = block.getState();
+		block.setType(material);
 		BlockPlaceEvent event = new BlockPlaceEvent(block, null, null, null, this, true, null);
 		Bukkit.getPluginManager().callEvent(event);
-		if (!event.isCancelled())
+		if (event.isCancelled())
 		{
-			block.setType(material);
+			blockState.update(true);
```
seeseemelk: I don't think this will actually reset the material of the block. However, that might be a bug in MockBukkit:
[here we set the block state of a block](https://github.com/PeachesMLG/MockBukkit/blob/95aeaa4b7e32b156b789bcedf24eed9adb064fbe/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java#L439), and [here we get the material of a block](https://github.com/PeachesMLG/MockBukkit/blob/95aeaa4b7e32b156b789bcedf24eed9adb064fbe/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java#L156) (both methods are declared in the same class).

Insprill: Indeed an issue with MockBukkit fixed with #403.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -243,11 +244,13 @@ protected BlockDamageEvent simulateBlockDamagePure(Block block)
 		if (gamemode == GameMode.ADVENTURE || gamemode == GameMode.SPECTATOR)
 			return null;
 		Block block = location.getBlock();
+		BlockState blockState = block.getState();
+		block.setType(material);
 		BlockPlaceEvent event = new BlockPlaceEvent(block, null, null, null, this, true, null);
```
Insprill:
```suggestion
		BlockPlaceEvent event = new BlockPlaceEvent(block, blockState, null, null, this, true, null);
```

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
Insprill:
```suggestion
			blockState.update(true, false);
```

Insprill (changes requested):

Insprill (approved):
LGTM, just need to wait for #403 to be merged.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -286,11 +287,13 @@ protected BlockDamageEvent simulateBlockDamagePure(Block block)
 		if (gamemode == GameMode.ADVENTURE || gamemode == GameMode.SPECTATOR)
 			return null;
 		Block block = location.getBlock();
-		BlockPlaceEvent event = new BlockPlaceEvent(block, null, null, null, this, true, null);
+		BlockState blockState = block.getState();
+		block.setType(material);
+		BlockPlaceEvent event = new BlockPlaceEvent(block, blockState, null, null, this, true, null);
 		Bukkit.getPluginManager().callEvent(event);
-		if (!event.isCancelled())
+		if (event.isCancelled())
```
Insprill:
```suggestion
		if (event.isCancelled() || !event.canBuild())
```

Insprill (changes requested):

Insprill (approved):

PR #466, opened 2022-06-16
Fixes ChunkMock#getChunkSnapshot passing in oversized maps to ChunkSnapshotMock

Insprill:
# Description
The size was being multiplied by an extra 256 😅

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ChunkMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ChunkMock.java
@@ -97,8 +97,8 @@ public class ChunkMock implements Chunk
 	@SuppressWarnings("UnstableApiUsage")
 	public @NotNull ChunkSnapshot getChunkSnapshot(boolean includeMaxblocky, boolean includeBiome, boolean includeBiomeTempRain)
 	{
-		// Cubic size of the chunk.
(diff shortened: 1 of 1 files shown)


src/main/java/be/seeseemelk/mockbukkit/ChunkMock.java
```diff
@@ -97,8 +97,8 @@ public long getChunkKey()
 	@SuppressWarnings("UnstableApiUsage")
 	public @NotNull ChunkSnapshot getChunkSnapshot(boolean includeMaxblocky, boolean includeBiome, boolean includeBiomeTempRain)
 	{
-		// Cubic size of the chunk.
-		int size = (16 << 4) * (16 << 4) * Math.abs((world.getMaxHeight() - world.getMinHeight()));
+		// Cubic size of the chunk. (16 * 16 * height, using a bitshift)
+		int size = (16 << 4) * Math.abs((world.getMaxHeight() - world.getMinHeight()));
```
TheBusyBiscuit: In all honesty why are we not writing out `16 * 16` or just straight-up `256`...?

Insprill: For the past while I've just been in the habit of using bitshifts for multiples of 2. I'll make it 16 * 16 since that's the clearest.

thelooter (approved):

PR #470, opened 2022-06-16
Target specific paper-api version

Insprill:
# Description
Targets a specific version of the Paper API. This will prevent pull requests' builds from failing when they add new methods we haven't stubbed yet.

I'm leaving this as a draft for now because ideally, we should have an action that runs daily checking for new builds, and if there is one, creating an issue alerting us to the update, or creating a draft pull request bumping the version and waiting on us to manually stub the new methods.


Insprill:
If anyone would like to give creating the aforementioned script/action a try that would be greatly appreciated.

PR #471, opened 2022-06-16
Prevent having objectives or teams with the same name

syldium:
# Description
Two teams or two objectives cannot be registered in the same scoreboard with the same name, since the name is a unique identifier. I added a check against this.

A team has the same life cycle as a objective in a scoreboard:
1. Creation and registration for the scoreboard
2. De-registration, the team should not be referenced or used anymore

A team could be retrieved by its name even if it had been unregistered. I added the removal of the team in the scoreboard when it is unregistered to fix this.

# Checklist
- [x] Unit tests added.

# For maintainer


thelooter (approved):
LGTM

PR #474, opened 2022-06-16
Fix smelly code

Insprill:
# Description
Fixes a bunch of SonarCloud code smells that are too small for their own PRs.

# Checklist
- [ ] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
@@ -894,9 +894,9 @@ public class WorldMock implements World
 			canceled = new CreatureSpawnEvent(living, reason).callEvent();
 		} else if (entity instanceof Item item) {
 			canceled = new ItemSpawnEvent(item).callEvent();
-		} else if (entity instanceof Player player) {
+		} else if (entity instanceof Player) {
 			canceled = true; // Shouldn't ever be called here but just for parody.
-		} else if (entity instanceof Projectile projectile) {
+		} else if (entity instanceof Projectile) {
 			canceled = new ProjectileLaunchEvent(entity).callEvent();
 		} else {
 			canceled = new EntitySpawnEvent(entity).callEvent();
@@ -1103,7 +1103,7 @@ public class WorldMock implements World
 		}
 		ThunderChangeEvent thunder = new ThunderChangeEvent(this, thundering, ThunderChangeEvent.Cause.PLUGIN); // Paper
 		Bukkit.getServer().getPluginManager().callEvent(thunder);
-		if (thunder.isCancelled()) 
+		if (thunder.isCancelled())
 		{
 			return;
 		}
--- a/src/main/java/be/seeseemelk/mockbukkit/profile/PlayerProfileMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/profile/PlayerProfileMock.java
@@ -142,13 +142,6 @@ public class PlayerProfileMock implements PlayerProfile
 	}
 
 	@Override
-	@SuppressWarnings("MethodDoesntCallSuperMethod")
-	public org.bukkit.profile.@NotNull PlayerProfile clone()
-	{
-		return new PlayerProfileMock(this);
-	}
-
-	@Override
 	public boolean completeFromCache()
 	{
 		// TODO Auto-generated method stub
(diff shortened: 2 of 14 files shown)


src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -894,9 +894,9 @@ private void callSpawnEvent(EntityMock entity, CreatureSpawnEvent.SpawnReason re
...
 		} else if (entity instanceof Item item) {
 			canceled = new ItemSpawnEvent(item).callEvent();
-		} else if (entity instanceof Player player) {
+		} else if (entity instanceof Player) {
 			canceled = true; // Shouldn't ever be called here but just for parody.
-		} else if (entity instanceof Projectile projectile) {
+		} else if (entity instanceof Projectile) {
 			canceled = new ProjectileLaunchEvent(entity).callEvent();
 		} else {
 			canceled = new EntitySpawnEvent(entity).callEvent();
```
thelooter: Brace Style

src/main/java/be/seeseemelk/mockbukkit/profile/PlayerProfileMock.java
```diff
@@ -191,25 +184,24 @@ public boolean complete(boolean textures, boolean onlineMode)
...
+	public int hashCode()
 	{
-		if (this == obj) return true;
-		if (!(obj instanceof PlayerProfileMock otherProfile)) return false;
-		return Objects.equals(this.name, otherProfile.name) && this.uuid.equals(otherProfile.uuid) && this.properties.equals(otherProfile.properties);
+		return this.properties.hashCode();
 	}
 
 	@Override
-	public String toString()
```
thelooter: Why was the toString Method removed?

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -226,23 +225,22 @@ public void waitAsyncTasksFinished()
...
 				return;
 			}
-			if (System.currentTimeMillis() > (systemTime + executorTimeout))
+
+			if (System.currentTimeMillis() <= (systemTime + executorTimeout))
+				continue;
+
+			// If a plugin has left a runnable going and not cancelled it we could call this bad practice.
+			// We should force interrupt all these runnables, forcing them to throw Interrupted Exceptions
+			// if they handle that.
```
thelooter, line 234: I think it's a good Idea to do this, but we should give Users a way to disable this behaviour on a Test to Test Basis, since there might be cases where a long running Runnable might be needed.

src/main/java/be/seeseemelk/mockbukkit/profile/PlayerProfileMock.java
Insprill: It's just unnecessary

src/main/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMock.java
Insprill: Agreed, but i think that's out of this PR's scope.

thelooter: Then this Comment and Interruption is out of scope for this PR too. I think this should be handled in it's own PR

Insprill: It's not part of this PR, it has just been moved to the left a bit.

thelooter: I see

src/main/java/be/seeseemelk/mockbukkit/profile/PlayerProfileMock.java
thelooter: Why though?

Insprill: Just no real reason to use it. Will add it back though.

thelooter: Good, we can't guarantee that people haven't used it so we wouldn't want to break compatibility

thelooter (approved):

PR #495, opened 2022-06-21
Implement a better system for type checking BlockStateMocks

Insprill:
# Description
Creates a helper method in BlockStateMock to easily check if the type is correct.

# Checklist
- [x] Unit tests added.


TheBusyBiscuit (changes requested):
I would suggest adding two overloads, one for `Material` and one for `Tag<Material>`.

It would make it a lot cleaner and reuseable.

src/main/java/be/seeseemelk/mockbukkit/block/state/CampfireMock.java
```diff
@@ -22,15 +22,13 @@ public class CampfireMock extends TileStateMock implements Campfire
...
-			throw new IllegalArgumentException("Cannot create a Campfire state from " + material);
+		checkType(material, Material.CAMPFIRE, Material.SOUL_CAMPFIRE);
 	}
 
 	protected CampfireMock(@NotNull Block block)
 	{
 		super(block);
-		if (block.getType() != Material.CAMPFIRE && block.getType() != Material.SOUL_CAMPFIRE)
-			throw new IllegalArgumentException("Cannot create a Campfire state from " + block.getType());
+		checkType(block, Material.CAMPFIRE, Material.SOUL_CAMPFIRE);
```
TheBusyBiscuit: I am 99% certain that there is a campfires tag

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
```diff
@@ -50,8 +52,30 @@ protected BlockStateMock(@NotNull BlockStateMock state)
...
+
+	protected void checkType(Material material, MaterialSetTag tag)
+	{
+		Preconditions.checkArgument(tag.isTagged(material), "Cannot create a " + getClass().getSimpleName() + " from " + material);
+	}
+
+	protected void checkType(Block block, MaterialSetTag expected)
+	{
+		checkType(block.getType(), expected);
+	}
```
TheBusyBiscuit: Why not use the base class `Tag<Material>`...?

src/main/java/be/seeseemelk/mockbukkit/block/state/BannerMock.java
```diff
@@ -25,15 +25,13 @@ public class BannerMock extends TileStateMock implements Banner
...
-			throw new IllegalArgumentException("Cannot create a Banner state from " + material.name());
+		checkType(material, Tag.BANNERS.getValues().toArray(new Material[0])); // Paper doesn't have the banners tag in the MaterialTags class (yet)
 	}
 
 	protected BannerMock(@NotNull Block block)
 	{
 		super(block);
-		if (!Tag.BANNERS.isTagged(block.getType()))
-			throw new IllegalArgumentException("Cannot create a Banner state from " + block.getType().name());
+		checkType(block, Tag.BANNERS.getValues().toArray(new Material[0]));
```
TheBusyBiscuit: Converting this into an array should not be necessary.

TheBusyBiscuit (changes requested):
Good progress you made here, thanks!

However I am confused as to why you use Paper's `MaterialSetTag` class instead of just using `Tag<Material>` as suggested, using the base class would still allow you to pass values like `MaterialTags.XYZ` from Paper's convenience class but also allow the passing of `Tag.BANNERS` without the need for any array conversions...

In general, you should always declare methods using the highest-possible class from the hierarchy or even the base class, i.e. `List<T>` instead of `ArrayList<T>` or in this case `Tag<Material>` instead of `MaterialSetTag`. This allows for flexibility and reusability.

src/main/java/be/seeseemelk/mockbukkit/block/state/BannerMock.java
```diff
@@ -25,15 +25,13 @@ public class BannerMock extends TileStateMock implements Banner
 	public BannerMock(@NotNull Material material)
 	{
 		super(material);
-		if (!Tag.BANNERS.isTagged(material))
-			throw new IllegalArgumentException("Cannot create a Banner state from " + material.name());
+		checkType(material, Tag.BANNERS); // Paper doesn't have the banners tag in the MaterialTags class (yet)
```
TheBusyBiscuit: This comment can be removed
```suggestion
		checkType(material, Tag.BANNERS);
```

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
```diff
@@ -50,8 +52,30 @@ protected BlockStateMock(@NotNull BlockStateMock state)
...
+
+	protected void checkType(Material material, Tag<Material> tag)
+	{
+		Preconditions.checkArgument(tag.isTagged(material), "Cannot create a " + getClass().getSimpleName() + " from " + material);
+	}
+
+	protected void checkType(Block block, Tag<Material> expected)
+	{
+		checkType(block.getType(), expected);
+	}
```
TheBusyBiscuit: Also, perhaps add `@NotNull` annotations here?

Insprill:
I do always try to use the highest class in the hierarchy, just didn't do my due diligence on this one :P

TheBusyBiscuit (changes requested):
Two more things then it will look good to go

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
Insprill: Yup, they'd be good. Gonna leave out the checks for them though since they should be caught in the constructor.

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
```diff
@@ -274,14 +298,10 @@ public boolean equals(Object obj)
 		{
 			return false;
 		}
-		if (this.isPlaced() && this.getLocation() != other.getLocation() && (this.getLocation() == null || !this.getLocation().equals(other.getLocation())))
-		{
-			return false;
-		}
+		return !this.isPlaced() || this.getLocation() == other.getLocation() || (this.getLocation() != null && this.getLocation().equals(other.getLocation()));
```
TheBusyBiscuit: Just a note in case anyone wants to bother checking this:
I was too lazy to check the equivalency of this statement 🤷🏻

TheBusyBiscuit (approved):
LGTM 👍🏻

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
Insprill: [Here's the CraftBukkit implementation](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/src/main/java/org/bukkit/craftbukkit/block/CraftBlockState.java#264-283). The only real difference is we skip the check if it isn't placed to avoid an exception being thrown when getting the location, which doesn't happen in CraftBukkit. That and we haven't implemented the BlockData part since `#getBlockData` isn't implemented.

TheBusyBiscuit: I know, I was just commenting on the fact that I didn't check if the statement is equivalent to the one before since you basically inverted it (while keeping the return value still the same). It was still an approving review, just a comment in case someone comes back here

Insprill: So can this be resolved? It's blocking the merge 😅

TheBusyBiscuit: of course

PR #498, opened 2022-06-23
Implement the PluginManagerMock default permission methods

Insprill:
# Description
Implements the PluginManagerMock default permission methods.

# Checklist
- [x] Unit tests added.

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
@@ -908,7 +908,9 @@ public class PluginManagerMock implements PluginManager
 		Map<Permissible, Boolean> map = this.defaultPermissionSubscriptions.get(op);
 
 		if (map == null)
+		{
 			return;
+		}
 
 		map.remove(permissible);
 		if (map.isEmpty())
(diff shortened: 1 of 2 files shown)


src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -895,28 +898,37 @@ public void unsubscribeFromPermission(@NotNull String permission, @NotNull Permi
...
 
 	@Override
 	public void unsubscribeFromDefaultPerms(boolean op, @NotNull Permissible permissible)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Map<Permissible, Boolean> map = this.defaultPermissionSubscriptions.get(op);
+
+		if (map == null)
+			return;
```
thelooter:
```suggestion
		if (map == null)
		{
			return;
		}
```

thelooter (changes requested):

src/main/java/be/seeseemelk/mockbukkit/plugin/PluginManagerMock.java
Insprill: Are braces really needed here? I think it makes it a lot more cluttered since it's a fairly small method.

thelooter: I think we should either have braces or inline that return-

Insprill: Hmm... I'd rather inline it but I'm not a huge ran of that either since it's a bit harder at a glance to see that it returns.

thelooter: I think we should ask @seeseemelk what he thinks

PeachesMLG: Should be inline :eyes:

Insprill: Ok, this is stupid. Braces it is.

thelooter (approved):

PR #501, opened 2022-06-25
Implement Brewer Inventory

thelooter:
# Description
Implement BrewerInventory in the quest to finally finish #113 

# Checklist
The following items should be checked before the pull request can be merged.
- [x] Unit tests added.
- [x] Code follows existing code format.

# Info on creating a pull request
- Make sure that unit tests are added which test the relevant changes.
- Make sure that the changes follow the existing code format.

# For maintainer
When a PR is approved, the maintainer should label this PR with `release/*`.

- If the PR fixes a bug in MockBukkit, update the patch version. (if the current version is `0.4.1`, the new version should be `0.4.2`)
- If the PR adds a new feature to MockBukkit, update minor version. (if the current version is `0.4.1`, the new version should be `0.5.0`)

Note that a PR that fixes an `UnimplementedOperationException` should be considered a new version and not a bugfix.

PR #503, opened 2022-06-25
Horse mock

jusdino:
# Description
Add entity.HorseMock and related classes, tests

# Checklist
- [*] Unit tests added.
- [*] Code follows existing code format.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/HorseMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/HorseMock.java
@@ -18,6 +18,9 @@ public class HorseMock extends AbstractHorseMock implements Horse {
     public HorseMock(ServerMock server, UUID uuid) {
         super(server, uuid);
         inventory = new HorseInventoryMock(this);
+        color = Color.WHITE;
+        style = Style.BLACK_DOTS;
+        inventory = new HorseInventoryMock(this);
     }
 
     @Override
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/HorseMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/HorseMock.java
@@ -13,59 +13,62 @@ import be.seeseemelk.mockbukkit.UnimplementedOperationException;
 import be.seeseemelk.mockbukkit.inventory.HorseInventoryMock;
 
 public class HorseMock extends AbstractHorseMock implements Horse {
-    private Color color;
-    private Style style;
-    private HorseInventory inventory;
+	private Color color;
+	private Style style;
+	private HorseInventory inventory;
 
-    public HorseMock(ServerMock server, UUID uuid) {
-        super(server, uuid);
-        inventory = new HorseInventoryMock(this);
-        color = Color.WHITE;
-        style = Style.BLACK_DOTS;
-        inventory = new HorseInventoryMock(this);
-    }
+	public HorseMock(ServerMock server, UUID uuid) {
+		super(server, uuid);
+		inventory = new HorseInventoryMock(this);
+		color = Color.WHITE;
+		style = Style.BLACK_DOTS;
+		inventory = new HorseInventoryMock(this);
+	}
 
-    @Override
-    public Color getColor() {
-        return this.color;
-    }
+	@Override
+	public Color getColor() {
+		Preconditions.checkState(this.color != null, "No color has been set");
+		return this.color;
+	}
 
-    @Override
-    public void setColor(@NotNull Color color) {
-        Preconditions.checkNotNull(color,"Color cannot be null");
-        this.color = color;
-    }
+	@Override
+	public void setColor(@NotNull Color color) {
+		Preconditions.checkNotNull(color, "Color cannot be null");
+		this.color = color;
+	}
 
-    @Override
-    public Style getStyle() {
-        return this.style;
-    }
+	@Override
+	public Style getStyle() {
+		Preconditions.checkState(this.style != null, "No style has been set");
+		return this.style;
+	}
 
-    @Override
-    public void setStyle(@NotNull Style style) {
-        Preconditions.checkNotNull(style,"Style cannot be null");
-        this.style = style;
-    }
+	@Override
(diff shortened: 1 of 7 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/HorseMock.java
```diff
@@ -0,0 +1,67 @@
...
+        super(server, uuid);
+        inventory = new HorseInventoryMock(this);
+        color = Color.WHITE;
+        style = Style.BLACK_DOTS;
+        inventory = new HorseInventoryMock(this);
+    }
+
+    @Override
+    public Color getColor() {
+        return color;
```
thelooter:
```suggestion
        return this.color;
```

src/main/java/be/seeseemelk/mockbukkit/entity/HorseMock.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public void setColor(Color color) {
+        this.color = color;
+    }
```
thelooter:
```suggestion
    public void setColor(@NotNull Color color) {
        Preconditions.checkNotNull(color,"Color cannot be null");
        this.color = color;
    }
```

src/main/java/be/seeseemelk/mockbukkit/entity/HorseMock.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public Style getStyle() {
+        return style;
```
thelooter:
```suggestion
        return this.style;
```

src/main/java/be/seeseemelk/mockbukkit/entity/HorseMock.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public void setStyle(Style style) {
+        this.style = style;
+    }
```
thelooter:
```suggestion
    public void setStyle(@NotNull Style style) {
        Preconditions.checkNotNull(style,"Style cannot be null");
        this.style = style;
    }
```

src/main/java/be/seeseemelk/mockbukkit/entity/HorseMock.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public void setCarryingChest(boolean chest) {
+        throw new UnimplementedOperationException();
+    }
+
+    @Override
+    public HorseInventory getInventory() {
+        return inventory;
```
thelooter:
```suggestion
        return this.inventory;
```

src/main/java/be/seeseemelk/mockbukkit/inventory/HorseInventoryMock.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public ItemStack getSaddle() {
+        return this.getItem(0);
+    }
+
+    @Override
+    public void setSaddle(ItemStack stack) {
+        this.setItem(0, stack);
+    }
```
thelooter:
```suggestion
    public void setSaddle(@Nullable ItemStack stack) {
        this.setItem(0, stack);
    }
```

src/main/java/be/seeseemelk/mockbukkit/inventory/HorseInventoryMock.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public ItemStack getArmor() {
+        return this.getItem(1);
+    }
+
+    @Override
+    public void setArmor(ItemStack stack) {
+        this.setItem(1, stack);
+    }
```
thelooter:
```suggestion
    public void setArmor(@Nullable ItemStack stack) {
        this.setItem(1, stack);
    }
```

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
```diff
@@ -263,6 +264,19 @@ void spawnZombieTest()
...
+	{
+		WorldMock world = new WorldMock();
+		Location location = new Location(world, 100, 20, 50);
+		Entity horse = world.spawnEntity(location, EntityType.HORSE);
+		assertEquals(100, horse.getLocation().getBlockX());
+		assertEquals(20, horse.getLocation().getBlockY());
+		assertEquals(50, horse.getLocation().getBlockZ());
+		assertTrue(world.getEntities().size() > 0);
+		assertInstanceOf(Horse.class, horse);
+	}
```
thelooter: I don't think the Position is necessary here

thelooter (changes requested):

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
jusdino: I don't follow. You mean don't use `Location location`? Or do you mean I don't need to check the `horse.getLocation()` attributes because that is tested elsewhere?

thelooter: If you look at the other Entity Spawn Tests you see that we don't check the Position there. Thats because the Method you are using is implemented in EntityMock and also tested there. And since we don't need special Handeling of the Method in the HorseMock class, we don't need to test it here. Only write a Test for something if you implement / override it in this class.

thelooter:
Also, something I forgot in the first round of Changes is a State Check on the Getter Methods. Look at #501 for a way to do that

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
jusdino: Got it. I removed the `.getLocation()` assertions. I assume I should still leave the rest of the test here since I did add logic to the `WorldMock` to spawn the `HorseMock`?

thelooter: Yes thats correct.

thelooter (approved):

PR #514, opened 2022-06-28
Add RickRoleMock

thelooter:
# Description
Add a RickRollMock

# Checklist
- [x] Unit tests added.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/RickRollMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/RickRollMock.java
@@ -1,25 +1,40 @@
 package be.seeseemelk.mockbukkit;
 
-import be.seeseemelk.mockbukkit.entity.EntityMock;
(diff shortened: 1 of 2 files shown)


Insprill (approved):
LGTM

seeseemelk:
I'm personally of the opinion that changes to MockBukkit should aim solely to improve either the experiences of the users that want to test their plugins and/or the maintenance burden for the maintainers.

This PR contributes very little to either. In fact, it provides a net negative - one that it increases our maintenance burden while projecting an image to our downstream users that MockBukkit is more about memes and fun rather than a serious tool used for testing plugins.

I saw this PR a few days ago, and I should have pressed `request changes` instead of simply not approving it until I could figure out if this is something I think would add value.

I've opened a reversal PR which removes these classes as these classes provide nothing extra to our users whilst increasing the size of our codebase needlessly.

PR #519, opened 2022-07-02
Fire teleportation events when teleporting an entity

syldium:
# Description
The `Entity#teleport` method now fires the `EntityTeleportEvent` if it's not a player. If it's a player who goes to another dimension, this method will also fire the `PlayerChangedWorldEvent` afterward.

Some preconditions have been added, namely for non-nullable arguments. Some checks prevent weird states, like if the entity is removed or dead.

The `Entity#teleport` method now also makes a defensive copy of the location to prevent external mutations. The entity is the only owner of this location.

# Checklist
- [x] Unit tests added.
- [x] Code follows existing (local) code format.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
@@ -26,6 +26,7 @@ import org.bukkit.entity.Pose;
 import org.bukkit.entity.SpawnCategory;
 import org.bukkit.event.entity.CreatureSpawnEvent;
 import org.bukkit.event.entity.EntityDamageEvent;
+import org.bukkit.event.entity.EntityTeleportEvent;
 import org.bukkit.event.player.PlayerTeleportEvent.TeleportCause;
 import org.bukkit.metadata.MetadataValue;
 import org.bukkit.permissions.Permission;
@@ -259,7 +260,6 @@ public abstract class EntityMock extends Entity.Spigot implements Entity, Messag
 	@Override
 	public boolean teleport(@NotNull Location location)
 	{
-		Preconditions.checkNotNull(location, "Location cannot be null");
 		return teleport(location, TeleportCause.PLUGIN);
 	}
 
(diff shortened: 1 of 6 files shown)


src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1363,6 +1368,27 @@ void testPlayerTeleport_WithoutCause_EventFired()
...
 		server.getPluginManager().assertEventFired(PlayerTeleportEvent.class);
+		server.getPluginManager().assertEventNotFired(EntityTeleportEvent.class);
+	}
+
+	@Test
+	void testPlayerTeleport_ChangedWorldEvent()
+	{
+		TestPlugin plugin = MockBukkit.load(TestPlugin.class);
+		World from = player.getWorld();
+		World[] playerWorldWhenEvent = new World[1];
```
Insprill: Is there a particular reason this is an array?

syldium: The variable needs to be effectively final to be accessed from within the inner class. The array is, or it could be any holder object like an `AtomicReference` but that would be overkill.

syldium: In fact, the assertion can be made in the inner class. This removes the array.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -259,18 +266,35 @@ public void removeMetadata(@NotNull String metadataKey, @NotNull Plugin owningPl
...
-		teleported = true;
-		teleportCause = cause;
-		return true;
+		Preconditions.checkNotNull(location, "Location cannot be null"); // The world can be null if it's not a player
+		location.checkFinite();
+		if (this.removed)
+		{
+			return false;
+		}
+
```
Insprill:
```suggestion
		//todo: Add passenger logic

File: src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java [hunk 3]
```diff
@@ -2742,7 +2743,13 @@ public int getPing()
 	public boolean teleport(@NotNull Location location, @NotNull PlayerTeleportEvent.TeleportCause cause)
 	{
 		Preconditions.checkNotNull(location, "Location cannot be null");
+		Preconditions.checkNotNull(location.getWorld(), "World cannot be null");
 		Preconditions.checkNotNull(cause, "Cause cannot be null");
+		location.checkFinite();
+		if (isDead())
+		{
+			return false;
+		}
```
Insprill, line 2746:
```suggestion
		//todo: Add passenger logic

File: src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java [hunk 4]
```diff
@@ -2742,7 +2743,13 @@ public int getPing()
...
 
 		PlayerTeleportEvent playerTeleportEvent = new PlayerTeleportEvent(this, getLocation(), location, cause);
 		Bukkit.getPluginManager().callEvent(playerTeleportEvent);
```
Insprill: This and the `#isCancelled` check can be replaced with `Event#callEvent`

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -2752,7 +2759,13 @@ public boolean teleport(@NotNull Location location, @NotNull PlayerTeleportEvent
 			return false;
 		}
```
Insprill, line 2752: Any open inventories should be closed ([source](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java#827-830))

Insprill (changes requested):

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
Insprill: Exceptions are just ignored in events so the new way can't fail. However, this is an issue with MockBukkit, so I'm going to mark this resolved and fix it in its own PR.

thelooter:
@Insprill Is this good to go

Insprill (approved):
LGTM

PR #530, opened 2022-07-05
Fix exceptions from event handlers not failing tests

Insprill:
# Description
Fixes exceptions thrown from event handlers being caught and just printing a log message. RuntimeExceptions will now be rethrown, and others will be wrapped in a RuntimeException and rethrown.

# Checklist
- [x] Unit tests added.


Insprill:
@thelooter since SonarCloud doesn't like the RuntimeException on line 511, what do you think it should be?
Edit: I've added an EventHandlerException.

thelooter (approved):

PR #546, opened 2022-07-14
Fix ServerMock#getBukkitVersion not returning the full version

### LINKED ISSUE #543: Value of `getBukkitVersion()` is nonstandard

### Minecraft Version

1.19

### Describe the bug

The value of `Bukkit.getBukkitVersion()` does not conform to the expected format containing a revision number and a snapshot qualifier.

### Reproducible Test

Compile this MWE using your tool of choice:

```java
package net.ssterling.example;

import org.bukkit.Bukkit;
import be.seeseemelk.mockbukkit.MockBukkit;
import be.seeseemelk.mockbukkit.ServerMock;

class ServerVersionTest
{
	public static void main(String[] args)
	{
		ServerMock server = MockBukkit.mock();
		String version_string = Bukkit.getBukkitVersion();
		System.out.println("Bukkit API version: " + version_string);
		MockBukkit.unmock();
	}
}
```

The output is `1.19`, not the expected `1.19-R0.1-SNAPSHOT`.

### Anything else?

I’m currently working on a utility to more easily process Bukkit API and vanilla Minecraft version strings.
I’m using MockBukkit for unit tests; one of those tests makes sure the value of `Bukkit.getBukkitVersion()` is correctly detected as a valid Bukkit API server string, making sure I didn’t accidentally screw up the regex.
You can see why the “server” itself providing an incorrect string is a problem for me.

I’d provide a link to the source for the project in question for context, but it’s still a pre-0.1.0 WIP.

Insprill:
# Description
Fixes the non-standard output of `ServerMock#getBukkitVersion`.
Also added tests to ensure this, and `ServerMock#getMinecraftVersion` return something matching the correct pattern.

Closes #543.

# Checklist
- [x] Unit tests added.

# For maintainer

### DIFF
--- a/src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
+++ b/src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
@@ -186,7 +186,7 @@ class ServerMockTest
 	@Test
 	void getBukkitVersion_CorrectPattern()
 	{
-		assertTrue(server.getBukkitVersion().matches("1\\.[0-9]+(\\.[0-9]+)?-R[0-9].[0-9]-SNAPSHOT"));
+		assertTrue(server.getBukkitVersion().matches("1\\.[0-9]+(\\.[0-9]+)?-.*SNAPSHOT.*"));
 	}
 
 	@Test
--- a/src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
+++ b/src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
@@ -184,6 +184,24 @@ class ServerMockTest
 	}
 
 	@Test
+	void getBukkitVersion_CorrectPattern()
+	{
+		assertTrue(server.getBukkitVersion().matches("1\\.[0-9]+(\\.[0-9]+)?-R[0-9].[0-9]-SNAPSHOT"));
+	}
+
+	@Test
+	void getMinecraftVersion_NotNull()
+	{
+		assertNotNull(server.getMinecraftVersion());
+	}
+
+	@Test
+	void getMinecraftVersion_CorrectPattern()
+	{
+		assertTrue(server.getMinecraftVersion().matches("1\\.[0-9]+(\\.[0-9]+)?"));
+	}
+
+	@Test
 	void getName_NotNull()
 	{
 		assertNotNull(server.getName());
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -485,18 +485,16 @@ public class ServerMock extends Server.Spigot implements Server
 	@Override
 	public @NotNull String getBukkitVersion()
 	{
-		return getMinecraftVersion();
+		Preconditions.checkNotNull(this.buildProperties, "Failed to load build properties!");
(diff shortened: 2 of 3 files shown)


src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -183,6 +183,24 @@ void getBukkitVersion_NotNull()
 		assertNotNull(server.getBukkitVersion());
 	}
 
+	@Test
+	void getBukkitVersion_CorrectPattern()
+	{
+		assertTrue(server.getBukkitVersion().matches("1\\.[0-9]+(\\.[0-9]+)?-R[0-9].[0-9]-SNAPSHOT"));
```
thelooter: According to https://www.regexplained.co.uk/ This doesn't seem to match for Dots

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -183,6 +183,24 @@ void getBukkitVersion_NotNull()
...
+	@Test
+	void getMinecraftVersion_NotNull()
+	{
+		assertNotNull(server.getMinecraftVersion());
+	}
+
+	@Test
+	void getMinecraftVersion_CorrectPattern()
+	{
+		assertTrue(server.getMinecraftVersion().matches("1\\.[0-9]+(\\.[0-9]+)?"));
```
thelooter, line 201: According to https://www.regexplained.co.uk/ this doesn't seem to match for dots

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
Insprill: It does match otherwise the test would not pass. If you copy it directly from the string you have to remove the extra backslash before the dots when putting it into a regex validator.

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
Insprill: It does match otherwise the test would not pass. If you copy it directly from the string you have to remove the extra backslash before the dots when putting it into a regex validator.

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
ssterling: If you browse the [SpigotMC Nexus repository](//hub.spigotmc.org/nexus/#view-repositories;snapshots~browsestorage), you’ll see that revision numbers aren’t mandatory, such as in `1.14.3-pre4-SNAPSHOT`.

ssterling (changes requested):
Regex might not match certain future versions; Bukkit devs are somewhat inconsistent with whether they attach a revision number.

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
Insprill: Perhaps ignoring what's between the Minecraft version and SNAPSHOT (and maybe after?) would be better?

ssterling: I didn’t even catch the `-pre4`, so yeah, I agree.

thelooter (approved):

PR #567, opened 2022-07-25
Implement Jumping

thelooter:
# Description
Implement Jumping

# Checklist
- [x] Unit tests added.

# For maintainer

### DIFF
--- a/src/test/java/be/seeseemelk/mockbukkit/entity/LivingEntityMockTest.java
+++ b/src/test/java/be/seeseemelk/mockbukkit/entity/LivingEntityMockTest.java
@@ -3,13 +3,13 @@ package be.seeseemelk.mockbukkit.entity;
 import be.seeseemelk.mockbukkit.MockBukkit;
 import be.seeseemelk.mockbukkit.ServerMock;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.util.UUID;
 
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 class LivingEntityMockTest
 {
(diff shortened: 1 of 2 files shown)


src/test/java/be/seeseemelk/mockbukkit/entity/LivingEntityMockTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+class LivingEntityMockTest
+{
+
+	private ServerMock server;
```
TheBusyBiscuit: This variable is only ever used within the `setup()` method. There is no need to declare it as a class member.

src/test/java/be/seeseemelk/mockbukkit/entity/LivingEntityMockTest.java
```diff
@@ -0,0 +1,45 @@
...
+	@Test
+	void testIsJumpingDefault()
+	{
+		assertFalse(livingEntity.isJumping());
+	}
+
+	@Test
+	void testSetJumping()
+	{
+		livingEntity.setJumping(true);
```
TheBusyBiscuit: This test has no assertion.
I assume you meant to add `assertTrue(livingEntity.isJumping());` here.

TheBusyBiscuit (changes requested):

src/test/java/be/seeseemelk/mockbukkit/entity/LivingEntityMockTest.java
thelooter: Thats how we always did it, even if it could be converted to a local variable. Should I change this or keep it like this to keep it in line with the other Tests?

TheBusyBiscuit: If its not needed, then dont declare it as a class member.

TheBusyBiscuit (approved):
Lgtm

PR #576, opened 2022-07-27
Add an implementation for the passenger API

syldium:
# Description
This PR resolves the TODOs left in #519 by implementing the passenger-related API.

An entity can now have a vehicle and have multiple passengers. This fires four new events:

- Bukkit's `VehicleEnterEvent` only when the vehicle implements `Vehicle`
- Spigot's `EntityMountEvent`
- Bukkit's `VehicleExitEvent` only when the vehicle implements `Vehicle`
- Spigot's `EntityDismountEvent`

In Vanilla/CraftBukkit, the rider always sets the reference to their vehicle first, then the vehicle updates its passenger list:
```mermaid
flowchart TB
    subgraph Passenger
    set[set vehicle]
    unset[unset vehicle]
    end
    subgraph Vehicle
    assertions
    events
    add[add passenger]
    end
    set --> assertions
    assertions --> events
    assertions -->|fail| unset
    events -->|cancelled| unset
    events --> add
```

Because of this, this sample code would dismount the sheep even if it's not the vehicle:
```java
Sheep a = world.spawn(location, Sheep.class);
Sheep b = world.spawn(location, Sheep.class);
a.addPassenger(b);
player.removePassenger(b); // The player is not the vehicle of the sheep, but can still have it dismounted from the real vehicle.
```
This behavior is replicated in this implementation.

Since #519, PaperMC/Paper@5deafd1969b3c521af9cabb46822ecb01e6ce498 has been added to Paper, allowing greater control over teleports with passengers. Teleportation methods have been updated to accept these new parameters. Note that these methods have been annotated with [`@ApiStatus.Experimental`](https://github.com/JetBrains/java-annotations/blob/d438b1132495b7fc5c272289441781d9a0275e4c/common/src/main/java/org/jetbrains/annotations/ApiStatus.java#L54).

# Checklist
- [x] Unit tests added.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
@@ -688,6 +688,11 @@ public abstract class EntityMock extends Entity.Spigot implements Entity, Messag
 	@Override
 	public void remove()
 	{
+		leaveVehicle();
+		if (hasPassengers())
+		{
+			new ArrayList<>(this.passengers).forEach(Entity::leaveVehicle);
+		}
 		this.removed = true;
 	}
 
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
@@ -40,6 +40,8 @@ import org.bukkit.util.Vector;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
@@ -61,6 +63,7 @@ public abstract class EntityMock extends Entity.Spigot implements Entity, Messag
 	private Location location;
 	private boolean teleported;
 	private TeleportCause teleportCause;
+	private final List<Entity> passengers = new ArrayList<>(0);
 	private final MetadataTable metadataTable = new MetadataTable();
 	private final PersistentDataContainer persistentDataContainer = new PersistentDataContainerMock();
 	private boolean operator = false;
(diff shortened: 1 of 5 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -785,22 +852,69 @@ public void playEffect(@NotNull EntityEffect type)
...
+			}
+		}
+		if (!new EntityMountEvent(entity, this).callEvent())
+		{
+			return false;
+		}
+		return this.passengers.add(entity);
+	}
+
+	private boolean tryRemovingPassenger(@NotNull Entity entity)
```
Insprill, line 926: This should have a Javadoc

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -785,22 +852,69 @@ public void playEffect(@NotNull EntityEffect type)
...
 	@Override
-	public Entity getVehicle()
+	public @Nullable Entity getVehicle()
 	{
-		// TODO Auto-generated constructor stub
-		throw new UnimplementedOperationException();
+		return this.vehicle;
+	}
+
+	private boolean tryAddingPassenger(@NotNull Entity entity)
```
Insprill, line 894: This should have a Javadoc

src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -566,13 +566,13 @@ public boolean getCanPickupItems()
 	}
 
 	@Override
-	public boolean teleport(@NotNull Location location, PlayerTeleportEvent.@NotNull TeleportCause cause)
+	public boolean teleport(@NotNull Location location, PlayerTeleportEvent.@NotNull TeleportCause cause, boolean ignorePassengers, boolean dismount)
```
Insprill: Why was this changed?

Insprill (changes requested):
Looks good, just a couple small things

src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
syldium: In order to maintain the health check when using the new Paper's API:
```java
// Redirects to teleport(Location, TeleportCause) that checks health, then uses teleport(Location, TeleportCause, boolean, boolean).
livingEntity.teleport(loc);
// Would skip the health check since this method is only implemented in the super class.
livingEntity.teleport(loc, TeleportCause.PLUGIN, true, true);
```
In fact, the `LivingEntity` class was not modified by PaperMC/Paper@5deafd1969b3c521af9cabb46822ecb01e6ce498, something that has been forgotten?

Insprill: Since it wasn't modified in Paper I feel as if it should not be modified here either to maintain behavior parody. If Paper ever fixes it then we can implement it here.

Insprill (approved):
LGTM

PR #594, opened 2022-08-02
Fix publishing to maven local not working

Insprill:
# Description
Fixes Gradle trying to sign jars locally by checking if the `CI` env var exists, rather than `JITPACK`. ([source](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables))

This also makes it so the version is the git revision prefixed with `dev-` when built locally, since the `mockbukkit.version` property only exists when running in the GitHub action.

# Checklist
- [ ] Unit tests added.

# For maintainer

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,7 +82,7 @@ artifacts {
 def isFork = isFork()
 
 signing {
-    required { !isFork && System.getenv('JITPACK') == null }
+    required { !isFork && System.getenv('CI') != null }
     sign publishing.publications
 }
 
@@ -97,7 +97,7 @@ jacoco {
 }
 
 group = 'com.github.seeseemelk'
-version = property('mockbukkit.version')
+version = this.getVersion()
 
 nexusPublishing {
     repositories {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -82,7 +82,7 @@ artifacts {
 def isFork = isFork()
 
 signing {
-    required { !isFork && System.getenv('JITPACK') == null }
+    required { !isFork && System.getenv('CI') != null }
```
TheBusyBiscuit, line 85: What about jitpack compatibility though?
This would break that, wouldn't it?

TheBusyBiscuit (changes requested):

build.gradle
Insprill: It shouldn't, no. Before, it wouldn't sign if `JITPACK` was present. Now, it won't sign if `CI` *isn't* present, which anywhere outside of GitHub actions, its not.

TheBusyBiscuit: Alright

TheBusyBiscuit (approved):
lgtm

PR #608, opened 2022-08-10
Add Creeper Mock

thelooter:
# Description
Add Creeper Mock

# Checklist
- [x] Unit tests added.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/CreeperMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/CreeperMock.java
@@ -36,17 +36,10 @@ public class CreeperMock extends MonsterMock implements Creeper
 	{
 		CreeperPowerEvent.PowerCause cause = powered ? CreeperPowerEvent.PowerCause.SET_ON : CreeperPowerEvent.PowerCause.SET_OFF;
 
-		if (!this.callPowerEvent(cause))
+		if (new CreeperPowerEvent( this, cause).callEvent())
 		{
 			this.powered = value;
 		}
-	}
-
-	private boolean callPowerEvent(CreeperPowerEvent.PowerCause cause)
-	{
-		CreeperPowerEvent event = new CreeperPowerEvent( this, cause);
-		this.getServer().getPluginManager().callEvent(event);
-		return event.isCancelled();
 	}
 
 	@Override
@@ -108,13 +101,15 @@ public class CreeperMock extends MonsterMock implements Creeper
 	@Override
 	public void setIgnited(boolean ignited)
 	{
-		if (isIgnited() != ignited)
+		if (isIgnited() == ignited)
 		{
-			CreeperIgniteEvent event = new CreeperIgniteEvent(this, ignited);
-			if (event.callEvent())
-			{
-				this.ignited = ignited;
-			}
+			return;
+		}
+
+		CreeperIgniteEvent event = new CreeperIgniteEvent(this, ignited);
+		if (event.callEvent())
+		{
+			this.ignited = ignited;
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/CreeperMock.java
```diff
@@ -0,0 +1,133 @@
...
+			this.powered = value;
+		}
+	}
+
+	private boolean callPowerEvent(CreeperPowerEvent.PowerCause cause)
+	{
+		CreeperPowerEvent event = new CreeperPowerEvent( this, cause);
+		this.getServer().getPluginManager().callEvent(event);
+		return event.isCancelled();
+	}
```
Insprill:
```suggestion
		if (!new CreeperPowerEvent(this, cause).callEvent())
		{
			this.powered = value;
		}
	}
```

src/main/java/be/seeseemelk/mockbukkit/entity/CreeperMock.java
```diff
@@ -0,0 +1,133 @@
...
+	@Override
+	public void ignite()
+	{
+		setIgnited(true);
+	}
+
+	@Override
+	public void setIgnited(boolean ignited)
+	{
+		if (isIgnited() != ignited)
```
Insprill: This can be inverted

Insprill (changes requested):

src/main/java/be/seeseemelk/mockbukkit/entity/CreeperMock.java
thelooter: This is how it's handled in CraftBukkit, thats why I think we should keep it

Insprill: It's only used in one place. We don't need to copy CraftBukkit's code 1 for 1, just as long as the functionality is the same. If it needs to be used in more places in the future, then a method would be good. But for now, doing it this way is much cleaner.

src/main/java/be/seeseemelk/mockbukkit/entity/CreeperMock.java
```diff
@@ -0,0 +1,128 @@
...
+	{
+		return this.powered;
+	}
+
+	@Override
+	public void setPowered(boolean value)
+	{
+		CreeperPowerEvent.PowerCause cause = powered ? CreeperPowerEvent.PowerCause.SET_ON : CreeperPowerEvent.PowerCause.SET_OFF;
+
+		if (new CreeperPowerEvent( this, cause).callEvent())
```
Insprill:
```suggestion
		if (new CreeperPowerEvent(this, cause).callEvent())

File: src/main/java/be/seeseemelk/mockbukkit/entity/CreeperMock.java [hunk 4]
```diff
@@ -0,0 +1,128 @@
...
+		{
+			this.powered = value;
+		}
+	}
+
+	@Override
+	public void setMaxFuseTicks(int ticks)
+	{
+		Preconditions.checkArgument(ticks >= 0, "Ticks need to be bigger than 0");
+
```
Insprill:
```suggestion

File: src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java [hunk 5]
```diff
@@ -1235,6 +1237,14 @@ void testSpawnBee()
 		assertTrue(entity.isValid());
 	}
 
+	@Test
+	void testSpawnCreeper()
+	{
+		WorldMock world = new WorldMock(Material.DIRT, 3);
+		Entity entity = world.spawnEntity(new Location(world, 0, 0, 0), EntityType.CREEPER);
+		assertInstanceOf(CreeperMock.class, entity);
+    }
```
Insprill:
```suggestion
	}

Insprill (changes requested):

Insprill (approved):
LGTM

PR #620, opened 2022-09-02
Implement server icons

Insprill:
# Description
Implements server icons. The `CachedServerIconMock` is currently located in the root package, which I don't like, but I also couldn't think of a better place to put it. If you have a better spot for it please let me know. CraftBukkit has it in the `util` package but I don't feel like that's a great place for it.

# Checklist
- [x] Unit tests added.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -168,7 +168,7 @@ public class ServerMock extends Server.Spigot implements Server
 	private final HelpMapMock helpMap = new HelpMapMock();
 	private final StandardMessenger messenger = new StandardMessenger();
 	private final Map<Integer, MapViewMock> mapViews = new HashMap<>();
-	private CachedServerIconMock serverIcon;
+	private CachedServerIconMock serverIcon = new CachedServerIconMock(null);
 	private int nextMapId = 1;
 
 	private GameMode defaultGameMode = GameMode.SURVIVAL;
@@ -208,8 +208,6 @@ public class ServerMock extends Server.Spigot implements Server
 		{
 			logger.warning("Could not load build properties");
 		}
-
-		loadDefaultIcon();
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -1492,24 +1495,50 @@ public void setWarningState(@NotNull WarningState warningState)
...
+		return new CachedServerIconMock(CachedServerIconMock.PNG_BASE64_PREFIX + encoded);
+	}
+
+	/**
+	 * Loads the server icon from the "server-icon.png" in the servers running directory.
+	 */
+	private void loadDefaultIcon()
+	{
+		this.serverIcon = new CachedServerIconMock(null);
+		File file = new File(new File("."), "server-icon.png");
```
syldium:
```suggestion
		File file = new File("server-icon.png");
```

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -205,8 +209,7 @@ public ServerMock()
 			logger.warning("Could not load build properties");
 		}
 
-		logger.setLevel(Level.ALL);
-
+		loadDefaultIcon();
```
syldium: Is that something we want to cache?

Insprill: That’s how the CraftBukkit implementation works. Perhaps we could add a setter? Not really sure what a use case of that would be. Getting it from disk each call would be deaviating too much since you can’t add the file later and have it update on a real server.

syldium: Yeah, my main concern is that the icon is loaded for every call to `MockBukkit.mock()` and most tests don't need the icon. But the icon is small enough, I guess it shouldn't be an issue to do this I/O operation each time.

Insprill: It doesn't seem to have much of an I/O load. Just looking at task manager while running tests, it looks the same, even on an HDD. I'm still not sure if it's necessary though, as putting an icon there before mocking would be a bit weird. Should we just have a setter instead?

syldium: In my opinion, unit tests should not depend on the file system, it's something for integration tests. In MockBukkit, a world is only in memory. On the another hand, tags are stored as json files and are not lazily initialized, they are read every time.

Having the capacity to just set the icon (`setServerIcon(CachedServerIconMock)`?) would be great for swapping things only in memory.

Insprill: Yup, I fully agree. I've removed the automatic loading and added a setter.

thelooter (approved):
LGTM

PR #626, opened 2022-09-09
Rename build.properties to mockbukkit.properties

Insprill:
# Description
The current build info file, `build.properties`, will get overridden if a project has its own file under the same name. This simply renames it to `mockbukkit.properties` to ensure there are no conflicts.

# Checklist
- [ ] Unit tests added.

# For maintainer


thelooter (approved):

thelooter:
@MockBukkit/maintainers Can we get a review on this?

TheBusyBiscuit (commented):
I don't really know what to say to this.
To me this change is not really useful but I get the idea.

Only concern is that `.properties` usually implies configuration options and `mockbukkit.properties` could signal "this is the place to configure your mockbukkit installation".
Perhaps this is too far-fetched but something like "build data", "metadata" or "build metadata", e.g. "mockbukkit_buildmeta.properties" or similar would make more sense semantically.

Anyway, this use case does not concern me anyway, so this is in no way a rejection or approval, just my thoughts.

Insprill:
That is a fair point. The point of the rename was to reduce the chance of conflicts, and i can see a potential conflict/point of confusion here. It doesn't even need to end with `.properties`, perhaps something like `mockbukkit.metadata` would be better?

Insprill:
Closing since #693 removes this entirely.

PR #629, opened 2022-09-09
Add MessageTarget#assertMoreSaid

Insprill:
# Description
Adds `MessageTarget#assertMoreSaid` which does the opposite of `MessageTarget#assertNoMoreSaid`

# Checklist
- [x] Unit tests added.

# For maintainer


thelooter:
What would be a Usecase of this Method?

Insprill:
A reasonably specific one for me was testing the response of a dynamically generated help command. Instead of testing the response of each line, which may change, it's less maintenance just to test the number of responses. Perhaps it's not the greatest solution.

PR #631, opened 2022-09-09
Remove Projectile's doesBounce implementation

Insprill:
# Description
This removes the implementations of `Projectile#doesBounce`, throwing an UnsupportedOperationException when used. As the Javadoc says, this is an old feature that does not do anything anymore, and therefore should not be used. Even as far back as 1.8.8, there are no usages of it in CraftBukkit.

# Checklist
- [x] Unit tests added.

# For maintainer


thelooter:
Is this Bukkit behavior? I don't think it's a good idea to break compatibility for a feature that while not used anymore, might still be used in unit tests

Insprill:
I went back to CraftBukkit 1.8.8 and there were no usages of it anywhere in the code base. Considering it simply doesn't do anything, it's probably best we fail tests for trying to use it, considering it hasn't been functional for at least the last 7 years.

thelooter (approved):

PR #652, opened 2022-10-02
Implement passenger/vehicle handling when spawning entities

Insprill:
# Description
Uncomments the passenger/vehicle handling code when spawning entities.
To properly support this, `EntityMock#getIndirectPassengers` has been added.

Another minor change that is a little small for it's own PR is passing the proper SpawnReason when dropping items.

# Checklist
- [x] Unit tests added.

# For maintainer


src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -687,6 +687,29 @@ public boolean setPassenger(@NotNull Entity passenger)
 		return Collections.unmodifiableList(this.passengers);
 	}
 
+	/**
+	 * Returns all passengers, and passengers of passengers.
+	 *
+	 * @return All indirect passengers.
+	 */
+	public @NotNull List<Entity> getIndirectPassengers()
```
TheBusyBiscuit: The name would imply that this would not include direct passengers.
The term you are looking for is "transitive" passengers.
Perhaps it would be better to just add an overloaded method `getPassengers(boolean transitive)`.

If true, then return the list as described, if not then fallback to #getPassengers()

The javadocs here should also state that the returned list is ~~immutable~~ unmodifiable.

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -1129,40 +1129,38 @@ private void callSpawnEvent(EntityMock entity, CreatureSpawnEvent.@NotNull Spawn
...
 		}
 		else
 		{
-			canceled = new EntitySpawnEvent(entity).callEvent();
+			canceled = !new EntitySpawnEvent(entity).callEvent();
 		}
 
-		/* EntityMock#getPassengers() and #getVehicle() isn't implemented
-		if (canceled || entity.isValid())
+		if (canceled || !entity.isValid())
```
TheBusyBiscuit: Also shouldn't this be "cancelled" with a double L?

TheBusyBiscuit (dismissed):

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
Insprill: [Both are correct](https://www.grammarly.com/blog/canceled-vs-cancelled/). It just depends where you're from.

TheBusyBiscuit: I know that it is a valid way to spell it. But it's not inline with the rest of the codebase and Bukkit... that's why I am pointing this inconsistency out.

Insprill: Ah, not sure if I've ever noticed that before. Fixed.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -687,6 +687,34 @@ public boolean setPassenger(@NotNull Entity passenger)
...
+		{
+			entities.add(passenger);
+			if (passenger instanceof EntityMock mock)
+			{
+				entities.addAll(mock.getPassengers(true));
+			}
+			else
+			{
+				entities.addAll(passenger.getPassengers());
+			}
```
syldium:
```suggestion
			entities.addAll(((EntityMock) passenger).getPassengers(transitive));
```
Passengers are [always instances of `EntityMock`](https://github.com/MockBukkit/MockBukkit/blob/e5631cc1b5da36af1cbfd7d04ee2412fcd55619c/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java#L713) in order to have a reference to the vehicle in the passenger.

Using `transitive` is more descriptive than using `true`. Most IDEs would warn that this could be simplified by using `true` directly, and that's the case. Conceptually, the `transitive` parameter completely changes the algorithm used, so in my opinion, a new method with another name (like the first implementation in this PR) would be preferable as it is quickly identifiable.

Insprill: I think leaving it as an overload is a little simpler, but I can definitely see where confusion could come from. @thelooter What do you think?

thelooter: I personally prefer the version  @syldium proposed

thelooter:
@Insprill  can this be merged

thelooter (approved):

PR #653, opened 2022-10-02
Add BlockDataMock support for TrapDoor

spscally:
# Description
Add `BlockDataMock` support for `TrapDoor`. Also:
- moved `BlockData` String keys into consts class `BlockDataKey` to reduce duplication
- refactored `BlockDataMock.mock()` since existing implementation would get rather large if more `Tag`-based subclasses are added

# Checklist
- [x] Unit tests added.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
@@ -22,16 +22,7 @@ import java.util.stream.Collectors;
 
 public class BlockDataMock implements BlockData
 {
-
-	protected static final String FACING = "facing";
-	protected static final String HALF = "half";
-	protected static final String OCCUPIED = "occupied";
-	protected static final String OPEN = "open";
-	protected static final String PART = "part";
-	protected static final String POWERED = "powered";
-	protected static final String SHAPE = "shape";
-	protected static final String TYPE = "type";
-	protected static final String WATERLOGGED = "waterlogged";
+	private static final String NULL_MATERIAL_EXCEPTION_MESSAGE = "Material cannot be null";
 
 	private final @NotNull Material type;
 	private final @NotNull Map<String, Object> data;
@@ -209,7 +200,7 @@ public class BlockDataMock implements BlockData
 
 	public static @NotNull BlockDataMock mock(@NotNull Material material)
 	{
-		Preconditions.checkNotNull(material, "Material cannot be null");
+		Preconditions.checkNotNull(material, NULL_MATERIAL_EXCEPTION_MESSAGE);
 		// Special Cases
 		BlockDataMock mock = attemptMockByPaperMaterialTags(material);
 		if (mock != null)
@@ -237,9 +228,9 @@ public class BlockDataMock implements BlockData
 	 * @param material Material which we will attempt to mock
 	 * @return BlockDataMock if matched, null otherwise
 	 */
-	static BlockDataMock attemptMockByPaperMaterialTags(@NotNull Material material)
+	private static BlockDataMock attemptMockByPaperMaterialTags(@NotNull Material material)
 	{
-		Preconditions.checkNotNull(material, "Material cannot be null");
+		Preconditions.checkNotNull(material, NULL_MATERIAL_EXCEPTION_MESSAGE);
 		if (MaterialTags.BEDS.isTagged(material))
 		{
 			return new BedMock(material);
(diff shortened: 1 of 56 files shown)


thelooter:
Please fix  the Codesmells and try to deduplicate the Code if possible

spscally:
> Please fix the Codesmells and try to deduplicate the Code if possible

Can do

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -201,23 +211,65 @@ public boolean equals(Object obj)
...
+	}
+
+	/**
+	 * Attempts to construct a BlockDataMock object by matching against Paper MaterialTags. Returns null if the given
+	 * material does not match any supported MaterialSetTag.
+	 *
+	 * @param material Material which we will attempt to mock
+	 * @return BlockDataMock if matched, null otherwise
+	 */
+	static BlockDataMock attemptMockByPaperMaterialTags(@NotNull Material material)
```
Insprill: Can this be private? I don't see a need for anything to ever access this outside of the mock method.

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -201,23 +211,65 @@ public boolean equals(Object obj)
...
+	}
+
+	/**
+	 * Attempts to construct a BlockDataMock object by matching against Bukkit Tags. Returns null if the given material
+	 * does not match any supported Tag.
+	 *
+	 * @param material Material which we will attempt to mock
+	 * @return BlockDataMock if matched, null otherwise
+	 */
+	static BlockDataMock attemptMockByTag(@NotNull Material material)
```
Insprill: Can this be private? I don't see a need for anything to ever access this outside of the mock method.

Insprill (changes requested):

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
spscally: Good point and yes I can change that

spscally:
I've been thinking about how to deduplicate the code and I'm not sure there's a way of doing so without rewriting `BlockDataMock`. Here's why I think that:

Sonar is complaining about how `getHalf()`/`setHalf()`/getFacing()`/setFacing()`/`getFaces()` are duplicated in `TrapDoorMock`. If we wanted to put the `BlockFace` methods in a separate class, we could have an `abstract class DirectionalMock` that `extends BlockDataMock` and `implements Directional` (though `getFaces()` will have to differ for different subclasses). `StairsMock`, for example, could then extend `DirectionalMock` and that would reduce the duplication of those methods. However, when we also try dedupe the `Half` methods, we'd need something like `BisectedMock`, and we'd run into a multiple inheritance problem.

We _could_ rewrite `BlockDataMock` to be a subinterface of `BlockData` and then allow for "mixing in" of `Directional`/`Bisected`/`Waterlogged`/etc methods, but I don't think I'm comfortable (nor do I think y'all would be) doing that having just started looking at the source code last week.

What are your thoughts? Is the duplication something you'd like for me to take a deeper look into?

(As for the code smells, I don't think there should be a problem fixing those.)

Insprill:
It's something that would be nice to have done, but not a huge priority. If you're interested in it you can take a shot at it, but it's not required to keep adding new block data.

spscally:
I addressed the smells. Also, I added `jacocoTestReport` to the `test` task, as I don't think it was running on `test` automatically.

spscally:
> It's something that would be nice to have done, but not a huge priority. If you're interested in it you can take a shot at it, but it's not required to keep adding new block data.

Let me sit and think on that. I might be interested in trying my hand at it, but as of right now I don't have any ideas.

build.gradle
```diff
@@ -84,6 +84,7 @@ signing {
 }
 
 jacocoTestReport {
+    dependsOn test
```
Insprill: Please create a separate PR for this as it would require seeseemelk or TheBusyBiscuit to approve.

build.gradle
```diff
@@ -147,6 +148,7 @@ publishToMavenLocal {
 
 test {
     useJUnitPlatform()
+    finalizedBy jacocoTestReport
```
Insprill: Same here.

Insprill (changes requested):

build.gradle
spscally: I moved those changes to #655

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
spscally: Addressed in my last commit

Insprill: Are you sure? They're still both package-private on the web viewer. Please don't force push unless you have to, it's nice to have a proper git history, and it makes things like this easier because I could've just looked at that specific commit.

spscally: Clearly not and I must have messed up resetting the commit. _Now_ it should be added.

And sorry about the force-push; I thought it would make the commit history for the changes more compact and less "noisy", but I didn't realize it would make reviewing it on your end more difficult. I will avoid that in the future.

Insprill (approved):
LGTM. Thank you!

spscally:
> LGTM. Thank you!

And thank you as well! This project wouldn't be possible without you two and I really appreciate your thoroughness and patience 😀

PR #656, opened 2022-10-08
Remove removed entities from the server

syldium:
# Description
~The `WorldMock` implementation was waiting for an implementation of `EntityMock#getPassengers()`. Now that is the case (#576), the vehicle and the passengers can be properly removed if the spawn event is cancelled.~ Removed in order not to conflict with #652 

Related to entity removal, the `ServerMock` class now also have an `unregisterEntity` method so that removed entities are not kept.

# Checklist
- [x] Unit tests added.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -232,6 +232,7 @@ public class ServerMock extends Server.Spigot implements Server
 	 */
 	public void registerEntity(@NotNull EntityMock entity)
 	{
+		Preconditions.checkNotNull(entity, "Entity cannot be null");
 		AsyncCatcher.catchOp("entity add");
 		entities.add(entity);
 	}
@@ -243,6 +244,8 @@ public class ServerMock extends Server.Spigot implements Server
 	 */
 	public void unregisterEntity(@NotNull EntityMock entity)
 	{
+		Preconditions.checkNotNull(entity, "Entity cannot be null");
+		Preconditions.checkArgument(!entity.isValid(), "Entity is not marked for removal");
 		AsyncCatcher.catchOp("entity remove");
 		entities.remove(entity);
 	}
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -236,6 +236,17 @@ public void registerEntity(@NotNull EntityMock entity)
 		entities.add(entity);
 	}
 
+	/**
+	 * Unregisters an entity from the server. Should only be used internally.
+	 *
+	 * @param entity The entity to unregister
+	 */
+	public void unregisterEntity(@NotNull EntityMock entity)
+	{
```
Insprill, line 246:
```suggestion
	{
		Preconditions.checkNotNull(entity, "Entity cannot be null");

File: src/main/java/be/seeseemelk/mockbukkit/ServerMock.java [hunk 2]
```diff
@@ -236,6 +236,17 @@ public void registerEntity(@NotNull EntityMock entity)
...
+		AsyncCatcher.catchOp("entity remove");
+		entities.remove(entity);
```
Insprill, line 250: Perhaps this should check if the entity is valid, and remove it if it is? Despite it being labeled for internal use only, people will still try to use it and I'm sure ask for help when it causes weird issues.

Insprill (changes requested):

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
Insprill: Was this intended to be marked as resolved?

syldium: A precondition has been added to check if the entity is not valid, in other words, if the `remove` method was called first. The server should continue to track any valid entities, right?

Insprill: Ah, I somehow looked over that; thought just the null check was added. Yeah, that's fine.

Insprill (approved):
LGTM

PR #667, opened 2022-10-16
Implement Adventure's #playSound methods

syldium:
# Description
Adventure's `#playSound` methods are now overridden. The `AudioExperience` type is reused, and Adventure's `Sound.Source` is converted to Bukkit's `SoundCategory`. For now, in MockBukkit, sounds cannot follow entities and are bound to a location: emitters are assumed to be immobile.

Some `assertSoundHeard` methods have been added to check with the Adventure's `Sound` type, as well as the volume, the pitch and the source.

Two stubs have been added in order to compile, but are not related to this PR. `BedMock#setOccupied` has been implemented because the necessary parts were already present.

# Checklist
- [x] Unit tests added.
- [x] Code follows existing (local) code format.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/data/BedMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/data/BedMock.java
@@ -43,6 +43,12 @@ public class BedMock extends BlockDataMock implements Bed
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


Insprill (approved):
LGTM, Thank you!

PR #670, opened 2022-10-18
Use BlockDataMock.mock() for internal BlockDataMock construction

spscally:
# Description
If you previously tried to create a `BlockData` indirectly, it would always create an instance of `BlockDataMock`, even if the given `Material` had a special subclass implementation. For example, if you did:

```
BlockMock block = new BlockMock(Material.OAK_SLAB);
BlockDataMock blockData = block.getBlockData();

assertInstanceOf(Slab.class, blockData);
```

Then the assertion would fail. `blockData` would _not_ be an instance of `Slab` or `SlabMock`, but would only be an instance of `BlockDataMock`. This is because the `BlockMock` constructor (and a couple of other places) were using the `BlockDataMock` constructor rather than the static factory method `BlockDataMock.mock()`.

I left all references to the `BlockDataMock` constructor that were just constructing with `Material.AIR`.

# Checklist
- [ ] Unit tests added.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/MockChunkData.java
@@ -63,14 +63,16 @@ public class MockChunkData implements ChunkGenerator.ChunkData
 	@Override
 	public void setBlock(int x, int y, int z, @NotNull Material material)
 	{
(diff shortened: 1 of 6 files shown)


Insprill (approved):
LGTM. Thank you!

PR #677, opened 2022-10-26
Add a MockBukkitExtension for use with JUnit

spscally:
# Description
Creating `MockBukkitExtension` that you can use to automatically:
- set up the `ServerMock` before each test
- inject the `ServerMock` into any fields/parameters of the extended test class
- unmock the `ServerMock` after each test

I've also converted some test classes to use it to provide some examples.

A previous iteration of this branch also had a `MockBukkitServer` annotation, which could be put on any `ServerMock` field/param you wanted to mark for injection. I decided to remove this because there can only ever be one `ServerMock` (AFAICT) so it seemed redundant. I also couldn't come up with a good name for this annotation that wouldn't be somewhat confusing.

You can now do the following, without having to have `@BeforeEach`- and `@AfterEach`-annotated methods:

```
@ExtendWith(MockBukkitExtension.class)
class ExampleTest
{
    private ServerMock serverMock;

    @Test
    void testStuff()
    {
        serverMock.addPlayer();
        // ...
    }
}
```

I've marked the extension class as `Experimental` for now, in case there is something I've overlooked. (I'm specifically uncertain about the interaction between multiple extensions on a single test class.) Also: IntelliJ does not recognize that the field in the example above _is_ actually being set by the extension and will give you a warning. I haven't been able to figure out a solution to that that doesn't involve the user having to modify their inspections settings (see, for example: https://stackoverflow.com/questions/6762012/suppress-variable-is-never-assigned-warning-in-intellij-idea).

# Checklist
- [ ] Unit tests added.

# For maintainer


thelooter:
Would it be possible to store the Server Instance created in the Extension and provide Accessors for it?

spscally:
> Would it be possible to store the Server Instance created in the Extension and provide Accessors for it?

I don't know (my guess is no), but I'm going to take a swing at implementing field injection using an attribute. (Here's how that looks with `MockitoExtension`: https://stackoverflow.com/a/40962941)

spscally:
I've managed to get parameter resolution working but don't quite have it for fields just yet. I pushed a commit showing how the former works and how the latter would ideally work.

TheBusyBiscuit:
I gotta say this is a very nice and useful addition to this project, thanks a lot! <3
I will have a look at the code later

spscally:
> I gotta say this is a very nice and useful addition to this project, thanks a lot! <3 I will have a look at the code later

No problem and sounds good!

spscally:
I've added field injection but there's still more testing/verification I want to do before marking the PR as ready for review

thelooter:
Also please fix that Code Smell

thelooter (approved):
From what I can tell this looks good to me, but I'd prefer some more Reviews from @Insprill @TheBusyBiscuit @seeseemelk

src/main/java/be/seeseemelk/mockbukkit/MockBukkitExtension.java
```diff
@@ -0,0 +1,136 @@
...
+
+	private void injectServerMockIntoFields(ExtensionContext context) throws IllegalAccessException
+	{
+		final Optional<Class<?>> classOptional = context.getTestClass();
+		if (classOptional.isEmpty())
+			return;
+
+		final List<Field> serverMockFields = FieldUtils.getAllFieldsList(classOptional.get())
+				.stream()
+				.filter(field -> field.getType() == ServerMock.class)
```
TheBusyBiscuit, line 101: It may be a wise idea to also require an annotation for fields, as it may lead to unexpected results and confusion if devs do not explicitly define this behaviour.

TheBusyBiscuit (dismissed):

src/main/java/be/seeseemelk/mockbukkit/MockBukkitExtension.java
```diff
@@ -0,0 +1,136 @@
...
+
+		final Optional<Object> optionalTestInstance = context.getTestInstance();
+		if (optionalTestInstance.isEmpty())
+			return;
+
+		final Object testInstance = optionalTestInstance.get();
+		for (final Field field : serverMockFields)
+		{
+			final String name = field.getName();
+			FieldUtils.writeDeclaredField(testInstance, name, MockBukkit.getOrCreateMock(), true);
```
TheBusyBiscuit: Also, instead of calling `MockBukkit.getOrCreateMock()` here again, you can pass the instance via a parameter since this method is called right after `MockBukkit.getOrCreateMock()` was already executed.

spscally: Sounds good. I had implemented it this way originally to avoid adding an extra param to the helper method, but passing it explicitly is probably cleaner and easier to read

src/main/java/be/seeseemelk/mockbukkit/MockBukkitExtension.java
spscally: I've added `MockBukkitServer` back but I'm still not fully convinced that it's the right name. What do you think?

Another option could be something like `MockBukkitInject`, which could open the door to more injected objects in the future (but could create more confusion in the present since only `ServerMock` is supported).

spscally: Hey apologies to bug you but any thoughts on the above?

thelooter:
@TheBusyBiscuit Do you still want any changes on this, or can this be merged?

seeseemelk (approved):

src/main/java/be/seeseemelk/mockbukkit/MockBukkitExtension.java
seeseemelk: I like the idea of calling it `MockBukkitInject`. I don't think it would cause much confusion, and I do like the idea of being able to inject more objects in the future (perhaps Worlds, Players, other entities, ...).

spscally: Ok I'll push a commit to rename it!

PR #682, opened 2022-10-30
Use the default implementation for Chunk#getChunkKey

syldium:
# Description
The `ChunkMock#getChunkKey(int chunkKey)` method was overridden in MockBukkit, despite having a functional default implementation. This PR removes the redefinition to use it.

A chunk key is hard-coded into the test instead of using the static `Chunk#getChunkKey()` method to avoid depending on it.

# Checklist
- [x] Unit tests added.

# For maintainer


Insprill (approved):
LGTM

PR #697, opened 2022-12-01
Stub new methods

Insprill:
# Description
Seems like the Paper devs have had some fun recently.

# Checklist
- [ ] Unit tests added.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/SchoolableFishMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/SchoolableFishMock.java
@@ -1,0 +1,65 @@
+package be.seeseemelk.mockbukkit.entity;
+
+import be.seeseemelk.mockbukkit.ServerMock;
+import be.seeseemelk.mockbukkit.UnimplementedOperationException;
+import io.papermc.paper.entity.SchoolableFish;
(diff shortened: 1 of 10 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/SchoolableFishMock.java
```diff
@@ -0,0 +1,65 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import java.util.UUID;
+
+/**
+ * Mock implementation of a {@link SchoolableFish}.
+ *
+ * @see FishMock
+ */
+public abstract class SchoolableFishMock extends FishMock implements SchoolableFish
```
thelooter, line 16: This should be it's own PR

thelooter (changes requested):
One small Change

src/main/java/be/seeseemelk/mockbukkit/entity/SchoolableFishMock.java
Insprill: SchoolableFish is new, so without it those methods would need to be duplicated on all the schoolable fish classes.

thelooter (approved):

PR #708, opened 2022-12-12
Remove hamcrest as a dependency

Insprill:
# Description
Hamcrest was being used in a total of 6 tests, and it's not worth having an entire dependency for that, when regular JUnit assertions do the job perfectly well.

# Checklist
- [x] Unit tests added.

# For maintainer


TheBusyBiscuit (approved):
Yeah, it is fair to assume the remaining tests won't ever be updated to hamcrest at this point.
Not sure if I ever brought it up but I am definitely in favour of removing this dependency.

PR #712, opened 2022-12-15
Add methods to the SignStateMock

Thorinwasher:
# Description
Implemented Sign#setColor, Sign#getColor, Sign#setGlowingText and Sign#getGlowingText

# Checklist
- [x] Unit tests added.

# For maintainer

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
@@ -141,7 +141,7 @@ public class SignMock extends TileStateMock implements Sign
 	@Override
 	public boolean isGlowingText()
 	{
-		return glowing;
+		return this.glowing;
 	}
 
 	@Override
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
@@ -64,8 +64,8 @@ public class SignMock extends TileStateMock implements Sign
 		{
 			lines[i] = state.getLine(i);
 		}
-		color = state.getColor();
-		glowing = state.isGlowingText();
+		this.color = state.getColor();
+		this.glowing = state.isGlowingText();
 	}
 
 	@Override
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
@@ -24,6 +24,8 @@ public class SignMock extends TileStateMock implements Sign
 {
 
 	private final String[] lines = { "", "", "", "" };
+	private DyeColor color = DyeColor.BLACK;
+	private boolean glowing;
 
 	/**
 	 * Constructs a new {@link SignMock} for the provided {@link Material}.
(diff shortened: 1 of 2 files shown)


src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
```diff
@@ -137,29 +141,26 @@ public void setEditable(boolean editable)
...
 	@Override
 	public DyeColor getColor()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return color;
 	}
 
 	@Override
 	public void setColor(DyeColor color)
```
Insprill: This should be annotated `@NotNull` and have a Preconditions check guarding it. A test will also need to be added. In Paper it has undefined nullability, but will throw an NPE if null (`CraftSign:99` as of `20503be`)

src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
```diff
@@ -137,29 +141,26 @@ public void setEditable(boolean editable)
...
-		throw new UnimplementedOperationException();
+		this.glowing = glowing;
 	}
 
 	@Override
 	public DyeColor getColor()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return color;
```
Insprill:
```suggestion
		return this.color;

File: src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java [hunk 3]
```diff
@@ -137,29 +141,26 @@ public void setEditable(boolean editable)
 	@Override
 	public boolean isGlowingText()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return glowing;
```
Insprill:
```suggestion
		return this.glowing;

File: src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java [hunk 4]
```diff
@@ -62,6 +64,8 @@ protected SignMock(@NotNull SignMock state)
 		{
 			lines[i] = state.getLine(i);
 		}
+		color = state.getColor();
+		glowing = state.isGlowingText();
```
Insprill:
```suggestion
		this.color = state.getColor();
		this.glowing = state.isGlowingText();

Insprill (changes requested):
Would it also be possible to implement `#setEditable`/`#isEditable`? It's also a simple getter/setting defaulting to true.

Thorinwasher:
About also setting setEditable / getEditable. I don't know if it has something more that needs to be implemented that deals with that behavior, for example when a player tries to edit a sign block it (I really don't understand when the editable feature will be used). I'm basically not knowledgable enough to do any pull requests about that

Insprill:
It looks like it's only used when a client sends the [update sign](https://wiki.vg/Protocol#Update_Sign) packet which we don't have implemented. If you don't want to do it that's fine, it can always be implemented later.

src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
Thorinwasher: I don't think I understand this. I am checking if not null in the setter, and I'm also testing it in SignMockTest#setGetColor

Insprill: Ahh, I completely glanced over that :joy:, sorry. It just needs the NotNull annotation then

Insprill (approved):
LGTM. Thank you!

PR #714, opened 2022-12-22
OfflinePlayerMock: fix incorrect return type

seeseemelk:
# Description
This fixes a build error currently present in main.
The issue was that upstream Paper changed the specific type returned by the `getPlayerProfile` method.


thelooter (approved):
LGTM

PR #747, opened 2023-03-17
Update to 1.19.4

thelooter:
# Description
Update to 1.19.4

# Checklist
- [x] Fix documentation
- [x] Write note in readme detailing the change to plugin constructors. The old constructors are no longer used


seeseemelk:
I added some things to the checklist

build.gradle
```diff
@@ -30,6 +30,8 @@ dependencies {
 
 	// Dependencies for Unit Tests
 	implementation 'org.junit.jupiter:junit-jupiter:5.9.2'
+	implementation 'net.bytebuddy:byte-buddy:1.14.3'
+	implementation 'net.bytebuddy:byte-buddy-agent:1.14.3'
```
seeseemelk: This one was only used during a single experiment, it can be removed

```suggestion
```

build.gradle
```diff
@@ -144,14 +146,15 @@ publishToMavenLocal {
...
+	dependsOn(gradle.includedBuild("TestPlugin").task(":jar"))
 	useJUnitPlatform()
 }
 
 eclipse.classpath.downloadJavadoc = true
 eclipse.classpath.downloadSources = true
 
 def isFork() {
-	try {
+	try {cdc
```
seeseemelk: Hmm, this doesn't seem right.
I'm frankly surprised this even compiles (or maybe it won't once merged)

```suggestion
	try {
```

extra/TestPlugin/TestPlugin.jar
seeseemelk, line 1: This file can be removed completely, we now build it as a composite build when running unit tests

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -307,16 +309,32 @@ public boolean teleport(@NotNull Location location)
...
+	public boolean teleport(@NotNull Location location, @NotNull TeleportCause cause, TeleportFlag @NotNull ... flags)
 	{
 		Preconditions.checkNotNull(location, "Location cannot be null"); // The world can be null if it's not a player
 		location.checkFinite();
+
+		Set<TeleportFlag> flagSet = Set.of(flags);
+		boolean dismount = !flagSet.contains(TeleportFlag.EntityState.RETAIN_VEHICLE);
+		boolean ignorePassengers = flagSet.contains(TeleportFlag.EntityState.RETAIN_PASSENGERS);
+
+		if (flagSet.contains(TeleportFlag.EntityState.RETAIN_PASSENGERS) && this.hasPassengers() && location.getWorld() != this.getWorld())
```
seeseemelk: `location.getWorld() != this.getWorld()`

Is this the right way to compare these?

src/main/java/be/seeseemelk/mockbukkit/entity/MushroomCowMock.java
```diff
@@ -79,4 +91,47 @@ public void setStewEffect(@Nullable PotionEffectType type)
...
 	}
 
+	@Override
+	public void shear(Sound.@NotNull Source source)
+	{
+		WorldMock world = (WorldMock) this.getWorld();
+		Location location = this.getLocation();
+		world.playSound(this, org.bukkit.Sound.ENTITY_MOOSHROOM_SHEAR, AdventureConverters.soundSourceToCategory(source), 1.0F, 1.0F);
+		Cow cow = world.spawn(this.getLocation(), Cow.class, null, CreatureSpawnEvent.SpawnReason.CUSTOM, true, false);
+//		world.spawnParticle(Particle.EXPLOSION_NORMAL, location.getX(), location.getY() + 0.5, location.getZ(), 1);
```
seeseemelk: Why is this in comments

src/main/java/be/seeseemelk/mockbukkit/entity/MushroomCowMock.java
```diff
@@ -79,4 +91,47 @@ public void setStewEffect(@Nullable PotionEffectType type)
 		return EntityType.MUSHROOM_COW;
 	}
 
+	@Override
+	public void shear(Sound.@NotNull Source source)
+	{
+		WorldMock world = (WorldMock) this.getWorld();
```
seeseemelk: Does this not already return a WorldMock?
If not, can we change it so that the cast is unnecessary here?

src/main/java/be/seeseemelk/mockbukkit/entity/SheepMock.java
```diff
@@ -54,11 +59,30 @@ public void setColor(DyeColor color)
...
 	{
 		return EntityType.SHEEP;
 	}
 
+	@Override
+	public void shear(Sound.@NotNull Source source)
+	{
+		this.getWorld().playSound(this, org.bukkit.Sound.ENTITY_SHEEP_SHEAR, AdventureConverters.soundSourceToCategory(source), 1.0F, 1.0F);
+		this.setSheared(true);
+		int i = 1 + ThreadLocalRandom.current().nextInt(3);
```
seeseemelk:
```suggestion
		int i = ThreadLocalRandom.current().nextInt(1, 4); 
```

seeseemelk (changes requested):

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -326,7 +326,7 @@ public boolean teleport(@NotNull Location location, @NotNull TeleportCause cause
 		boolean dismount = !flagSet.contains(TeleportFlag.EntityState.RETAIN_VEHICLE);
 		boolean ignorePassengers = flagSet.contains(TeleportFlag.EntityState.RETAIN_PASSENGERS);
 
-		if (flagSet.contains(TeleportFlag.EntityState.RETAIN_PASSENGERS) && this.hasPassengers() && location.getWorld() != this.getWorld())
+		if (flagSet.contains(TeleportFlag.EntityState.RETAIN_PASSENGERS) && this.hasPassengers() && location.getWorld().equals(this.getWorld()))
```
Insprill, line 330: The other way is also correct, both in MockBukkit and Paper

src/main/java/be/seeseemelk/mockbukkit/entity/MushroomCowMock.java
Insprill: Not implemented, added a note in the code.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
Insprill: Yes. `#equals` isn't overridden in MockBukkit or Paper.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
thelooter: The usual rule of thumb I follow is to only use == / != for comparing primitives. For Objects I use equals

seeseemelk: `==`/`!=` compares identity while `Object#equals` compares content. Although the default implementation for `Object#equals` does just implement an identity check.

In addition, `a.equals(b)` will throw a null-pointer exception if `a` is null while `a == b` will never throw exceptions (although, I'm not sure of this actually. Perhaps it can throw a null pointer exception with automatic unboxing of a null?).

Null-pointer safe way of doing a content check: `Objects.equals(a, b)`

seeseemelk (approved):

seeseemelk:
Looks good to me. Now just for @Insprill to resolve the last conversation

Insprill (approved):
LGTM

PR #750, opened 2023-04-03
Add Camel Mock

thelooter:
# Description
Adds a Mock for Camels

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/CamelMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/CamelMock.java
@@ -15,7 +15,7 @@ public class CamelMock extends AbstractHorseMock implements Camel
 	private boolean isSitting = false;
 
 	/**
-	 * Constructs a new {@link AbstractHorseMock} on the provided {@link ServerMock} with a specified {@link UUID}.
+	 * Constructs a new {@link CamelMock} on the provided {@link ServerMock} with a specified {@link UUID}.
 	 *
 	 * @param server The server to create the entity on.
 	 * @param uuid   The UUID of the entity.
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/CamelMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/CamelMock.java
@@ -1,0 +1,64 @@
+package be.seeseemelk.mockbukkit.entity;
+
+import be.seeseemelk.mockbukkit.ServerMock;
+import org.bukkit.entity.Camel;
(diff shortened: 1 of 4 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/CamelMock.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.UUID;
+
+public class CamelMock extends AbstractHorseMock implements Camel
+{
+
+	private boolean isDashing = false;
+	private boolean isSitting = false;
+
+	/**
+	 * Constructs a new {@link AbstractHorseMock} on the provided {@link ServerMock} with a specified {@link UUID}.
```
Insprill: Please update this Javadoc with the correct class name

Insprill (changes requested):

src/test/java/be/seeseemelk/mockbukkit/entity/CamelMockTest.java
```diff
@@ -0,0 +1,71 @@
...
+	{
+		assertFalse(camel.isDashing());
+	}
+
+	@Test
+	void testIsDashing()
+	{
+		camel.setDashing(true);
+		assertTrue(camel.isDashing());
+	}
```
tatanpoker09, line 56: We could just theoretically include a test to see if setting dashing back to false also reverts back the isDashing condition, might be a little overkill and unnecessary, it does cover the other condition.

src/test/java/be/seeseemelk/mockbukkit/entity/CamelMockTest.java
```diff
@@ -0,0 +1,71 @@
...
+	{
+		assertFalse(camel.isSitting());
+	}
+
+	@Test
+	void testSetSitting()
+	{
+		camel.setSitting(true);
+		assertTrue(camel.isSitting());
+	}
```
tatanpoker09, line 69: Same here with sitting, up to your decision. Other than that LGTM

tatanpoker09 (approved):
LGTM. Couple minor things that *might* be considerable, but it looks good.

src/test/java/be/seeseemelk/mockbukkit/entity/CamelMockTest.java
Insprill: We don't do that anywhere else, so it'd have to be updated everywhere for consistency, and that's out of the scope of this PR. The odds of that being an issue are extremely negligible, so we'll likely never do it.

Insprill (approved):
LGTM

PR #757, opened 2023-04-21
Fix PlayerMock#disconnect not firing PlayerQuitEvent

### LINKED ISSUE #691: Disconnect doesn't call PlayerQuitEvent if PlayerMock is constructed with ServerMock#addPlayer(String)

### Minecraft Version

1.19

### Describe the bug

Disconnect doesn't appear to be calling PlayerQuitEvent if the PlayerMock is built using only a name and not a UUID. If a player is created with just a name, no UUID, the `online` field in the PlayerMock is set to false, even though `isOnline()` will return true because they're in the online player list. `disconnect()` uses that `online` field to determine if the event should fire but it is `false`, so the event doesn't fire.

If I create the player with new and with a UUID, the disconnect method works as I would assume it should but with just the ServerMock#addPlayer it does not.

### Reproducible Test

```java
    @Test
    public void quitNoWork() {
        PlayerMock travja = server.addPlayer("Travja");
//        Comment ^ and uncomment below and it works
//        PlayerMock travja = new PlayerMock(server, "Travja", UUID.randomUUID());
//        server.addPlayer(travja);
        travja.disconnect();
        server.getPluginManager().assertEventFired(PlayerQuitEvent.class, e -> {
            System.out.println("Event was fired!");
            return true;
        });
    }
```

### Anything else?

[PlayerMock:164](https://github.com/MockBukkit/MockBukkit/blob/v1.19/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java#L164) indicates the player should be considered offline, but this field is not updated when `ServerMock#addPlayer` is called.

thelooter:
# Description
Fixes #691

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -345,7 +345,7 @@ public class ServerMock extends Server.Spigot implements Server
 	public @NotNull PlayerMock addPlayer(@NotNull String name)
 	{
 		AsyncCatcher.catchOp("player add");
-		PlayerMock player = new PlayerMock(this, name);
+		PlayerMock player = new PlayerMock(this, name, UUID.randomUUID());
 		addPlayer(player);
 		return player;
 	}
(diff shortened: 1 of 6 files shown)


Insprill (changes requested):
We should keep the constructor, but just make it call the other one, passing in `UUID#randomUUID()`

Insprill (approved):
LGTM

PR #761, opened 2023-04-23
Fix loading of .properties files from MockBukkit resources

sowelipililimute:
# Description
use the correct method to obtain the resource stream that would get packaged with the mockbukkit jar

# Checklist


Insprill (approved):
LGTM

PR #786, opened 2023-05-28
Documentation additions

Zorua162:
# Description

 - Made the note about updating the version in the Maven dependency be highlighted, so that its more likely to be read
 - Updated the version in the Maven pom example to be 3.3.0, to ensure that those who don't realize that they need to update it get a version that is >3, to avoid some common issues that are seen on the discord
 - Added to the version note that people should put the maven dependency above their bukkit providing dependency
 - Updated the example code to use Junit5 syntax instead of Junit4, this should mean that once its copied in IDE's will import Junit5 for those who do this
# Checklist


thelooter (approved):
LGTM, but this needs to be reviewed by either @seeseemelk or @TheBusyBiscuit

thelooter:
Can you rebase and adapt this for 1.20

Zorua162:
I don't think I did the rebase right, if you aren't happy with this feel free to close this PR, and I will try again locally (the rebase is likely to have caused too many issues in this PR now). Alternatively I might just create a new branch from v1.20 and manually copy the changes over, as I think that may be easier at this point from my previous experience with rebasing.

Insprill (approved):
The rebase seems fine to me. Thank you!

TheBusyBiscuit (approved):
LGTM

PR #796, opened 2023-06-27
Fix Gamerule Value not being set due to type Mismatch

thelooter:
# Description
FIxes the Value of Gamerules not being set since there was a Mismatch between `GameRule#getType()` and `Boolean.TYPE` / `Integer.Type`

# Checklist


Insprill (approved):
LGTM

PR #811, opened 2023-07-13
implements ServerMock.getRegistry

NeumimTo:
# Description
Partially implements Server.getRegistry

Registries for Structure, StructureType, TrimMaterial, TrimPattern, ConfiguredStructure are not implemented due to bukkit weird implementation, ill leave these for somebody else to handle.

# Checklist
- [ x] Code follows existing style.
- [ x] Unit tests added (if applicable).

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/RegistryMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/RegistryMock.java
@@ -1,0 +1,81 @@
+package be.seeseemelk.mockbukkit;
+
+import io.papermc.paper.world.structure.ConfiguredStructure;
+import org.bukkit.Keyed;
+import org.bukkit.NamespacedKey;
+import org.bukkit.Registry;
+import org.bukkit.generator.structure.Structure;
+import org.bukkit.generator.structure.StructureType;
+import org.bukkit.inventory.meta.trim.TrimMaterial;
+import org.bukkit.inventory.meta.trim.TrimPattern;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.ParameterizedType;
+import java.util.Iterator;
+import java.util.stream.Stream;
+
+public class RegistryMock
+{
+
+	public static <T extends Keyed> Registry<?> createRegistry(Class<T> tClass)
+	{
+
+		if (tClass == Structure.class
+				|| tClass == StructureType.class
+				|| tClass == TrimMaterial.class
+				|| tClass == TrimPattern.class
+				|| tClass == ConfiguredStructure.class)
+		{
+			return new Registry<>()
+			{
+				@Override
+				public @Nullable Keyed get(@NotNull NamespacedKey key)
+				{
+					throw new UnimplementedOperationException("Registry for type " + tClass + " not implemented");
+				}
+
+				@NotNull
+				@Override
+				public Iterator<Keyed> iterator()
+				{
+					throw new UnimplementedOperationException("Registry for type " + tClass + " not implemented");
+				}
+			};
+		}
+
+
+		return Stream.of(Registry.class.getDeclaredFields())
+				.filter(a -> Registry.class.isAssignableFrom(a.getType()))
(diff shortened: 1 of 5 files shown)


src/main/java/be/seeseemelk/mockbukkit/RegistryMock.java
```diff
@@ -0,0 +1,81 @@
...
+				}
+
+				@NotNull
+				@Override
+				public Iterator<Keyed> iterator()
+				{
+					throw new UnimplementedOperationException("Registry for type " + tClass + " not implemented");
+				}
+			};
+		}
```
Insprill, line 52: This isn't necessary, as an UnimplementedOperationException will be thrown at the end of the following stream. It's better to throw the exception when trying to create the registry rather than stubbing it.

src/main/java/be/seeseemelk/mockbukkit/RegistryMock.java
```diff
@@ -0,0 +1,81 @@
...
+
+	private static Registry<?> getValue(Field a)
+	{
+		try
+		{
+			return (Registry<?>) a.get(null);
+		}
+		catch (IllegalAccessException e)
+		{
+			throw new RuntimeException(e);
```
Insprill: Please find a more specific exception to throw or create a new one extending from RuntimeException.

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -2126,8 +2128,11 @@ public StructureManager getStructureManager()
 	@Override
 	public @Nullable <T extends Keyed> Registry<T> getRegistry(@NotNull Class<T> tClass)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		if (!registry.containsKey(tClass))
+		{
+			registry.put(tClass, RegistryMock.createRegistry(tClass));
+		}
+		return (Registry<T>) registry.get(tClass);
```
Insprill, line 2138:
```suggestion
		return (Registry<T>) registry.computeIfAbsent(tClass, clazz -> RegistryMock.createRegistry(clazz));

**Review by Insprill — CHANGES_REQUESTED** [review]
Please add more unit tests as well.

File: src/main/java/be/seeseemelk/mockbukkit/RegistryMock.java (hunk 1 above)
Comment (NeumimTo) [reply src/main/java/be/seeseemelk/mockbukkit/RegistryMock.java:52]: It is necessary.

Within the class Registry theres a circular reference

```
Registry<Structure> STRUCTURE = Bukkit.getRegistry(Structure.class);
```
Registry class initializes -> Bukkit.getRegistry -> servermock.getregistry -> get field value -> ... 

Theres no other way around it, unless you change code in bukkit api directly.

Comment (Insprill) [reply src/main/java/be/seeseemelk/mockbukkit/RegistryMock.java:52]: I see. Can we still throw the exception here instead of returning a stubbed class?

Comment (NeumimTo) [reply src/main/java/be/seeseemelk/mockbukkit/RegistryMock.java:52]: No, the whole mock registry initialization would still fail.

File: src/main/java/be/seeseemelk/mockbukkit/ServerMock.java (hunk 3 above)
Comment (NeumimTo) [reply src/main/java/be/seeseemelk/mockbukkit/ServerMock.java:2138]: Not possible to change. Once this method is called due to how Registry class is made all static fields are in fact initialized at once. result would end with concurrent mod ex.

File: src/main/java/be/seeseemelk/mockbukkit/RegistryMock.java (hunk 1 above)
Comment (Insprill) [reply src/main/java/be/seeseemelk/mockbukkit/RegistryMock.java:52]: Oh, I see. I misunderstood your first explanation.

File: src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java [hunk 4]
```diff
@@ -1391,6 +1416,33 @@ void testSetMotd()
 		assertEquals("Test", server.getMotd());
 	}
 
+	@Test
+	void getRegistry() {
```
Insprill: This should be split up into two tests, one for implemented types and one for non-implemented types. A parameterized test can also be used to reduce duplication.

Insprill (changes requested):
Please fix the code smells from SonarCloud as well.

Insprill (approved):
LGTM. Thank you!

PR #817, opened 2023-07-13
Add MockBukkitException

Insprill:
# Description
Adds a new type of exception, `MockBukkitException`. This can be used as a general exception when a more specific `RuntimeException` isn't available.

This also removes the `ReflectionAccessException` implemented in #811 since it's located in the wrong class, and the new exception will do just fine.

# Checklist


thelooter:
Not really the biggest fan of just replacing one Generalized Exception with another one without adding anything to it really. I would either go for static Instances of the Exception in the Mockbukkit exception for different cases or just implement the Special Exceptions on top of RuntimeException

PR #823, opened 2023-07-15
Fix runTaskLaterAsync and runTaskTimerAsync not clamping delay

Insprill:
# Description
Without this, the task will never run if a delay of `0` is passed in.

Tests for the scheduler are a complete mess and should be almost entirely re-written in its own PR.

# Checklist


thelooter (approved):
LGTM

PR #828, opened 2023-07-30
Fix data race when scheduling a task from an async thread

frengor:
# Description
This fixes a data race (and a race condition) when calling `runTask`, `runTaskLater`, `runTaskTimer`, ... methods from an async thread.  
This also opens up the possibility to implement `callSyncMethod(...)` by simply calling `runTask` (I'll do another PR after this if you like the idea).

# Checklist


src/test/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMockTest.java
```diff
@@ -553,4 +553,211 @@ void runTask_AsyncConsumer() throws Exception
...
+		// Wait for thread to register the task
+		while (!completed.get())
+		{
+			scheduler.performOneTick();
+			Thread.yield();
+		}
+
+		while (!executed.get())
+		{
+			scheduler.performOneTick();
```
seeseemelk, line 598: Should we also add a Thread.yield() here?

seeseemelk:
Would it be possible to add some timeout-safety check to the timer tests? I expect that they'll always finish within a second, so a timeout of 60s seems like a okay value to prevent a broken test from eating one's CPU/hang a CI script

src/test/java/be/seeseemelk/mockbukkit/scheduler/BukkitSchedulerMockTest.java
frengor: No, because the other thread has already terminated when this cycle executes.

frengor:
Yes, I'll add it now.

frengor:
I've added the timeout checks to the tests.

thelooter (approved):

PR #831, opened 2023-08-05
fix(item-meta): Implement placeable and destroyable keys

Distractic:
# Description

For [ItemMeta](https://github.com/MockBukkit/MockBukkit/blob/v1.20/src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java), the method [getDestroyableKeys](https://github.com/MockBukkit/MockBukkit/blob/68960687f262c50b2978c9c660ed76f8d5ddacec/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java#L344), [setDestroyableKeys](https://github.com/MockBukkit/MockBukkit/blob/68960687f262c50b2978c9c660ed76f8d5ddacec/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java#L351), [getPlaceableKeys](https://github.com/MockBukkit/MockBukkit/blob/68960687f262c50b2978c9c660ed76f8d5ddacec/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java#L358C34-L358C52), [setPlaceableKeys](https://github.com/MockBukkit/MockBukkit/blob/68960687f262c50b2978c9c660ed76f8d5ddacec/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java#L365) are not implemented.

The goal of this PR is to implement these methods.

The methods `canDestroy` and `canPlaceOn` are not implemented because there are annotated as **Deprecated**

# Checklist


src/main/java/be/seeseemelk/mockbukkit/inventory/meta/ItemMetaMock.java
```diff
@@ -315,71 +331,63 @@ else if (hasDamage())
...
+		throw new UnimplementedOperationException("Deprecated method");
 	}
 
 	@Override
 	public void setCanPlaceOn(Set<Material> canPlaceOn)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		throw new UnimplementedOperationException("Deprecated method");
 	}
```
Insprill: These should be annotated with `@Deprecated`.

Insprill: When were these deprecated? If it was after 1.8, or if the replacement methods weren't available by last update of the same major version they should be implemented or stay stubs. We only fail when calling methods deprecated before 1.8, or that had replacements by the end of the major version. For example, if they were added in 1.13 and were deprecated by 1.13.2, it's ok to fail the original.

Insprill (changes requested):

Insprill (approved):
LGTM

PR #839, opened 2023-09-03
Allows setting lines as null for a sign

Thorinwasher:
# Description
Just noticed an inconsistency with paper behavior for signs. Yes lines are allowed to be set as null, although this just gives an empty line.

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
@@ -120,7 +120,7 @@ public class SignMock extends TileStateMock implements Sign
 
 	@Override
 	@Deprecated
-	public void setLine(int index, String line) throws IndexOutOfBoundsException
+	public void setLine(int index, @NotNull String line) throws IndexOutOfBoundsException
 	{
 		this.lines[index] = line != null ? line : "";
 	}
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
@@ -120,10 +120,9 @@ public class SignMock extends TileStateMock implements Sign
 
 	@Override
 	@Deprecated
-	public void setLine(int index, @NotNull String line) throws IndexOutOfBoundsException
+	public void setLine(int index, String line) throws IndexOutOfBoundsException
 	{
-		Preconditions.checkNotNull(line, "Line cannot be null!");
-		this.lines[index] = line;
+		this.lines[index] = line != null ? line : "";
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


thelooter:
This would be unexpected behaviour since the Interface Method declares this parameter explicitly as Non Null. Whats the reason for this change?

Thorinwasher:
Interesting. Yes that's true. As, when you look at the actual implementation, the line can be set as null. Looks like this (And I also haven't encountered any issues on any bukkit fork with this)
```java
public void setLine(int index, String line) throws IndexOutOfBoundsException {
        this.loadLines();
        this.lines.set(index, line != null ? net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer.legacySection().deserialize(line) : net.kyori.adventure.text.Component.empty()); // Paper
    }
```
But if it's unintended behavior, then maybe it's best to not merge this pull request.

Thorinwasher:
So, as the actual API of paper has changed to remedy this inconsistency. Should we:
1. Wait till the API changes.
2. Merge it right now, as it follows the implementation logic already.

Insprill:
We should wait for the changes to be merged into Paper in case there's any changes to it.

src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
```diff
@@ -120,10 +120,9 @@ public void line(int index, @NotNull Component line) throws IndexOutOfBoundsExce
 
 	@Override
 	@Deprecated
-	public void setLine(int index, @NotNull String line) throws IndexOutOfBoundsException
+	public void setLine(int index, String line) throws IndexOutOfBoundsException
```
thelooter: While the Implementation doesn't directly enforce Nullability here, we should keep the Annotation here, sine this would be inconsistent with upstream Spigot

thelooter (changes requested):

src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
Thorinwasher: If I'm correct (I'm not exactly sure of the changes of the paper API, if any), this will be a conflict of behavior between the spigot and the paper API; paper would not have the NotNull annotation while spigot will.

I can also confirm that the implementation in spigot allows for null sign lines.

I'm unsure on what would be best here though.

thelooter: As can be seen [here](https://jd.papermc.io/paper/1.20/org/bukkit/block/Sign.html#setLine(int,java.lang.String)), Paper still has the Annotation. The only change to the paper API is that [org.bukkit.block.Sign#line(int,net.kyori.adventure.text.Component)](https://jd.papermc.io/paper/1.20/org/bukkit/block/Sign.html#line(int,net.kyori.adventure.text.Component)) now enforces the Nullability with a Precondition

Thorinwasher: Okay, noted. Then definitely

thelooter (approved):

PR #844, opened 2023-09-06
Fix onDisable function being called twice

benwoo1110:
# Description
The `Plugin#onDisable` method is called twice. This is due to paper's setEnable(false) already calling the onDisable method, there shouldn't be a need to call it again.

image

# Checklist


Insprill (approved):
LGTM. Thank you!

PR #845, opened 2023-09-06
Implement MockServer#unloadWorld methods

benwoo1110:
# Description
Basic implementation of the unload world method. Basically just removes the world object from the world list if it exists. Multiverse requires this as we are porting over our tests to MockBukkit. Hope this can be merged soon, thanks!

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -1418,20 +1418,20 @@ public class ServerMock extends Server.Spigot implements Server
 	public boolean unloadWorld(World world, boolean save)
 	{
 		// TODO Handle save
-        if (!(world instanceof WorldMock worldMock))
-		{
-            return false;
-        }
-        if (!worldMock.getPlayers().isEmpty())
-		{
-            return false;
-        }
-        if (new WorldUnloadEvent(worldMock).callEvent())
+		if (!(world instanceof WorldMock worldMock))
 		{
 			return false;
 		}
-        return removeWorld(worldMock);
-    }
+		if (!worldMock.getPlayers().isEmpty())
+		{
+			return false;
+		}
+		if (new WorldUnloadEvent(worldMock).callEvent())
+		{
+			return false;
+		}
+		return removeWorld(worldMock);
+	}
 
 	@Override
 	public @NotNull MapViewMock createMap(@NotNull World world)
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -1418,17 +1418,18 @@ public class ServerMock extends Server.Spigot implements Server
 	public boolean unloadWorld(World world, boolean save)
 	{
 		// TODO Handle save
-        if (!(world instanceof WorldMock worldMock)) {
+        if (!(world instanceof WorldMock worldMock))
+		{
             return false;
         }
-        if (!worldMock.getPlayers().isEmpty()) {
+        if (!worldMock.getPlayers().isEmpty())
+		{
             return false;
         }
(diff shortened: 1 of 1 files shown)


src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -450,6 +450,11 @@ public void addWorld(WorldMock world)
 		worlds.add(world);
 	}
 
+	public boolean removeWorld(WorldMock world) {
```
Insprill, line 460: Does this need to be public?

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -1399,15 +1404,17 @@ public World createWorld(@NotNull WorldCreator creator)
 	@Override
 	public boolean unloadWorld(String name, boolean save)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return unloadWorld(getWorld(name), save);
 	}
 
 	@Override
 	public boolean unloadWorld(World world, boolean save)
```
Insprill, line 1435: Looking at Paper's implementation, there needs to be some logic added here.

- Overworld dimensions cannot be unloaded.
- Worlds with players cannot be unloaded.
- The `WorldUnloadEvent` should be called.

Insprill (changes requested):

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
benwoo1110: The addWorld method above was public, so i just followed that. I can change to private if that's your preference.

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
benwoo1110: The issue is uh, there isn't any default worlds loaded by mockbukkit in the first place, so there isnt a way to know which is the default overworld dimension.

Will implement the player and event call.

Also I realise.. there isn't `WorldLoadEvent` and `WorldInitEvent` being called for worldCreate method

Insprill: > there isn't any default worlds loaded by mockbukkit in the first place, so there isnt a way to know which is the default overworld dimension.

It isn't just the default overworld that it doesn't let you unload; It's any overworld world.
```java
if (handle.dimension() == net.minecraft.world.level.Level.OVERWORLD) {
    return false;
}
```

---

> Also I realise.. there isn't WorldLoadEvent and WorldInitEvent being called for worldCreate method

That'd be a bug, you can fix it in another PR if you'd like, otherwise we'll get around to it soon.

benwoo1110: hmmmm... Doing Bukkit#unloadWorld() on an actual server worlds on all worlds except the default world specific in server.properties `level-name`

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -1399,16 +1411,27 @@ public World createWorld(@NotNull WorldCreator creator)
...
+        if (!worldMock.getPlayers().isEmpty())
+		{
+            return false;
+        }
+        if (new WorldUnloadEvent(worldMock).callEvent())
+		{
+			return false;
+		}
+        return removeWorld(worldMock);
+    }
```
Insprill: Please replace the spaces with tabs

Insprill (changes requested):

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
benwoo1110: Done

Insprill (approved):
LGTM. Thank you!

PR #854, opened 2023-09-19
Implement World PVP API

thelooter:
# Description
Implements the World PVP API

# Checklist


Thorinwasher:
When I look at the source code, the default world pvp value is not necessarily always 'true'; it depends on whether pvp has been set to enabled for the server, that is in the server.properties file.

Could it be possible to store a default property in the ServerMock or something. This can later on can be included in a possible implementation which acts as replacement for the server.properties file? Or am I overthinking this?

TheBusyBiscuit (approved):
LGTM

Thorinwasher (approved):
LGTM

PR #855, opened 2023-09-20
Add internal tags

Thorinwasher:
# Description
Added a new type of tag "InternalTag" with required loading methods class resources. My main goal with this pull request is to implement the Block#isSolid() and World#getHighestBlockYAt() method, but that's for later. I'm trying to implement stuff where the solution can be reused later on for implementing similar methods such as the Block#isSolid() method

I don't know whether I've done everything the best way possible; like there will now be 3 different types of registries. Maybe it could be possible to reuse the preexisting TagRegistry or something.

# Checklist


src/main/java/be/seeseemelk/mockbukkit/tags/internal/InternalTagParser.java
```diff
@@ -0,0 +1,91 @@
...
+{
+	private static final Pattern MINECRAFT_MATERIAL = Pattern.compile("minecraft:[a-z0-9_]+");
+	private static final Pattern MINECRAFT_TAG = Pattern.compile("#minecraft:[a-z_]+");
+
+	public void insertInternalTagValues(InternalTagRegistry internalTagRegistry) throws IOException, InternalTagMisconfigurationException
+	{
+		String path = "/internal_tags/" + internalTagRegistry.name().toLowerCase() + "/";
+		for(InternalTag<?> internalTag : internalTagRegistry.getRelatedTags()){
+			String filePath = path + internalTag.getName().toLowerCase() + ".json";
+			System.out.println(filePath);
```
thelooter:
```suggestion

```
I think a debug Statement slipped through

thelooter (changes requested):

src/main/java/be/seeseemelk/mockbukkit/tags/internal/InternalTagParser.java
Thorinwasher: HAHA! I missed that XD

thelooter (approved):

Thorinwasher:
I accidentally merged in changes this pull request before it merged into the mockbukit branch, Trying to revert it

There removed the thing that needed to be removed, also added a check for if the thing has been initialized

PR #863, opened 2023-09-22
Fix Minecraft Version attribute in bug report template

JorelAli:
Numeric attributes that have decimals that end with a 0 will be truncated (e.g. 1.20 turns into 1.2). To fix this, we use string attributes instead

# Description

Fixes an issue in the bug template where "1.2" was being displayed instead of "1.20":

image

# Checklist


Insprill (approved):
Thank you!

seeseemelk (approved):

PR #866, opened 2023-09-30
Implement Lootable API for Block States

thelooter:
# Description
Implements the Lootable API for container block states.

Rebuilt on top of `minecraft/v26`: the original branch predates the
`be.seeseemelk.mockbukkit` -> `org.mockbukkit.mockbukkit` rename and targeted
`v1.21`, so it could not be rebased. Previous branch tip for reference: ee56bfbc.

Adds `LootableContainerStateMock`, a shared base implementing Paper's
`LootableBlockInventory`, and moves the seven container states whose Bukkit
interfaces require it onto that base: Barrel, Chest, Dispenser, Dropper, Hopper,
ShulkerBox and Crafter. This removes ~530 lines of duplicated
`UnimplementedOperationException` stubs.

Refills are plain data rather than a scheduled task, addressing the review
feedback that a block state is a snapshot and should not change over time. This
also matches Paper, where `setNextRefill` is a timestamp setter and refills are
only evaluated when a player opens the container. Behaviour follows
`PaperLootableInventory` and the Paper world config defaults.

`LootableStateMock` now stores the loot table it is given instead of throwing,
which also covers `DecoratedPotStateMock`. Generating loot from a table stays
out of scope; that needs the deterministic loot framework discussed in review.

# Checklist


src/main/java/be/seeseemelk/mockbukkit/block/state/LootableContainerMock.java
```diff
@@ -0,0 +1,208 @@
...
+				this.lootedPlayers.remove(player);
+			}
+		}
+		return hasLooted;
+	}
+
+	@Override
+	public boolean hasPendingRefill()
+	{
+		return nextRefill != -1;
```
Thorinwasher:
```suggestion
		return nextRefill != -1 && nextRefil > lastFilled;
```
This is what's done in paper. I don't see why we shouldn't follow that

src/main/java/be/seeseemelk/mockbukkit/block/state/LootableContainerMock.java
```diff
@@ -0,0 +1,208 @@
...
+	public long getNextRefill()
+	{
+		return this.nextRefill;
+	}
+
+	@Override
+	public long setNextRefill(long refillAt)
+	{
+		final long oldRefill = this.nextRefill;
+		this.nextRefill = refillAt;
```
Thorinwasher, line 157:
```suggestion
		if (refillAt < -1)
		{
			refillAt = -1
		}
		this.nextRefill = refillAt;
```
Something like this has to be done, or throw an error.

src/main/java/be/seeseemelk/mockbukkit/block/state/LootableContainerMock.java
```diff
@@ -0,0 +1,208 @@
...
+			public void run()
+			{
+				if (nextRefill == MockBukkit.getMock().getScheduler().getCurrentTick())
+				{
+					nextRefill = -1;
+					lastFilled = MockBukkit.getMock().getScheduler().getCurrentTick();
+				}
+
+			}
+		}.runTaskTimer(null, 1, 1);
```
Thorinwasher: Why does it need to be done 1+ tick later on every tick? (maybe I can learn something here)

src/main/java/be/seeseemelk/mockbukkit/block/state/LootableContainerMock.java
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	@Override
+	public long getNextRefill()
+	{
+		return this.nextRefill;
+	}
+
+	@Override
+	public long setNextRefill(long refillAt)
```
Insprill, line 149: Should we add a custom JavaDoc here mentioning the scheduler behavior?

Insprill (changes requested):

src/main/java/be/seeseemelk/mockbukkit/block/state/LootableContainerMock.java
thelooter: What do you mean?

Insprill: We should mention that you have to manually tick the scheduler in order to refill.

Thorinwasher:
I did some testing on when the container refills occurs. 

- Happens when a player opens a lootable inventory when `nextRefill <= System.currentTimeMillis()`
- Happens immediately if the container has not been filled before.

This is when the following conditions are filled:
- The inventory has a loot table set.
- The maximum amount of refills have not been reached

src/main/java/be/seeseemelk/mockbukkit/block/state/LootableContainerMock.java
```diff
@@ -0,0 +1,221 @@
...
+			refillAt = -1;
+		}
+		this.nextRefill = refillAt;
+
+		new BukkitRunnable()
+		{
+			@Override
+			public void run()
+			{
+				if (nextRefill == MockBukkit.getMock().getScheduler().getCurrentTick())
```
Thorinwasher, line 164: As I wrote in https://github.com/MockBukkit/MockBukkit/pull/866#issuecomment-1773794893 , the paper implementation for nextRefill is in relation to the milliseconds after a specific date, not the amount of ticks the server has performed. One possible alternative to fix this is to something like this
```suggestion
if (nextRefill <= System.currentTimeMillis() + MockBukkit.getMock().getScheduler().getCurrentTick()*50)
```
This implementation is flawed as well, as it technically requires the system to run the test under a certain time. The most reasonable approach is to take this in to relation to the current system time at the moment the property nextRefill was most recently assigned

src/main/java/be/seeseemelk/mockbukkit/block/state/LootableContainerMock.java
```diff
@@ -0,0 +1,221 @@
...
+			}
+		}.runTaskTimer(null, 0, 1);
+
+		return oldRefill;
+	}
+
+	@Override
+	public void setLootTable(LootTable table)
+	{
+		// TODO Auto-generated method stub
```
Thorinwasher, line 179: As I said in https://github.com/MockBukkit/MockBukkit/pull/866#issuecomment-1773794893 the whole functionality  of refilling lootable inventories depend on that it has a loot table. When I looked into this, loot tables have not been implemented. So should we just ignore this?

src/main/java/be/seeseemelk/mockbukkit/block/state/LootableContainerMock.java
```diff
@@ -0,0 +1,221 @@
...
+	{
+		final long oldRefill = this.nextRefill;
+
+		if (refillAt < -1)
+		{
+			refillAt = -1;
+		}
+		this.nextRefill = refillAt;
+
+		new BukkitRunnable()
```
Thorinwasher, line 159: Any logic like this should require the block state to have updated the block (BlockState#update() method) and not to act within the block state itself; a block state should not change over time, as it is a state

Thorinwasher (commented):
I did not think implementing lootable inventories would become this complicated.

Thorinwasher:
As I said [here](https://github.com/MockBukkit/MockBukkit/pull/866#discussion_r1367782573) you can't add functionality that modifies the block state over time, as it is indeed a state... But why not add a block state simulator, i.e. something that has a block state and ticks to simulate as an input, and a new simulated block state as an output.

Thorinwasher:
Also about loot tables, isn't fully copying the functionality disjoint with testing for this as loot tables are based out of randomness, which is not viable when unit testing?

We simply just have to have our own deterministic implementation for this, which means a lot of freedom on our end. I will write an pull request on randomness guidelines plus framework for mockbukkit.

Thorinwasher:
This might be easier to do now with the more data oriented approach that mojang has, we just need to read from these types of files, which are available from the server:
```json
{
  "type": "minecraft:chest",
  "pools": [
    {
      "bonus_rolls": 0.0,
      "entries": [
        {
          "type": "minecraft:item",
          "name": "minecraft:golden_apple",
          "weight": 20
        },
        {
          "type": "minecraft:item",
          "name": "minecraft:enchanted_golden_apple"
        },
        {
          "type": "minecraft:item",
          "name": "minecraft:name_tag",
          "weight": 30
        },
        {
          "type": "minecraft:item",
          "functions": [
            {
              "function": "minecraft:enchant_randomly",
              "options": "#minecraft:on_random_loot"
            }
          ],
          "name": "minecraft:book",
          "weight": 10
        },
        {
          "type": "minecraft:item",
          "name": "minecraft:iron_pickaxe",
          "weight": 5
        },
        {
          "type": "minecraft:empty",
          "weight": 5
        }
      ],
      "rolls": 1.0
    },
    ...
  ],
  "random_sequence": "minecraft:chests/abandoned_mineshaft"
}
```

PR #868, opened 2023-09-30
Add Ocelot Mock

Nirbose:
# Description

Adding the OcelotMock class
Added class OcelotMockTest
Testing the Ocelot spawn in WorldMockTest

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/OcelotMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/OcelotMock.java
@@ -10,8 +10,6 @@ import java.util.UUID;
 public class OcelotMock extends AnimalsMock implements Ocelot
 {
 
-	@Deprecated
-	private Ocelot.Type type = Ocelot.Type.WILD_OCELOT;
 	private boolean isTrusting = false;
 
 	/**
@@ -41,14 +39,14 @@ public class OcelotMock extends AnimalsMock implements Ocelot
 	@Deprecated
 	public @NotNull Type getCatType()
 	{
-		return this.type;
+		return Type.WILD_OCELOT;
 	}
 
 	@Override
 	@Deprecated
 	public void setCatType(@NotNull Type type)
 	{
-		this.type = type;
+		throw new UnsupportedOperationException("Cats are a different entity!");
 	}
 
 	@NotNull
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/OcelotMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/OcelotMock.java
@@ -1,0 +1,61 @@
+package be.seeseemelk.mockbukkit.entity;
+
+import be.seeseemelk.mockbukkit.ServerMock;
+import org.bukkit.entity.EntityType;
+import org.bukkit.entity.Ocelot;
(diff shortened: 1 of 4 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/OcelotMock.java
```diff
@@ -0,0 +1,61 @@
...
+	public @NotNull Type getCatType()
+	{
+		return this.type;
+	}
+
+	@Override
+	@Deprecated
+	public void setCatType(@NotNull Type type)
+	{
+		this.type = type;
```
TheBusyBiscuit: This should rather be
```suggestion
		 throw new UnsupportedOperationException("Cats are a different entity!");
```

src/main/java/be/seeseemelk/mockbukkit/entity/OcelotMock.java
```diff
@@ -0,0 +1,61 @@
...
+	public void setTrusting(boolean trust)
+	{
+		this.isTrusting = trust;
+	}
+
+	@Override
+	@Deprecated
+	public @NotNull Type getCatType()
+	{
+		return this.type;
```
TheBusyBiscuit:
```suggestion
		return Type.WILD_OCELOT;
```

TheBusyBiscuit (dismissed):

src/main/java/be/seeseemelk/mockbukkit/entity/OcelotMock.java
Nirbose: But why put an unsupported operation, because then people who are going to test with this outdated method won't be able to do so?

Thorinwasher: This is exactly the behavior observed in paper, and as MockBukkit is meant to reflect paper, it should fail as well with the same exception.

Nirbose: Oh dear, I hadn't seen that, I'll change it.

Nirbose: Done in the commit : [d3cea18](https://github.com/MockBukkit/MockBukkit/pull/868/commits/d3cea1842cdd7e5f07992e3474a02ede3eae4438)

src/main/java/be/seeseemelk/mockbukkit/entity/OcelotMock.java
Nirbose: Done in the commit : [d3cea18](https://github.com/MockBukkit/MockBukkit/pull/868/commits/d3cea1842cdd7e5f07992e3474a02ede3eae4438)

src/main/java/be/seeseemelk/mockbukkit/entity/OcelotMock.java
Thorinwasher: Cool. Hmm could you modify the unit tests as well for this? Otherwise, looks good to me

Nirbose: This was done here 8669a4539fe1e314411dced3ae4c51961e1a95ba

thelooter (approved):

Issue #825, opened 2023-07-19
Implement Entities


thelooter:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

This is a List of Entities that needs to be implemented

- [X] AcaciaBoat
- [X] AcaciaChestBoat
- [X] Allay
- [X] AreaEffectCloud
- [X] Armadillo
- [X] ArmorStand
- [X] Arrow
- [X] Axolotl
- [X] BambooChestRaft
- [X] BambooRaft
- [X] Bat
- [X] Bee
- [X] BirchBoat
- [X] BirchChestBoat
- [X] Blaze
- [X] BlockDisplay
- [X] Bogged
- [X] Breeze
- [X] BreezeWindCharge
- [X] Camel
- [X] Cat
- [X] CaveSpider
- [X] CherryBoat
- [X] CherryChestBoat
- [X] Chicken
- [X] Cod
- [X] CommandMinecart
- [X] Cow
- [X] Creaking
- [X] CreakingTransient
- [X] Creeper
- [X] DarkOakBoat
- [X] DarkOakChestBoat
- [X] Dolphin
- [X] Donkey
- [X] DragonFireball
- [X] Drowned
- [X] Egg
- [X] ElderGuardian
- [X] EnderCrystal
- [X] EnderDragon
- [X] Enderman
- [X] Endermite
- [X] EnderPearl
- [X] EnderSignal
- [X] Evoker
- [X] EvokerFangs
- [X] ExperienceOrb
- [X] ExplosiveMinecart
- [X] FallingBlock
- [X] Firework
- [X] FishHook
- [X] Fox
- [X] Frog
- [X] Ghast
- [X] Giant
- [X] GlowItemFrame
- [X] GlowSquid
- [X] Goat
- [X] Guardian
- [X] HappyGhast
- [X] Hoglin
- [X] HopperMinecart
- [X] Horse
- [X] Husk
- [X] Illusioner
- [X] Interaction
- [X] IronGolem
- [X] Item
- [X] ItemDisplay
- [X] ItemFrame
- [X] JungleBoat
- [X] JungleChestBoat
- [X] LargeFireball
- [X] LeashHitch
- [x] LightningStrike
- [X] Llama
- [X] LlamaSpit
- [X] MagmaCube
- [X] MangroveBoat
- [X] MangroveChestBoat
- [X] Marker
- [X] Mule
- [X] MushroomCow
- [X] OakBoat
[...]

PR #871, opened 2023-10-02
Add slime mock

Nirbose:
# Description
Add SlimeMock class
Add SlimeMockTest class
Update WorldMock class
Update WorldMockTest

Progresses #825 

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/SlimeMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/SlimeMock.java
@@ -1,0 +1,65 @@
+package be.seeseemelk.mockbukkit.entity;
+
+import be.seeseemelk.mockbukkit.ServerMock;
+import org.bukkit.entity.EntityType;
+import org.bukkit.entity.Slime;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.UUID;
+
+public class SlimeMock extends MobMock implements Slime
+{
+
+	private int size = 1;
+	private boolean canWander = true;
+
+	/**
+	 * Constructs a new {@link SlimeMock} on the provided {@link ServerMock} with a specified {@link UUID}.
+	 *
+	 * @param server The server to create the entity on.
+	 * @param uuid   The UUID of the entity.
+	 */
+	public SlimeMock(@NotNull ServerMock server, @NotNull UUID uuid)
+	{
+		super(server, uuid);
+	}
+
+	@Override
+	public int getSize()
+	{
+		return this.size;
+	}
+
+	@Override
+	public void setSize(int sz)
+	{
+		if (sz < 1)
+		{
+			throw new IllegalArgumentException("Size cannot be less than 1");
+		}
+		if (sz > 127)
+		{
+			throw new IllegalArgumentException("Size cannot be greater than 127");
+		}
+		this.size = sz;
+	}
+
+	@Override
+	public boolean canWander()
+	{
+		return this.canWander;
+	}
+
+	@Override
+	public void setWander(boolean canWander)
+	{
+		this.canWander = canWander;
+	}
+
+	@Override
+	public @NotNull EntityType getType()
+	{
+		return EntityType.SLIME;
+	}
+
+}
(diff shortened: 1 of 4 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/SlimeMock.java
```diff
@@ -0,0 +1,65 @@
...
+	{
+		if (sz < 1)
+		{
+			throw new IllegalArgumentException("Size cannot be less than 1");
+		}
+		if (sz > 127)
+		{
+			throw new IllegalArgumentException("Size cannot be greater than 127");
+		}
+		this.size = sz;
```
Thorinwasher: By some reason the health should also be set to the max health whenever the setSize operation is made (if the entity is alive)

Thorinwasher: The max health should also be set to size*size

Nirbose: I didn't quite understand, the function only allows you to set the size of the slime, unless it also affects its health?

Thorinwasher: Yes, it also affects the health, look at the net.minecraft.world.entity.monster.Slime#setSize(int, boolean) method

src/main/java/be/seeseemelk/mockbukkit/entity/SlimeMock.java
```diff
@@ -0,0 +1,65 @@
...
+	public void setSize(int sz)
+	{
+		if (sz < 1)
+		{
+			throw new IllegalArgumentException("Size cannot be less than 1");
+		}
+		if (sz > 127)
+		{
+			throw new IllegalArgumentException("Size cannot be greater than 127");
+		}
```
thelooter: Technically this doesnt throw in Paper but clamps it to those limits. We should consider keeping to the Paper Implementation here

src/main/java/be/seeseemelk/mockbukkit/entity/SlimeMock.java
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	@Override
+	public int getSize()
+	{
+		return this.size;
+	}
+
+	@Override
+	public void setSize(int sz)
```
thelooter:
```suggestion
	public void setSize(int size)
```
Please use descriptive Parameter names and try not to abbreviate

thelooter (changes requested):

src/main/java/be/seeseemelk/mockbukkit/entity/SlimeMock.java
Insprill: I'd prefer this impl. You can't set it outside that range; if you try to, that should be considered a bug. We should have stricter tolerances than Paper since we're a testing framework, not an actual server.

src/main/java/be/seeseemelk/mockbukkit/entity/SlimeMock.java
Nirbose: Done in commit 4efdeb1 , I leave the exeptions related to this conversation: https://github.com/MockBukkit/MockBukkit/pull/871#discussion_r1344178260

src/main/java/be/seeseemelk/mockbukkit/entity/SlimeMock.java
Nirbose: Commit 46013cb

src/main/java/be/seeseemelk/mockbukkit/entity/SlimeMock.java
thelooter: While I agree that we should fail things that don't have an effect in Paper, I think this still should not fail, since it actually performs a certain action. We should still throw a warning, but I think we should emulate Papers behavior here

Thorinwasher (approved):
Looks good to me

src/main/java/be/seeseemelk/mockbukkit/entity/SlimeMock.java
Thorinwasher: When I think about this, I think I agree on Insprill in this circumstance. I also think it's easier for us developers if we have a clear paradigm for this. I think we need to agree to either fully replicate the behavior of Paper, and throw warnings whenever the input is wrong, or to always be more stringent on user input than paper and throw errors if the input is wrong.

Nirbose: I also think it's better to be safe with a stricter tolerance than Papermc.

thelooter (approved):

PR #874, opened 2023-10-05
Deep clones BlockData + Change test for BlockData#matches()

Thorinwasher:
# Description
Previously when copying any block data, the copied BlockData has referenced to the same map which stores the data, this in essence gives odd behavior such as changing the value on the copied data modifies the original data (vise versa), see test BlockDataMock#clone_isCloneChangeInValue(). This becomes especially an issue when dealing with BlockStates as the BlockState should be a state, and not have objects that reference to BlockData in a block or another blockstate

What this pull request does is to change the reference in the clone to a new map with the same values as the object being cloned.

I also modified a unit test for BlockData#matches(), this broke with this change and when testing that test in a real instance, the proven behavior was the opposite of what was tested previously.

# Checklist


Insprill:
Is this the same behavior as Paper? I left it like this on purpose since last I checked it was how Paper handled it, could be wrong though.

Thorinwasher:
I made a simple testplugin for this:
```java
        WallSign wallSign = (WallSign) block.getBlockData();
        wallSign.setFacing(BlockFace.EAST);
        WallSign wallSignCopy = (WallSign) wallSign.clone();
        wallSignCopy.setFacing(BlockFace.NORTH);
        this.getLogger().log(Level.INFO, wallSignCopy.getFacing().toString());
        this.getLogger().log(Level.INFO, wallSign.getFacing().toString());
```
Output:
```
[20:22:35 INFO]: [TestPlugin] NORTH
[20:22:35 INFO]: [TestPlugin] EAST
```
I have no clue what happens in paper, I agree it looks like this would not be the case, but the testplugin disagrees with that.

Thorinwasher:
Here's an update on this on the code end of things; Everytime the value is set this following code is run:
```java
public <T extends Comparable<T>, V extends T> void set(Property<T> ibs, V v) {
        // Straight integer or boolean setter
        this.parsedStates = null;
        this.state = this.state.setValue(ibs, v);
}
```
Which means that whenever the BlockData is changed, it changes the reference to the field `state`. Which is what holds all the data. This reference does not change whenever an instance of a BlockData is cloned. But the exact same behavior can be seen as if it was cloned, as the only way to modify a BlockData instance is through the BlockData#set method.

The way paper achieves this is probably somewhat more efficient than deep cloning the instance (as I understand it, all possible nms BlockStates is stored in a table or something like that), but I believe  the performance gain in relation to complexity is not worth it. Therefore let's just deep clone everything.

thelooter (approved):

PR #878, opened 2023-10-11
Fix BlockState#equals

SplotyCode:
# Description
Bukkit itself had the check because it later compares BlockPos (not world bound) where as we compare Location (world bound)

# Checklist


Insprill (approved):
LGTM

PR #881, opened 2023-10-12
Fix AsyncPlayerPreLoginEvent and PlayerLoginEvent to respect login result

Paroxial:
# Description

`AsyncPlayerPreLoginEvent` and `PlayerLoginEvent` will now respect the login result set by listeners. If login events are disallowed, players will be disconnected and the `PlayerConnectionCloseEvent` will be called. The motivation behind this was that, when writing tests with MockBukkit, I expected the `PlayerJoinEvent` to not be called after I had already cancelled the `PlayerLoginEvent` (this matches Spigot behavior).

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -298,6 +298,18 @@ public 	class ServerMock extends Server.Spigot implements Server
 		try
 		{
 			conditionLatch.await();
+			if (preLoginEvent.getLoginResult() != AsyncPlayerPreLoginEvent.Result.ALLOWED)
+			{
+				PlayerConnectionCloseEvent playerConnectionCloseEvent =
+						new PlayerConnectionCloseEvent(player.getUniqueId(),
+								player.getName(),
+								player.getAddress().getAddress(),
+								false);
+
+				getPluginManager().callEvent(playerConnectionCloseEvent);
+				playerList.disconnectPlayer(player);
+				return;
+			}
 		}
 		catch (InterruptedException e)
 		{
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -298,6 +298,18 @@ public void addPlayer(@NotNull PlayerMock player)
...
 			conditionLatch.await();
+			if (preLoginEvent.getLoginResult() != AsyncPlayerPreLoginEvent.Result.ALLOWED)
+			{
+				PlayerConnectionCloseEvent playerConnectionCloseEvent =
+						new PlayerConnectionCloseEvent(player.getUniqueId(),
+								player.getName(),
+								player.getAddress().getAddress(),
+								false);
+
+				getPluginManager().callEvent(playerConnectionCloseEvent);
```
Insprill, line 309: Should this be called async?

src/test/java/be/seeseemelk/mockbukkit/TestPlugin.java
```diff
@@ -126,6 +130,24 @@ public void onAsyncChat(AsyncPlayerChatEvent event)
...
+
+	@EventHandler
+	public void onLogin(PlayerLoginEvent event)
+	{
+		if (denyLoginEvent)
+		{
+			event.setResult(PlayerLoginEvent.Result.KICK_OTHER);
+		}
+	}
+
```
Insprill: I don't think this needs to be added, as it's a bit of a niche use case. This is just for common stuff that people will test.

Insprill (dismissed):

thelooter:
@Paroxial hows the progress on this?

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
thelooter: Is there a reason why it shouldn't

Thorinwasher: From documentation it says that the event may be both synchronous or asynchronous

thelooter:
@Thorinwasher are those two tests the same as Sonar reports?

Thorinwasher:
@thelooter They seem to be duplicates, so I removed one

thelooter (approved):

PR #886, opened 2023-10-14
Implement ServerMock#getCurrentTick

Insprill:
# Description
Implements `ServerMock#getCurrentTick`.

Normally this returns `MinecraftServer.currentTick` which increments each tick, but since our only concept of ticking is the scheduler, we just return its current tick. 

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
@@ -2422,8 +2422,7 @@ public class ServerMock extends Server.Spigot implements Server
 	@Override
 	public int getCurrentTick()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return (int) getScheduler().getCurrentTick();
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


thelooter:
> Normally this returns `MinecraftServer.currentTick` which increments each tick, but since our only concept of ticking is the scheduler, we just return its current tick.

What about introducing a Tick System? This would give us a cleaner way to start implementing Things like Physics too, and gives us a cleaner separation from the scheduler.

Insprill:
I agree we should implement a tick system, but we also must be careful to draw the line somewhere for what we simulate to avoid becoming an entire server or having unexpected behavior because some things are simulated and others aren't. That's definitely something for MB 4 though.

Thorinwasher (approved):
Looks good to me

thelooter:
> I agree we should implement a tick system, but we also must be careful to draw the line somewhere for what we simulate to avoid becoming an entire server or having unexpected behavior because some things are simulated and others aren't. That's definitely something for MB 4 though.

Can you make an issue for this?

thelooter (approved):

PR #887, opened 2023-10-14
ParrotMock implemented with tests

Farlimb:
Adds both the Parrot entity

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/ParrotMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/ParrotMock.java
@@ -47,6 +47,9 @@ public class ParrotMock extends AnimalsMock implements Parrot
 		return this.isDancing;
 	}
 
+	public void setDancing(boolean isDancing){
+		this.isDancing = isDancing;
+	}
 	@Override
 	public boolean isTamed()
 	{
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/ParrotMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/ParrotMock.java
@@ -12,11 +12,13 @@ import java.util.UUID;
 
 public class ParrotMock extends AnimalsMock implements Parrot
 {
+
 	private @NotNull Parrot.Variant variant = Variant.RED;
 	private boolean isDancing = false;
 	private boolean isSitting = false;
 	private boolean isTamed = false;
 	private AnimalTamer animalTamer = null;
+
 	/**
 	 * Constructs a new {@link ParrotMock} on the provided {@link ServerMock} with a specified {@link UUID}.
 	 *
@@ -47,9 +49,11 @@ public class ParrotMock extends AnimalsMock implements Parrot
 		return this.isDancing;
 	}
 
-	public void setDancing(boolean isDancing){
+	public void setDancing(boolean isDancing)
+	{
 		this.isDancing = isDancing;
 	}
+
 	@Override
 	public boolean isTamed()
 	{
@@ -97,4 +101,5 @@ public class ParrotMock extends AnimalsMock implements Parrot
 	{
 		return EntityType.PARROT;
 	}
+
 }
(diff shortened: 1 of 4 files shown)


src/test/java/be/seeseemelk/mockbukkit/entity/ParrotMockTest.java
```diff
@@ -0,0 +1,83 @@
...
+		parrot.setSitting(true);
+		assertTrue(parrot.isSitting());
+	}
+	@Test
+	public void testSetOwner()
+	{
+		AnimalTamer owner = server.addPlayer();
+		parrot.setOwner(owner);
+		assertEquals(owner, parrot.getOwner());
+	}
```
Nirbose: remove all `public` to stay in the same style as the other unit test files and your `testGetType()`.

src/main/java/be/seeseemelk/mockbukkit/entity/ParrotMock.java
```diff
@@ -0,0 +1,97 @@
...
+	public void setVariant(@NotNull Variant variant)
+	{
+		Preconditions.checkNotNull(variant, "Variant cannot be null");
+		this.variant = variant;
+	}
+
+	@Override
+	public boolean isDancing()
+	{
+		return this.isDancing;
```
Thorinwasher: This is always going to return false. You should probably add an access method for the isDancing field.

Farlimb: you mean a setter method? or a toogle method would be good one?

thelooter: Add a `setDancing(boolean)` method, that sets the `isDancing`field

src/main/java/be/seeseemelk/mockbukkit/entity/ParrotMock.java
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	public void setDancing(boolean isDancing)
+	{
+		this.isDancing = isDancing;
+	}
```
Insprill: `#isDancing` only returns true when there's a jukebox playing a record nearby. I don't believe we have jukeboxes implemented yet, so this shouldn't be implemented until we do.

Insprill (changes requested):

src/main/java/be/seeseemelk/mockbukkit/entity/ParrotMock.java
Nirbose: The juckbox is well implemented, but there are 3 methods that aren't "done", including `startPlaying`, and I wonder if it isn't this method that sets the peroquer's dancing to `true` (I haven't checked on papermc yet).

Farlimb: Should I add TO DO comment in setDancing method and make it throw new UnimplementedOperationException()?

Insprill: setDancing should be removed, and isDancing should throw, along with a comment about why it isn't implemented.

Nirbose:
it would be nice to respect the code style of commit 79914cdf6bac6383ed2d521e7e8e21cec96bbd7d in the `ParotMockTest` file no?

src/main/java/be/seeseemelk/mockbukkit/entity/ParrotMock.java
```diff
@@ -0,0 +1,102 @@
...
+
+	@Override
+	public void setTamed(boolean tame)
+	{
+		this.isTamed = tame;
+	}
+
+	@Override
+	public @Nullable UUID getOwnerUniqueId()
```
thelooter, line 67: Please add a Test for this Method

thelooter (changes requested):

thelooter (approved):
LGTM

Farlimb (commented):
done

Insprill (approved):
LGTM; Thank you!

Issue #900, opened 2023-11-11
Remove use of Deprecated Constructor


thelooter:
https://github.com/MockBukkit/MockBukkit/blob/2e865fa5c78782af64763d2dc3763b2b65f49dbb/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java#L362

The used Constructor for `BlockDamageEvent`is annotated as Deprecated for Removal and and DoNotUse. Therefore we should replace this Constructor

JonnyMc94:
Hi, I would like to work on this issue. Could you assign it to me?

thelooter:
Sure, done

JonnyMc94:
@thelooter I can't seem to find a depreciation marker or a discussion about possible deprecation of this constructor anywhere. Could you point me in the direction of some info on it? 

This file only has one constructor 
https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/event/block/BlockDamageEvent.java

thelooter:
We're building on Paper, not Bukkit (which has been unmaintained since 2014), and they have deprecated this constructor as you can see [here](https://github.com/PaperMC/Paper/blob/f186318a91cbd3b2a2259d39cb88576989a496b8/patches/api/0429-Expose-clicked-BlockFace-during-BlockDamageEvent.patch#L17-L22). Therefore we wanna remove this usage

PR #902, opened 2023-11-13
Fix constructor call

JonnyMc94:
# Description
Address #900 .

The constructor call for BlockDamageEvent has been updated. A small update to a Javadoc has also been made. 

# Checklist


thelooter:
This currently doesn't compile, since the `Block#getFace(Block)` Method expects an argument. I'm not super shure if we need the Face here, or if we need to adapt the method / add an Overload to account for Blockfaces

JonnyMc94:
> This currently doesn't compile, since the `Block#getFace(Block)` Method expects an argument. I'm not super shure if we need the Face here, or if we need to adapt the method / add an Overload to account for Blockfaces

Could you give some guidance as to how we would call the constructor without having one of the parameters please?

thelooter:
@Insprill can you please take a look at this

Thorinwasher:
> Could you give some guidance as to how we would call the constructor without having one of the parameters please?

The method `PlayerMock#simulateBlockDamagePure()` is simply lacking enough informaation, it would need to also be given the blockface that the player has damaged the block on. To add the `BlockFace` parameter into the constructor of `BlockDamageEvent`, we would either need to:
- Default to a facing, i.e. BlockFace.EAST (or maybe better BlockFace.SELF)
- Add a duplicate method that has the facing as a parameter

4everTheOne (approved):
Looks good to me

PR #909, opened 2023-11-22
feat(world & server): Implement Spawn Limit API

thelooter:
# Description
Implements the Spawn Limits API

# Checklist


src/main/java/be/seeseemelk/mockbukkit/configuration/ServerConfiguration.java
```diff
@@ -49,6 +51,13 @@ public ServerConfiguration()
 		ticksPerSpawn.put(SpawnCategory.WATER_UNDERGROUND_CREATURE, 1);
 		ticksPerSpawn.put(SpawnCategory.WATER_ANIMAL, 1);
 		ticksPerSpawn.put(SpawnCategory.AMBIENT, 1);
+
+		spawnLimits.put(SpawnCategory.MONSTER, 70);
```
Thorinwasher, line 55: Would it not be better if MetaMiner fetched these default values (same in WorldMock)? (i.e load the values from a json file)

thelooter: From where would you pull that. Those default values are pulled from the Config File

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -3174,15 +3172,29 @@ public void setTicksPerSpawns(@NotNull SpawnCategory spawnCategory, int ticksPer
 	@Override
 	public int getSpawnLimit(@NotNull SpawnCategory spawnCategory)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Preconditions.checkArgument(spawnCategory != null, "SpawnCategory cannot be null");
+		Preconditions.checkArgument(spawnCategory != SpawnCategory.MISC,
```
Thorinwasher, line 3244: This check does not consider if a new spawn limit has been implemented in paper. It would be good if something like an UnimplementedOperationException gets thrown whenever that is the case.

src/main/java/be/seeseemelk/mockbukkit/configuration/ServerConfiguration.java
Thorinwasher: Yeah, that's true... Let me see if it's possible to get the absolute default values somehow. One stupid idea would be to always clear the config files used by the server somehow, either by making a specific environment where MetaMiner is run, or by implementing a script in MetaMiner that clears these files after the server has loaded.

Edit: If I'm not mistaken, I should be able to load the class file "/configurations/bukkit.yml" and from that read the config. This should always be the default values. What do you say, is it worth the effort?

src/main/java/be/seeseemelk/mockbukkit/ServerMock.java
```diff
@@ -2529,8 +2522,11 @@ public int getTicksPerSpawns(@NotNull SpawnCategory spawnCategory)
 	@Override
 	public int getSpawnLimit(@NotNull SpawnCategory spawnCategory)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Preconditions.checkArgument(spawnCategory != null, "SpawnCategory cannot be null");
+		Preconditions.checkArgument(spawnCategory != SpawnCategory.MISC,
+				"SpawnCategory.%s are not supported", spawnCategory);
+
+		return this.serverConfiguration.getSpawnLimits().getOrDefault(spawnCategory, -1);
```
Thorinwasher, line 2474:
```suggestion
		if(!this.serverConfiguration.getSpawnLimits().containsKey(spawnCategory)){
			throw new UnimplementedOperationException();
		}
		return this.serverConfiguration.getSpawnLimits().getOrDefault(spawnCategory, -1);
```
Update for comment https://github.com/MockBukkit/MockBukkit/pull/909/files#r1410812763

src/main/java/be/seeseemelk/mockbukkit/configuration/ServerConfiguration.java
Thorinwasher: This is mostly me overthinking stuff, let's not

seeseemelk (approved):

Thorinwasher (approved):
Looks good to me

PR #917, opened 2023-11-29
feat(world): Implement RespawnAnchor API

thelooter:
# Description
Implements both `isRespawnAnchorWorks()` and `doesRespawnAnchorWork()`. Ideally this would be registry backed, so it would stay in sync with potential new World Types, but since thats not something we have at the moment, this is the best we got.

# Checklist 


Insprill (approved):
LGTM

PR #925, opened 2023-11-30
feat(entities): Implement Rest of Chicken Methods

thelooter:
# Description

# Checklist


src/main/java/be/seeseemelk/mockbukkit/entity/ChickenMock.java
```diff
@@ -46,30 +51,25 @@ public boolean isBreedItem(@NotNull ItemStack stack)
...
+		return this.isChickenJockey;
 	}
 
 	@Override
 	public void setIsChickenJockey(boolean isChickenJockey)
 	{
-		//TODO: Auto-generated method stub
-		throw new UnimplementedOperationException();
-
+		this.isChickenJockey = isChickenJockey;
```
Insprill, line 60: How does this work in Paper? Should we be spawning the baby zombie and setting it to ride the chicken?

Insprill (changes requested):

src/main/java/be/seeseemelk/mockbukkit/entity/ChickenMock.java
thelooter: From what I can tell it's just a variable thats set. I don't see the spawning of the baby zombie anywhere

Thorinwasher: Yep, it's just a boolean that is set to indicate things for the chicken, for example not to lay eggs, the xp rewards, and whether to despawn when player is too far away.

Insprill (approved):
LGTM

PR #927, opened 2023-12-05
feat(entities): Implement LeashHitch Entity

thelooter:
# Description
feat(entities): Implement LeashHitch Entity

# Checklist


src/main/java/be/seeseemelk/mockbukkit/entity/LeashHitchMock.java
```diff
@@ -0,0 +1,30 @@
...
+	{
+		super(server, uuid);
+	}
+
+	@Override
+	public @NotNull EntityType getType()
+	{
+		return EntityType.LEASH_HITCH;
+	}
+
```
Thorinwasher, line 31:
```suggestion
    @Override
    public BlockFace getFacing()
     {
        return BlockFace.SELF;
    }
    
    @Override
    public boolean setFacingDirection(BlockFace face, boolean force) 
    {
        Preconditions.checkNotNull(face);
        Preconditions.checkArgument(face == BlockFace.SELF, "%s is not a valid facing direction", face);

        return true);
    }
```
Warning for spaces instead of tabs

Might be something more needed to be done with the return value

Thorinwasher (approved):
Looks good to me

Issue #928, opened 2023-12-07
Internal resources for plugins are not isolated


Thorinwasher:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Are you using the latest version of MockBukkit?

- [X] I am using the latest version of MockBukkit.

### Minecraft Version

1.20.2

### Describe the bug

One common internal file for plugins is the `config.yml`. Whenever the default config is saved, this internal file is used. What I understand from class loading, these internal files are linked with a package name and a classloader. The issue that arrives here is that when multiple plugins are loaded which share the same path for an internal file. Then currently for MockBukkit only one of these instances is possible to get whenever that internal file is fetched. So to say, one of the plugins will get the internal file from the other plugin.

### Reproducible Test

```java
@Test
void aTest()
{
    JavaPlugin plugin1 = MockBukkit.load(Plugin1);
    JavaPlugin plugin2 = MockBukkit.load(Plugin2);
    assertDoesNotHaveSameContent(plugin1.class.getResourceAsStream("config.yml"),
        plugin2.class.getResourceAsStream("config.yml")))
}
```

### Anything else?

One solution would be to have the plugins load from completely separate class loaders, this should isolate the resource, therefore avoiding the issue

I have no clue how to do this though. As everything is already loaded by the SystemClassLoader (or whatever it's called) to run the test, i.e. before triggering the MockBukkit#load() command.

TheDarkSword:
https://github.com/MockBukkit/MockBukkit/pull/929
I fixed here

Thorinwasher:
- [ ] Fixed for MockBukkit#load(...)
- [x] Fixed for MockBukkit#loadJar(...)

PR #929, opened 2023-12-09
Create URLClassLoader for plugins loaded from jar file

TheDarkSword:
# Description
This commit addresses the issue where, if you load more than one plugin, the files in the resources (such as plugin.yml and config.yml) are always the same as those of the application undergoing testing. This way, external JARs are loaded using a URLClassLoader, and when a call to #getResourceAsStream is made, the resource is first searched for internally within the JAR and then in the parent (opposite to the default behavior).

Issue reference: https://github.com/MockBukkit/MockBukkit/issues/928

# Checklist


Thorinwasher:
Sorry for not responding sooner. I mainly just don't consider myself qualified enough for this.

Have you tested the changes, does it work?

Is it possible to remove reduntant code in MockBukkitConfiguredPluginClassLoader, which is used to load jar files?

Except for this, I don's see any issues. Again sorry for neglecting this PR

TheDarkSword:
> Sorry for not responding sooner. I mainly just don't consider myself qualified enough for this.
> 
> Have you tested the changes, does it work?
> 
> Is it possible to remove reduntant code in MockBukkitConfiguredPluginClassLoader, which is used to load jar files?
> 
> Except for this, I don's see any issues. Again sorry for neglecting this PR

Oh, no problem!
So, basically, the issue itself is that by default the ClassLoader first checks if the parent ClassLoader (in our case, it's our plugin) already has that file/class loaded, for example, the config.yml, and then it obtains the one already loaded.
Obviously, that's not the correct file. What I did is simply reverse the normal process, first looking for the resource within the jar itself and then trying to find it by calling the parent ClassLoader.
This way, I managed to test and everything works as it should.
I'll remove the code that is no longer needed from the MockBukkitConfiguredPluginClassLoader class, as you instructed. I'll do that as soon as possible!

thelooter:
> > Sorry for not responding sooner. I mainly just don't consider myself qualified enough for this.
> > Have you tested the changes, does it work?
> > Is it possible to remove reduntant code in MockBukkitConfiguredPluginClassLoader, which is used to load jar files?
> > Except for this, I don's see any issues. Again sorry for neglecting this PR
> 
> Oh, no problem! So, basically, the issue itself is that by default the ClassLoader first checks if the parent ClassLoader (in our case, it's our plugin) already has that file/class loaded, for example, the config.yml, and then it obtains the one already loaded. Obviously, that's not the correct file. What I did is simply reverse the normal process, first looking for the resource within the jar itself and then trying to find it by calling the parent ClassLoader. This way, I managed to test and everything works as it should. I'll remove the code that is no longer needed from the MockBukkitConfiguredPluginClassLoader class, as you instructed. I'll do that as soon as possible!

Would it be possible to create a Method that loads an arbitrary specified config file too? This would allow us to create Integration Tests with custom config files, something thats required by many plugins

TheDarkSword:
@thelooter In the coming days, I will try to find a way. However, the simplest way to do it currently is to modify the config and load it directly into the JAR file.

thelooter:
@TheDarkSword we need some Unit Tests to get this merged. Can you add those?

seeseemelk:
> @TheDarkSword we need some Unit Tests to get this merged. Can you add those?

There is already an `extra/TestPlugin` directly which will be built into a jar file. You might be able to use this jar file in your test.

thelooter:
@TheDarkSword is there anything missing from you're point of view?

Thorinwasher:
As this PR is dormant, I will self assign myself this issue. I guess adding tests for this?

thelooter (approved):

PR #931, opened 2023-12-11
Update to 1.20.4

thelooter:
# Description
Update to 1.20.4

# Checklist


Thorinwasher (approved):

Thorinwasher (dismissed):
Approved this without noticing build failures, let's see if this revokes that review.

andrewkm:
Thank you for this! :)

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
```diff
@@ -1461,23 +1461,6 @@ void getHighestBlockYAt_Water()
 		assertEquals(20, worldMock.getHighestBlockYAt(2, 3, HeightMap.MOTION_BLOCKING));
 	}
 
-	@Test
-	void getHighestBlockYAt_Waterlogged()
```
Insprill, line 1465: Why was this test removed?

src/main/java/be/seeseemelk/mockbukkit/potion/MockPotionEffectType.java
```diff
@@ -39,13 +58,18 @@ public class MockPotionEffectType extends PotionEffectType
...
+	}
 
-		this.id = id;
-		this.name = name;
-		this.instant = instant;
-		this.color = color;
-		this.attributeModifiers = new EnumMap<>(Attribute.class);
+	public MockPotionEffectType(JsonObject data)
+	{
+		this(Objects.requireNonNull(NamespacedKey.fromString(data.get("key").getAsString())),
```
Insprill: Is the `Objects.requireNonNull` necessary?

src/main/java/be/seeseemelk/mockbukkit/enchantments/EnchantmentMock.java
```diff
@@ -193,14 +254,14 @@ public boolean isCursed()
...
-		this.isCursed = isCursed;
+		this.cursed = isCursed;
 	}
 
 	@Override
 	public boolean conflictsWith(@NotNull Enchantment other)
 	{
-		Preconditions.checkNotNull(other, "Enchantment cannot be null");
-		return false;
+		Preconditions.checkArgument(other != null, "others can't be null") ;
```
Insprill: Why was `checkNotNull` replaced `checkArgument`? We shouldn't copy Spigot's bad practices. If anything, we should convince upstream to use the proper method.

src/main/java/be/seeseemelk/mockbukkit/enchantments/EnchantmentMock.java
```diff
@@ -21,66 +26,122 @@ public class EnchantmentMock extends Enchantment
...
+	private final int[] maxModifiedCosts;
+	private boolean cursed;
+	private final int[] minModifiedCosts;
 	private int maxLevel;
 	private int startLevel;
 	private EnchantmentTarget itemTarget;
-	private boolean isTreasure;
-	private boolean isCursed;
+	private static final String LEVEL = "level";
+	private static final String COST = "cost";
```
Insprill: Constants should be at the top of the class.

src/test/java/be/seeseemelk/mockbukkit/ChunkSnapshotMockTest.java
```diff
@@ -29,7 +29,7 @@ class ChunkSnapshotMockTest
 	void setUp()
 	{
 		MockBukkit.mock();
-		world = new WorldMock(Material.GRASS, 0, 319, 4);
+		world = new WorldMock(Material.SHORT_GRASS, 0, 319, 4);
```
Insprill: Why `SHORT_GRASS` here? We should either use `GRASS_BLOCK` or something different entirely.

src/test/java/be/seeseemelk/mockbukkit/MockChunkDataTest.java
```diff
@@ -90,7 +90,7 @@ void test_get_out_of_bounds()
 	@Test
 	void test_neg_min_height()
 	{
-		world = new WorldMock(Material.GRASS, -60, 256, 70);
+		world = new WorldMock(Material.SHORT_GRASS, -60, 256, 70);
```
Insprill: Same here.

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
```diff
@@ -674,12 +674,12 @@ void setBlock_ChangeBlock()
 		Block block = world.getBlockAt(0, 1, 0);
 		BlockData data = block.getBlockData();
 		assertEquals(data.getMaterial(), world.getBlockData(location).getMaterial());
-		BlockDataMock mock = new BlockDataMock(Material.GRASS);
+		BlockDataMock mock = new BlockDataMock(Material.SHORT_GRASS);
```
Insprill, line 677: And here

src/test/java/be/seeseemelk/mockbukkit/tags/internal/InternalTagParserTest.java
```diff
@@ -24,12 +24,6 @@ void insertInternalTagValues_registries()
 		assertFalse(InternalTag.NON_SOLID_BLOCKS.getValues().isEmpty());
 	}
 
-	@Test
-	void checkTagged_fromTag()
```
Insprill, line 28: Was this tag removed from the game?

Insprill (changes requested):
Looks pretty good overall. Just a few notes, and the JSON files should be using tabs instead of spaces.

src/main/java/be/seeseemelk/mockbukkit/enchantments/EnchantmentMock.java
thelooter: Because we are checking an Argument. Usually checkNotNull should be used later on to check if something is not null, e.g. after trying to retrieve an object from a collection

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
Thorinwasher: Long story short, I automatically generated the data needed for isSolid, and noticed that the manually created data was invalid, therefore invalidating some tests. So to say, I could not find a waterlog-able material which already is not considered solid, which is required for that specific test

src/main/java/be/seeseemelk/mockbukkit/potion/MockPotionEffectType.java
Thorinwasher: Nope, it is not, It's getting checked in the targeted constructor later on

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
Thorinwasher: For how this test looked initially, one of the materials was Material.GRASS, and another Material.GRASS_BLOCK. By just thinking quickly about this, it seemed like substituting Material.GRASS -> Material.SHORT_GRASS was a valid choice.

src/test/java/be/seeseemelk/mockbukkit/tags/internal/InternalTagParserTest.java
Thorinwasher: Because I replaced the data with automatically generated data, which does not use tags. Test therefore became invalid

src/main/java/be/seeseemelk/mockbukkit/enchantments/EnchantmentMock.java
Insprill: From the checkNotNull Javadoc:
> Ensures that an object reference passed as a parameter to the calling method is not null.

thelooter: I still think that checkArgument is the better choice here

Insprill: Why? We're checking to see if it's null, and there's an explicit method & exception for precisely that purpose.

Thorinwasher: If we take the actual code that we are mocking into relation, it looks like this:
```java
    @Override
    public boolean conflictsWith(Enchantment other) {
        if (other instanceof EnchantmentWrapper) {
            other = ((EnchantmentWrapper) other).getEnchantment();
        }
        if (!(other instanceof CraftEnchantment)) {
            return false;
        }
        CraftEnchantment ench = (CraftEnchantment) other;
        return !this.handle.isCompatibleWith(ench.getHandle());
    }
```
It will in essence just return false.

I think I agree with @Insprill on this one, as NullPointerException is clearer, but it does not matter too much.

Insprill (approved):
LGTM 🥳

PR #933, opened 2023-12-12
Implement PlayerTextures in playerprofilemock

J3fftw1:
# Description

This PR implements getTextures and setTextures along with the PlayerTextures class.

# Checklist


src/test/java/be/seeseemelk/mockbukkit/profile/PlayerProfileMockTest.java
```diff
@@ -97,6 +100,22 @@ void setUuid()
...
 	}
 
+	@Test
+	void setTextures() throws MalformedURLException
+	{
+		UUID uuid = UUID.fromString("22815ad5-2a54-44c0-8f83-f65cfe5310f8");
+		PlayerProfileMock profile = new PlayerProfileMock("Test", uuid);
+		PlayerTexturesMock playerTextures = new PlayerTexturesMock(profile);
+		playerTextures.setSkin(new URL("http://textures.minecraft.net/texture/208bd79e7d56584e57bbeee5f4cb2172d9a56c49ea9df405002efcf213d3ce52"));
+		playerTextures.setCape(new URL("http://textures.minecraft.net/texture/afd553b39358a24edfe3b8a9a939fa5fa4faa4d9a9c3d6af8eafb377fa05c2bb"));
```
thelooter: Do these need to be web links or would a local Picture also work

J3fftw1: it shouldnt matter much since we dont validate the actual textures.
We just validate if 1 matches the other

Insprill: Could we use an easier-to-read URL, then? Make one, say, Google and one GitHub. That would make it easier to see what went wrong if the test fails

src/main/java/be/seeseemelk/mockbukkit/profile/PlayerTexturesMock.java
```diff
@@ -0,0 +1,111 @@
...
+		if (dirty) return false;
+		ProfileProperty property = getProperty();
+		return property != null;
+	}
+
+	@Nullable
+	ProfileProperty getProperty() {
+		return profile.getProperty(PROPERTY_NAME);
+	}
+
```
Thorinwasher, line 127: I saw that there's a modification to the Object#equals(Object) method, the Object#hashCode() method (and also the toString method, but that's not as important) in CraftPlayerTextures.

For the equals method, it seems from my understanding that this check compares the skin URL, the skin model, and the cape URL.  If one modifies the equals method, then one has to modify the hash code as well, such that if two objects equals each other, they will have the same hash code.

src/main/java/be/seeseemelk/mockbukkit/profile/PlayerTexturesMock.java
```diff
@@ -0,0 +1,111 @@
+package be.seeseemelk.mockbukkit.profile;
```
Thorinwasher, line 1: Could you make a simple unit test for this class?

Hmmm, at least for the equals method (if you are willing to implement that). Otherwise it's mostly setters and getters, which I myself believe does not need to be unit tested (but feel free to)

src/main/java/be/seeseemelk/mockbukkit/profile/PlayerTexturesMock.java
```diff
@@ -0,0 +1,111 @@
...
+	private URL skin;
+	private SkinModel skinModel = SkinModel.CLASSIC;
+	private URL cape;
+
+	// Dirty: Indicates a change that requires a rebuild of the property.
+	// This also indicates an invalidation of any previously present textures data that is specific to official
+	// GameProfiles, such as the property signature, timestamp, profileId, playerName, etc.: Any modifications by
+	// plugins that affect the textures property immediately invalidate all attributes that are specific to official
+	// GameProfiles (even if these modifications are later reverted).
+	private boolean dirty = false;
```
Thorinwasher, line 34: This property is always false

src/main/java/be/seeseemelk/mockbukkit/profile/PlayerTexturesMock.java
thelooter: Yes they should be unit tested.

src/main/java/be/seeseemelk/mockbukkit/profile/PlayerTexturesMock.java
```diff
@@ -0,0 +1,173 @@
...
```
Thorinwasher, line 34: I looked into the `dirty` field... By some reason it actually updates to false whenever the hashCode and equals method are called. I have no clue why who decided to do this though (Might be unwanted behavior I have no clue).

I'm honestly confused by the `dirty` field why is this needed at all? Would it not be better to directly schedule an update of this whenever there's a change in the skin/cape URL?

Also found that these methods marks `dirty` as false:
- isEmpty
- getSkin
- getSkinModel
- getCape
- getTimestamp
- toString

Thorinwasher: This CraftPlayerTextures class is such a mess for being only 300 lines

J3fftw1: Is there anything you want me to change within this PR about this code?

Thorinwasher: Looks good so far, but I think as you are already implementing `dirty`, you can do it properly i.e. mark it as `false` when the abovementioned methods are used

Thorinwasher (approved):
Looks good to me

PR #940, opened 2024-01-05
fix BlockMock.setType(Material type, boolean applyPhysics) doing nothing

Figishe:
Method argument was ignored at all


Thorinwasher (approved):
I found the commit from the one who made this heinous mistake. It's 8957adacfedee8e795a9c3d05519688a0275d20b , wow really one of the initial merged pull requests to this project.

Anyhow, thanks for noticing this small mistake! The proposed change looks good to me

PR #944, opened 2024-01-10
fix playerlist hasplayedbefore

J3fftw1:
# Description

This PR fixes the playerlist hasplayedbefore.
When setting setlastseen hasplayedbefore would never be set.
This would always return false.

# Checklist


Thorinwasher (approved):
Looks good to me

PR #946, opened 2024-01-13
Add property to identify mockbukkit environment

J3fftw1:
# Description

This PR adds a system property to identify mockbukkit environment.
This can be used for checking if we are in a unittest.
This would help solve having multiple constructors in the main class.

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
@@ -80,6 +80,8 @@ public class MockBukkit
 		Bukkit.setServer(mock);
 		mock.getLogger().setLevel(defaultLevel);
 
+		System.setProperty("mockbukkit.unit.test", "true");
+
 		return serverMockImplementation;
 	}
 
--- a/src/test/java/be/seeseemelk/mockbukkit/MockBukkitTest.java
+++ b/src/test/java/be/seeseemelk/mockbukkit/MockBukkitTest.java
@@ -80,6 +80,12 @@ class MockBukkitTest
 	}
 
 	@Test
+	void mock_MockPropertySet() {
+		MockBukkit.mock();
+		assertEquals("true", System.getProperty("mockbukkit.unit.test"));
+	}
+
+	@Test
 	void isMocked_ServerNotMocked_False()
 	{
 		assertFalse(MockBukkit.isMocked());


WalshyDev:
Particularly, this lets us validate at runtime rather than just within the test folder. In updating from MockBukkit 2 (1.18) to 3 (1.20.4) we noticed the constructor which used to be invoked (`Slimefun(JavaPluginLoader loader, PluginDescriptionFile description, File dataFolder, File file)`) no longer is and instead it's just the default no param one (`Slimefun()`). This means we can no longer rely on the constructor being used (this change was made in https://github.com/MockBukkit/MockBukkit/pull/747)

Therefore, for us to know we're within a unit test within our standard code, we need an identifier. Java properties are a nice way to do this. 
The alternatives are hacky and not something we'd like to do, we can add a new constructor and parameter to every single `MockBukkit.load` or we can check if the class loader has mockbukkit in the package name  or other similar alternatives 

We think this would be a low cost but good general improvement. 
(this is further backed by our convo in Discord - https://canary.discord.com/channels/792754410576019477/809866105275023381/1195747031440953445)

Thorinwasher:
I'm not the best qualified person for deciding whether this is a good idea. But I do believe that we should not have this behavior as it encourages running slightly modified code when testing. If this would have been a good idea, why haven't junit done anything similar?

Instead of doing that, you can also do something like this:
```java
public static final boolean IS_TEST = isRunningTest();

private static boolean isRunningTest() {
        try {
            Class.forName("org.junit.Test");
            return true;
        } catch (ClassNotFoundException e) {
            return false;
        }
}
```
I hope this solves your problems

WalshyDev:
Oh, it 100% is not good and not recommended practice however, within a plugin env it's quite difficult to cover all cases. We had to make a number of PRs to just get 1.20.4 working for example and that will continue. There's also just other plugin interaction (slimefun works with like 15-20 other plugins...) which I wish we could easily test but we cannot.

The aim for us is definitely to move off using a secondary code path and not covering specific code but it just isn't simple today.

This PR is not a requirement for us and we already have a new way (a simple class detection like you alluded to). It was more of a "we're having to do this so others may be too". Y'all are obviously the best judges for if this should go in. Feel free to close if y'all do not wish for it to land :)

Thorinwasher:
All good, this is just my opinion. I will leave this open for a while if anyone dissagrees with me in the dev team.

thelooter:
I agree with the stance brought forward by @Thorinwasher

Insprill:
I also agree with Thorinwasher. We shouldn't have an official way in the library to do this. If it needs to be done, you can use a utility method like he posted.

PR #961, opened 2024-02-07
Permissions from plugin file

### LINKED ISSUE #612: Parent permission nodes do not enable permissions

### Minecraft Version

1.19

### Describe the bug

I have tested on mockbukkit mc versions 1.16 and 1.19
When using the Player#addAttachment with a permission that is a parent of other nodes it does not enable them permissions.

For example take the permission myplugin.* that is the parent of all other permission in the plugin
When adding the attachment myplugin.* the child node permissions will not be enabled

### Reproducible Test

1) Make your plugin yaml have child permissions. For example
```yml
permissions:
  myplugin.*:
    description: Gives access to all permissions
    children:
      myplugin.reload: true
      myplugin.use: true
      myplugin.enable: true
```

2) Add the attachment during testing `myplugin.*` to the player object

3) When trying to complete an action that requires the permissions reload, use, or enable, they will not work. The player will be sent the message: "I'm sorry, but you do not have permission to perform this command"

### Anything else?

_No response_

Thorinwasher:
# Description
Fixes #612 (Hopefully)

# Checklist


Insprill (changes requested):
We should add the exact case the issue was reporting to the TestPlugin and write some tests using it to make sure this fixes the intended issue at all.

Thorinwasher:
> We should add the exact case the issue was reporting to the TestPlugin and write some tests using it to make sure this fixes the intended issue at all.

There, I now added tests with respective fixes.

svaningelgem:
It'd be nice if this could be included in the next release as I would be using it in my plugin development.

thelooter:
@Insprill @Thorinwasher anything needed from your side? Otherwise I would merge.

Thorinwasher:
> Insprill Thorinwasher anything needed from your side? Otherwise I would merge.

I don't think so, I can't currently think of anything more that needs o change in this pull request.

Insprill (approved):
LGTM

PR #965, opened 2024-02-08
Implement translation keys

fhnaumann:
# Description

Implement basic translation keys.

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
@@ -330,14 +330,9 @@ public class MockUnsafeValues implements UnsafeValues
 		if(!material.isItem()) {
 			return null;
 		}
-		// edge cases: WHEAT and NETHER_WART are blocks, but still use the "item" prefix (therefore this check has to be done BEFORE the isBlock check below)
-		if(material == Material.WHEAT || material == Material.NETHER_WART) {
-			return formatTranslatable("item", material);
-		}
-		// edge case: If a translation key from an item is requested from anything that is also a block, the block translation key is always returned
-		// e.g: Material#STONE is a block (but also an obtainable item in the inventory). However, the translation key is always "block.minecraft.stone".
-		if(material.isBlock()) {
-			return formatTranslatable("block", material);
+		String edgeCaseHandledTranslationKey = handleTranslateItemEdgeCases(material);
+		if(edgeCaseHandledTranslationKey != null) {
+			return edgeCaseHandledTranslationKey;
 		}
 		return formatTranslatable("item", material);
 	}
@@ -371,14 +366,9 @@ public class MockUnsafeValues implements UnsafeValues
 			if(!material.isItem()) {
 				return null;
 			}
-			// edge cases: WHEAT and NETHER_WART are blocks, but still use the "item" prefix (therefore this check has to be done BEFORE the isBlock check below)
-			if(material == Material.WHEAT || material == Material.NETHER_WART) {
-				return formatTranslatable("item", material);
-			}
-			// edge case: If a translation key from an item is requested from anything that is also a block, the block translation key is always returned
-			// e.g: Material#STONE is a block (but also an obtainable item in the inventory). However, the translation key is always "block.minecraft.stone".
-			if(material.isBlock()) {
-				return formatTranslatable("block", material);
+			String edgeCaseHandledTranslationKey = handleTranslateItemEdgeCases(material);
+			if(edgeCaseHandledTranslationKey != null) {
+				return edgeCaseHandledTranslationKey;
 			}
 			return formatTranslatable("item", material, true);
 		}
@@ -390,6 +380,21 @@ public class MockUnsafeValues implements UnsafeValues
 		}
 	}
 
+	private String handleTranslateItemEdgeCases(Material material)
+	{
+		// edge cases: WHEAT and NETHER_WART are blocks, but still use the "item" prefix (therefore this check has to be done BEFORE the isBlock check below)
+		if(material == Material.WHEAT || material == Material.NETHER_WART) {
+			return formatTranslatable("item", material);
+		}
+		// edge case: If a translation key from an item is requested from anything that is also a block, the block translation key is always returned
+		// e.g: Material#STONE is a block (but also an obtainable item in the inventory). However, the translation key is always "block.minecraft.stone".
+		if(material.isBlock()) {
+			return formatTranslatable("block", material);
+		}
+		// not an edge case
+		return null;
+	}
+
 	private <T extends Keyed & Translatable> String formatTranslatable(String prefix, T translatable, boolean fromItemStack)
 	{
 		// enforcing Translatable is not necessary, but translating only makes sense when the object is really translatable by design.
@@ -398,8 +403,8 @@ public class MockUnsafeValues implements UnsafeValues
 			if(Tag.WALL_HANGING_SIGNS.isTagged(material) || Tag.WALL_SIGNS.isTagged(material) || value.endsWith("wall_banner") || value.endsWith("wall_torch") || value.endsWith("wall_skull") || value.endsWith("wall_head")) {
 				value = value.replace("wall_", "");
 			}
-			final Set<Material> EMPTY_EFFECTS = Set.of(Material.POTION, Material.SPLASH_POTION, Material.TIPPED_ARROW, Material.LINGERING_POTION);
-			if(fromItemStack && EMPTY_EFFECTS.contains(material)) {
+			final Set<Material> emptyEffects = Set.of(Material.POTION, Material.SPLASH_POTION, Material.TIPPED_ARROW, Material.LINGERING_POTION);
+			if(fromItemStack && emptyEffects.contains(material)) {
 				value += ".effect.empty";
 			}
 		}
--- a/src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
@@ -302,7 +302,8 @@ public class MockUnsafeValues implements UnsafeValues
 	}
 
 	@Override
-	public String getBlockTranslationKey(Material material)
+	@Nullable
+	public String getBlockTranslationKey(@NotNull Material material)
 	{
 		if (!material.isBlock())
 		{
@@ -317,7 +318,8 @@ public class MockUnsafeValues implements UnsafeValues
 	}
 
 	@Override
-	public String getItemTranslationKey(Material material)
+	@Nullable
+	public String getItemTranslationKey(@NotNull Material material)
 	{
 		if (!material.isItem())
 		{
@@ -332,14 +334,16 @@ public class MockUnsafeValues implements UnsafeValues
 	}
 
 	@Override
-	public String getTranslationKey(EntityType type)
+	@Nullable
+	public String getTranslationKey(@NotNull EntityType type)
 	{
 		Preconditions.checkArgument(type.getName() != null, "Invalid name of EntityType %s for translation key", type);
 		return formatTranslatable("entity", type);
 	}
 
 	@Override
-	public String getTranslationKey(ItemStack itemStack)
+	@Nullable
+	public String getTranslationKey(@NotNull ItemStack itemStack)
 	{
 		if (itemStack.getType().isItem())
 		{
--- a/src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
@@ -301,82 +301,62 @@ public class MockUnsafeValues implements UnsafeValues
 		throw new UnimplementedOperationException();
 	}
 
-	/**
-	 * Gets the translation key for a {@link Material} that {@link Material#isBlock() is a block}.
-	 * @param material the material to translate
-	 * @return a string of the structure "block.{@literal <namespace>}.{@literal <material>}" (e.g. "block.minecraft.stone"), or null if not a block.
-	 */
 	@Override
 	public String getBlockTranslationKey(Material material)
 	{
-		if(!material.isBlock()) {
+		if (!material.isBlock())
+		{
 			return null;
 		}
 		// edge cases: WHEAT and NETHER_WART are blocks, but still use the "item" prefix
-		if(material == Material.WHEAT || material == Material.NETHER_WART) {
+		if (material == Material.WHEAT || material == Material.NETHER_WART)
+		{
 			return formatTranslatable("item", material);
 		}
 		return formatTranslatable("block", material);
 	}
 
-	/**
-	 * Gets the translation key for a {@link Material} that {@link Material#isItem() is an item}.
-	 * @param material the material to translate
-	 * @return a string of the structure "item.{@literal <namespace>}.{@literal <material>}" (e.g. "item.minecraft.carrot"), or null if not an item.
-	 */
 	@Override
 	public String getItemTranslationKey(Material material)
 	{
-		if(!material.isItem()) {
+		if (!material.isItem())
+		{
 			return null;
 		}
 		String edgeCaseHandledTranslationKey = handleTranslateItemEdgeCases(material);
-		if(edgeCaseHandledTranslationKey != null) {
+		if (edgeCaseHandledTranslationKey != null)
+		{
 			return edgeCaseHandledTranslationKey;
 		}
 		return formatTranslatable("item", material);
 	}
 
-	/**
-	 * Gets the translation key for an {@link EntityType}. Throws an error for custom entities.
-	 * @param type the entity to translate
-	 * @return a string of the structure "item.{@literal <namespace>}.{@literal <entity_type>}" (e.g. "entity.minecraft.pig").
-	 */
 	@Override
 	public String getTranslationKey(EntityType type)
 	{
 		Preconditions.checkArgument(type.getName() != null, "Invalid name of EntityType %s for translation key", type);
-		// prevent translation from any custom entity type
-		Arrays.stream(EntityType.values())
-				.filter(entityType ->type.getName().equals(entityType.getName()))
-				.findFirst()
-				.orElseThrow();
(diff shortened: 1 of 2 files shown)


src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
```diff
@@ -296,32 +299,68 @@ public Entity deserializeEntity(byte[] data, World world, boolean preserveUUID)
...
+	 */
 	@Override
 	public String getTranslationKey(ItemStack itemStack)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		if(!itemStack.getType().isItem()) {
+			return null;
+		}
+		return formatTranslatable("item", itemStack.getType());
```
fhnaumann: I'm unsure how to handle translation keys for ItemStacks. I don't know how paper handles translation of custom items...

fhnaumann:
I don't know how to implement tests in a meaningful way. Of course I can assert that my code follows the "{block, item, entity}.minecraft.<Key#value>" pattern, but that pattern is just based on my understanding and testing of translation keys. I don't know a full list I could compare it to, to make sure it works for every case.

Thorinwasher:
> I don't know how to implement tests in a meaningful way. Of course I can assert that my code follows the "{block, item, entity}.minecraft.<Key#value>" pattern, but that pattern is just based on my understanding and testing of translation keys. I don't know a full list I could compare it to, to make sure it works for every case.

You could make a test plugin, which loops through all the translatables when a paper instance starts. That's probably how I would have done it.

fhnaumann:
> > I don't know how to implement tests in a meaningful way. Of course I can assert that my code follows the "{block, item, entity}.minecraft.<Key#value>" pattern, but that pattern is just based on my understanding and testing of translation keys. I don't know a full list I could compare it to, to make sure it works for every case.
> 
> You could make a test plugin, which loops through all the translatables when a paper instance starts. That's probably how I would have done it.

Should I save them in a txt/csv/json and load them in the test?

Thorinwasher:
> Should I save them in a txt/csv/json and load them in the test?

Then you might as well just use those values later on. No, I meant that you already have a script to convert from NamespacedKey to translationkey, just copy that script, like do something like this in a real paper instance:
```java
public void checkTranslatables(String type, Stream<? implements Translatable & Keyed> stream)
{
    stream.filter((keyed) -> !this.getTranslatableKey(keyed, type).equals(keyed.getTranslatable())).forEach((keyed) -> this.getLogger().log(Level.INFO, keyed))
}

String getTranslatableKey(NamespacedKey namespacedKey, String type){
    return String.format("%s.%s.%s",type, material.key().namespace(), material.key().value()); // this is your script, right?
}
```

fhnaumann:
But how can I get a running paper instance in a unit test?

Thorinwasher:
> But how can I get a running paper instance in a unit test?

You don't need to, I meant make an actual test plugin and scan a fully normal server one time to find out. Anyhow, this only matters if you want to be really sure it's fine.

fhnaumann:
> > But how can I get a running paper instance in a unit test?
> 
> You don't need to, I meant make an actual test plugin and scan a fully normal server one time to find out. Anyhow, this only matters if you want to be really sure it's fine.

I just testet it and holy moly there are some edge cases. I will list them there:
1. Items that are also blocks never use the "item" prefix, they always start with "block"
2. "WHEAT" and "NETHER_WART" are considered items and blocks, but they start with "item" instead of "block"
3. Wall signs, hanging wall signs, wall banners, wall torches, wall skulls and wall_heads all don't contain "wall_" in their translate key.
4. ItemStacks behave like materials that are also items, with the exception of "POTION", "SPLASH_POTION", "TIPPED_ARROW", "LINGERING_POTION" (anything that has an "empty effect"). When they are translated from just Material, they translate as expected. But when translated from an ItemStack, an ".effect.empty" is added to the end of it.
This is the code I used:
```java
List<String> translations = new ArrayList<>(Stream.of(
        Arrays.stream(Material.values())
                .filter(material -> !material.isLegacy())
                .filter(Material::isBlock)
                .filter(material -> !Objects.equals(material.translationKey(), getBlockTranslationKey(material))) // comment out to get everything
                .map(material -> String.format("%s, %s", material.translationKey(), getBlockTranslationKey(material)))
                .toList(),
        Arrays.stream(Material.values())
                .filter(material -> !material.isLegacy())
                .filter(Material::isItem)
                .filter(material -> !Objects.equals(material.translationKey(), getItemTranslationKey(material))) // comment out to get everything
                .map(material -> String.format("%s, %s", material.translationKey(), getItemTranslationKey(material)))
                .toList(),
        Arrays.stream(EntityType.values())
                .filter(entityType -> entityType.getName() != null)
                .filter(entityType -> entityType != EntityType.UNKNOWN)
                .filter(entityType -> !entityType.translationKey().equals(formatTranslatable("entity", entityType))) // comment out to get everything
                .map(entityType -> String.format("%s,%s", entityType.translationKey(), getTranslationKey(entityType)))
                .toList(),
        Arrays.stream(Material.values())
                .filter(material -> !material.isLegacy())
                .filter(Material::isItem)
                .map(ItemStack::new)
                .filter(itemStack -> !Objects.equals(itemStack.translationKey(), getTranslationKey(itemStack))) // comment out to get everything
                .map(itemStack -> String.format("%s,%s", itemStack.translationKey(), getTranslationKey(itemStack)))
                .toList()
).flatMap(Collection::stream).toList());
translations.
        add(0, "native, mocked");
try {
    FileWriter writer = new FileWriter(new File(getDataFolder(), "translations.csv"));
    for (String s : translations) {
        writer.write(s + System.lineSeparator());
    }
    writer.close();
} catch (IOException e) {
}
```

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
Thorinwasher: You don't necessarily need to implement translation keys for ItemStacks, you can alternatively throw UnimplementedOperationException

src/test/java/be/seeseemelk/mockbukkit/UnsafeValuesTest.java
```diff
@@ -157,4 +157,28 @@ void fromLegacy_MaterialData_Null_ReturnsNull()
 		assertThrows(NullPointerException.class, () -> mockUnsafeValues.fromLegacy((MaterialData) null));
 	}
 
+	@Test
+	void testBlockTranslationKey()
+	{
+
```
Thorinwasher, line 172: About tests, as you now are able to fetch the data from a real server instance, you can simply just make simple tests on ordinary cases, and some edgecases. So to say, assert the translatable from a certain Namespacedkey is equal to hardcoded values like this:
```java
assertEquals("item.minecraft.mockitem", itemStack.getTranslationKey());
```

fhnaumann: Ok. Lets just hope Mojang doesn't change too much in the future.

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
```diff
@@ -296,32 +301,119 @@ public Entity deserializeEntity(byte[] data, World world, boolean preserveUUID)
...
+		// not an edge case
+		return null;
+	}
+
+	private <T extends Keyed & Translatable> String formatTranslatable(String prefix, T translatable, boolean fromItemStack)
+	{
+		// enforcing Translatable is not necessary, but translating only makes sense when the object is really translatable by design.
+		String value = translatable.key().value();
+		if(translatable instanceof Material material) {
+			if(Tag.WALL_HANGING_SIGNS.isTagged(material) || Tag.WALL_SIGNS.isTagged(material) || value.endsWith("wall_banner") || value.endsWith("wall_torch") || value.endsWith("wall_skull") || value.endsWith("wall_head")) {
```
fhnaumann: Require help here. The tag registry returns null for `WALL_HANGING_SIGNS`. I checked and `WALL_HANGING_SIGNS` is missing in the json file for the tags.

Thorinwasher: It's not currently implemented, I have been working on something that scans all tags in a real server and saves them to JSON files, but I haven't been able to get it to fully work yet. You simply have to ignore this for now. Accept a good enough solution for now.

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
```diff
@@ -296,32 +301,121 @@ public Entity deserializeEntity(byte[] data, World world, boolean preserveUUID)
 		throw new UnimplementedOperationException();
 	}
 
+	/**
+	 * Gets the translation key for a {@link Material} that {@link Material#isBlock() is a block}.
+	 * @param material the material to translate
+	 * @return a string of the structure "block.{@literal <namespace>}.{@literal <material>}" (e.g. "block.minecraft.stone"), or null if not a block.
+	 */
```
Thorinwasher: You don't need to add a comment for the overriden methods.

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
```diff
@@ -296,32 +301,121 @@ public Entity deserializeEntity(byte[] data, World world, boolean preserveUUID)
...
 	@Override
 	public String getBlockTranslationKey(Material material)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		if(!material.isBlock()) {
```
Thorinwasher: Wrong formatting (please reformat all the related files according to mockbukkit specifications; if you're using IntelliJ, this is done automatically)

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
```diff
@@ -296,32 +301,121 @@ public Entity deserializeEntity(byte[] data, World world, boolean preserveUUID)
...
 	public String getTranslationKey(EntityType type)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Preconditions.checkArgument(type.getName() != null, "Invalid name of EntityType %s for translation key", type);
+		// prevent translation from any custom entity type
+		Arrays.stream(EntityType.values())
+				.filter(entityType ->type.getName().equals(entityType.getName()))
+				.findFirst()
+				.orElseThrow();
```
Thorinwasher: This should never throw an error, you don't need to do this check as EntityType is an enum.

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
```diff
@@ -296,32 +301,121 @@ public Entity deserializeEntity(byte[] data, World world, boolean preserveUUID)
 		throw new UnimplementedOperationException();
 	}
 
+	/**
+	 * Gets the translation key for a {@link Material} that {@link Material#isBlock() is a block}.
+	 * @param material the material to translate
+	 * @return a string of the structure "block.{@literal <namespace>}.{@literal <material>}" (e.g. "block.minecraft.stone"), or null if not a block.
+	 */
 	@Override
 	public String getBlockTranslationKey(Material material)
```
Thorinwasher: Even though this is not specified in the API, could you add the Nullable anotation for the output, and NotNull annotation for the input?

Eh, this includes all of the other cases as well, where necessary.

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
```diff
@@ -296,32 +301,121 @@ public Entity deserializeEntity(byte[] data, World world, boolean preserveUUID)
...
+		if(itemStack.getType().isItem()) {
+			Material material = itemStack.getType();
+			if(!material.isItem()) {
+				return null;
+			}
+			String edgeCaseHandledTranslationKey = handleTranslateItemEdgeCases(material);
+			if(edgeCaseHandledTranslationKey != null) {
+				return edgeCaseHandledTranslationKey;
+			}
+			return formatTranslatable("item", material, true);
```
Thorinwasher: Code duplication, use the getItemTranslationKey(Material) method

src/test/java/be/seeseemelk/mockbukkit/UnsafeValuesTest.java
```diff
@@ -157,4 +159,93 @@ void fromLegacy_MaterialData_Null_ReturnsNull()
 		assertThrows(NullPointerException.class, () -> mockUnsafeValues.fromLegacy((MaterialData) null));
 	}
 
+	@Test
+	void testMaterialThatIsOnlyItemTranslationKey()
```
Thorinwasher, line 167: You have multiple tests in one test. Either make this a parameterized test (with for example a methodsource) or separate the tests. Same applies to the other testcases

Thorinwasher (changes requested):
Here's some changes that needs to be addressed before merging.

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
fhnaumann: I cant though. I need to somehow "mark" a Material that is "coming from an itemstack", because "empty effect" materials (POTION for example) have different translation keys when used from an itemstack vs. when used as a material.

src/test/java/be/seeseemelk/mockbukkit/UnsafeValuesTest.java
fhnaumann: I tried grouping them together to be interpretable in a meaningful way. 
```java
assertEquals("block.minecraft.stone", mockUnsafeValues.getBlockTranslationKey(Material.STONE));
assertEquals("block.minecraft.stone", mockUnsafeValues.getItemTranslationKey(Material.STONE));
```
This clearly shows that both translation keys result in the same value.
```java
assertEquals("block.minecraft.acacia_sign", mockUnsafeValues.getBlockTranslationKey(Material.ACACIA_SIGN));
assertEquals("block.minecraft.acacia_sign", mockUnsafeValues.getBlockTranslationKey(Material.ACACIA_WALL_SIGN));
```
This clearly shows that a material and its "wall" counterpart material have the same translation key.

I worry that this readability is lost when extracting them into single tests.

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
Thorinwasher: Ah, I missed the `true` mark there.

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
fhnaumann: Thought I'd "publicly" share the way I faked the vanilla behaviour, but if it's against your style guidelines I can remove them

src/test/java/be/seeseemelk/mockbukkit/UnsafeValuesTest.java
Thorinwasher: Look into Parameterized tests for Junit

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
Thorinwasher: There's no guidline for this, but it seems unnecessary.

src/main/java/be/seeseemelk/mockbukkit/MockUnsafeValues.java
fhnaumann: What annotations do you want? Jetbrains or javax?

Thorinwasher: I think we use jetbrains

src/test/java/be/seeseemelk/mockbukkit/UnsafeValuesTest.java
fhnaumann: How can I assert something programmatically without relying on string names?
```java
	@ParameterizedTest
	@EnumSource(value = Material.class, names = {"STONE", "DIRT"})
	void testMaterialThatIsItemAndBlockTranslationKey(Material material)
	{
		assertEquals("block.minecraft." + material.toString().toLowerCase(), mockUnsafeValues.getBlockTranslationKey(material));
	}
```
I don't really like this because I am relying on the string material name. This becomes especially complicated for the signs, where the material enum string is not correct.
Maybe I will just write a million separate tests.

Thorinwasher: Try methodsource instead. Something like this should work:
```java


@ParameterizedTest
@MethodSource("stringIntAndListProvider")
void testWithMultiArgMethodSource(String str, int num, List<String> list) {
    assertEquals(5, str.length());
    assertTrue(num >=1 && num <=2);
    assertEquals(2, list.size());
}

static Stream<Arguments> stringIntAndListProvider() {
    return Stream.of(
        arguments("apple", 1, Arrays.asList("a", "b")),
        arguments("lemon", 2, Arrays.asList("x", "y"))
    );
}
```

fhnaumann: Do you have a one-assert-per-test rule or were your comments just suggesting I should split the test cases up even further?
Because if you don't then I can split the tests up a little bit.

Thorinwasher: There's no rule for this. I do think it's good practice to split the tests as much as possible. Why? For your case, you have multiple checks that are independent of each other, you could say that they are really individual tests which are placed in the same test case. If the first test in the testcase fails, then the following tests in that testcase will not run, which will give less information from running the test. It's not possible to know if those following tests would have failed as well.

Also please use a parameterized test, it's a cleaner way to do tests that are very similar to each other. (Sorry for being a bit over the place here, btw, I retract my statement about splitting the tests, parameterized tests are that much better).

(edited)

Thorinwasher (approved):
Looks good to me, definitely one of the harder initial pull requests to make!

Thorinwasher:
Unfortuntely, we have to wait for #966 to be completed before merging into v1.20

PR #967, opened 2024-02-12
Possibility to get VOID_AIR blocks above and below world Y border

Figishe:
# Description

Now it's available to get blocks above and below world Y limits. It will return valid VOID_AIR blocks instead of throwing exception.

# Checklist


Thorinwasher:
Even though throwing exceptions here are not exact parity with paper, I still think it's a good idea as the purpose of MockBukkit is different. MockBukkit is only there to test plugins, and can therefore be more stringent with input values than paper.

PR #969, opened 2024-02-14
Implement ChunkMock isForceLoaded

Figishe:
# Description

Basic implementation of isForceLoaded, setForceLoaded for the ChunkMock.

# Checklist


src/main/java/be/seeseemelk/mockbukkit/ChunkMock.java
```diff
@@ -252,15 +253,13 @@ else if (obj instanceof ChunkMock chunk)
 	@Override
 	public boolean isForceLoaded()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return isForceLoaded;
```
Thorinwasher, line 258: Although it seems like this will not fail, could you add one test for the default value of this, and one for setting the value of this?

Thorinwasher (approved):
Looks good to me

Issue #953, opened 2024-01-30
Add Block#getLightFromSky and Block#getLightFromBlocks


boooiil:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

Create an implementation for BlockMock#getLightFromSky and BlockMock#getLightFromBlocks

Thorinwasher:
Just wondering. How exactly should this be implemented? Should there be a script that figures out light levels based on block configurations in the world, or is there an alternative simpler solution for this?

If we're going with a full script that does this, then there are loads of special blocks with specific light behavior, which would need to be implemented.

Insprill:
I think we should simply add setter methods for it. There's no way we can maintain an entire lighting engine that's 1:1 with vanilla/paper, and anything less than that would deviate from expected behaviour

Thorinwasher:
I agree. That would definitely be a lot to maintain.

Thorinwasher:
Fixed with #970

PR #970, opened 2024-02-16
BlockMock | Light level, LightFromSky and LightFromBlocks

4everTheOne:
# Description
Implement getters and setters for BlockMock to resolve issue https://github.com/MockBukkit/MockBukkit/issues/953.

I did not include any hard restriction on the values, so end-users can decide what values suites well. But this includes a soft-warning with the annotation `@Range(from=0, to=15)` to notify the user's that they providing values outside of default Paper values.

See: https://jd.papermc.io/paper/1.20/org/bukkit/block/Block.html#getLightLevel()

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
@@ -30,6 +30,7 @@ import org.bukkit.util.Vector;
 import org.bukkit.util.VoxelShape;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.Range;
 
 import java.util.Collection;
 import java.util.List;
@@ -49,6 +50,10 @@ public class BlockMock implements Block
 	private Material material;
 	private byte data;
 	private BlockData blockData;
+
+	private byte lightLevel = 0;
+	private byte lightFromSky = 15;
+	private byte lightFromBlocks = 0;
 
 	/**
 	 * Creates a basic block made of air.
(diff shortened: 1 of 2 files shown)


src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
```diff
@@ -171,22 +176,49 @@ public void assertType(@NotNull Material material)
 	@Override
 	public byte getLightLevel()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return lightLevel;
```
Thorinwasher: Looked into the source code of paper, this value is the max value of `lightFromSky` and `lightFromBlocks`. Not it's own field

src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
```diff
@@ -171,22 +176,49 @@ public void assertType(@NotNull Material material)
...
+	}
+
+	/**
+	 * Sets the light level received from sky.
+	 *
+	 * @param lightFromSky Value between 0 and 15.
+	 */
+	public void setLightFromSky(@Range(from=0, to=15) byte lightFromSky)
+	{
+		this.lightFromSky = lightFromSky;
```
Thorinwasher, line 195: Be more stringent with the input; have a precondition check for this and `setLightFromBlocks`

Thorinwasher (changes requested):

4everTheOne:
Thank you for the feedback! I have updated the code according to the suggested changes.

src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
```diff
@@ -171,22 +174,47 @@ public void assertType(@NotNull Material material)
...
+	}
+
+	/**
+	 * Sets the light level received from sky.
+	 *
+	 * @param lightFromSky Value between 0 and 15.
+	 */
+	public void setLightFromSky(byte lightFromSky)
+	{
+		if (lightFromSky < 0 || lightFromSky > 15) {
```
Thorinwasher: Use Preconditions.checkArgument(boolean, String) instead. (This is what we use everywhere else, so good for code consistency)

Thorinwasher (approved):

Issue #479, opened 2022-06-18
Add Tests that have been missed in the past


thelooter:
Many Mocks that have been created in the Past have not been covered completely. This should be fixed.

thelooter:
This is not yet finished

Insprill:
GitHub automatically closes issues linked to pull requests when merged.

PR #973, opened 2024-02-22
Unit tests for DamageSourceMock

4everTheOne:
# Description
Added unit tests for class DamageSourceMock (Related with #479).
This also includes commit 49e2007 which is a fix for pipeline to build.

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/damage/DamageTypeMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/damage/DamageTypeMock.java
@@ -48,18 +48,18 @@ public class DamageTypeMock implements DamageType
 		Preconditions.checkArgument(data.has(EXHAUSTION), "JsonObject does not have a field named 'exhaustion'");
 
 		// Retrieve values from JSON
-		var keyValue = data.get(KEY).getAsString();
-		var damageScalingValue = data.get(DAMAGE_SCALING).getAsString();
-		var soundValue = data.get(SOUND).getAsString();
-		var deathMessageTypeValue = data.get(DEATH_MESSAGE_TYPE).getAsString();
-		var exhaustion = data.get(EXHAUSTION).getAsFloat();
+		String keyValue = data.get(KEY).getAsString();
+		String damageScalingValue = data.get(DAMAGE_SCALING).getAsString();
+		String soundValue = data.get(SOUND).getAsString();
+		String deathMessageTypeValue = data.get(DEATH_MESSAGE_TYPE).getAsString();
+		float exhaustion = data.get(EXHAUSTION).getAsFloat();
 
 		// Parse values
-		var key = NamespacedKey.fromString(keyValue);
-		var damageScaling = DamageScaling.valueOf(damageScalingValue);
+		NamespacedKey key = NamespacedKey.fromString(keyValue);
+		DamageScaling damageScaling = DamageScaling.valueOf(damageScalingValue);
 		Sound sound = Registry.SOUNDS.get(NamespacedKey.fromString(soundValue));
-		var damageEffect = new DamageEffectMock(sound);
-		var deathMessageType = DeathMessageType.valueOf(deathMessageTypeValue);
+		DamageEffectMock damageEffect = new DamageEffectMock(sound);
+		DeathMessageType deathMessageType = DeathMessageType.valueOf(deathMessageTypeValue);
 
 		// Create object
 		return new DamageTypeMock(damageScaling, damageEffect, key, deathMessageType, exhaustion);
--- a/src/main/java/be/seeseemelk/mockbukkit/damage/DamageTypeMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/damage/DamageTypeMock.java
@@ -22,7 +22,7 @@ public class DamageTypeMock implements DamageType
 
 	/**
 	 * Create a {@link DamageTypeMock} from a {@link JsonObject}.
-	 * <p />
+	 * <p>
 	 * Example:
 	 * <pre>
 	 * {
--- a/src/main/java/be/seeseemelk/mockbukkit/damage/DamageTypeMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/damage/DamageTypeMock.java
@@ -15,11 +15,11 @@ import org.jetbrains.annotations.NotNull;
 public class DamageTypeMock implements DamageType
 {
 
-	private static final String EXHAUSTION = "exhaustion";
-	private static final String DEATH_MESSAGE_TYPE = "deathMessageType";
-	private static final String SOUND = "sound";
-	private static final String DAMAGE_SCALING = "damageScaling";
-	private static final String KEY = "key";
+	static final String EXHAUSTION = "exhaustion";
+	static final String DEATH_MESSAGE_TYPE = "deathMessageType";
+	static final String SOUND = "sound";
+	static final String DAMAGE_SCALING = "damageScaling";
+	static final String KEY = "key";
 
 	/**
 	 * Create a {@link DamageTypeMock} from a {@link JsonObject}.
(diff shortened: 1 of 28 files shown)


src/test/java/be/seeseemelk/mockbukkit/damage/DamageSourceMockTest.java
```diff
@@ -0,0 +1,204 @@
...
+		assertNotNull(actual);
+		assertNotSame(causingEntity.getLocation(), actual);
+		assertEquals(causingEntity.getLocation(), actual);
+	}
+
+	@Test
+	void getSourceLocation_WhenDamageLocationIsNullAndDirectEntityIsNull() {
+		DamageSourceMock damageSource = new DamageSourceMock(damageType, causingEntity, null, null);
+
+		@Nullable Location actual = damageSource.getSourceLocation();
```
Thorinwasher: This annotation on a local variable is uneccesary.

src/test/java/be/seeseemelk/mockbukkit/damage/DamageSourceMockTest.java
```diff
@@ -0,0 +1,204 @@
...
+		float actual = damageSourceMock.getFoodExhaustion();
+
+		assertEquals(damageType.getExhaustion(), actual);
+	}
+
+	@Test
+	void scalesWithDifficulty_WhenDamageScalingIsNever() {
+
+		JsonObject json = JsonParser.parseString("{\"key\":\"minecraft:in_fire\",\"damageScaling\":\"NEVER\",\"sound\":\"minecraft:entity.player.hurt_on_fire\",\"deathMessageType\":\"DEFAULT\",\"exhaustion\":0.1}").getAsJsonObject();
+		DamageType neverDamage = new DamageTypeMock(json);
```
Thorinwasher: I know I'm he villain here, for making all keyed classes have a JsonObject as an input argument. Maybe it would be better if you added a new constructor for DamageTypeMock, to avoid having to write such heinous json string within the test?

Thorinwasher (changes requested):
Really good that you noticed this! (I was the one that missed adding a testcase for this class).

I have some minor requested changes, but overall good.

4everTheOne:
Hi @Thorinwasher, thank you for the review!

I have migrated the logic from constructor to a method DamageTypeMock#from(JsonObject) in commit [1cb49f0](https://github.com/MockBukkit/MockBukkit/pull/973/commits/1cb49f0658c5609b0df7801c3432edb590b83ec0). I think this approach allow us to have the best of both worlds! But let me know your opinion on that :)

src/test/java/be/seeseemelk/mockbukkit/damage/DamageTypeMockTest.java
```diff
@@ -0,0 +1,79 @@
...
+			"{ 'key': 'minecraft:explosion', 'damageScaling': 'ALWAYS', 'sound': 'minecraft:entity.player.hurt', 'deathMessageType': 'DEFAULT', 'exhaustion': 0.1 }",
+			"{ 'key': 'minecraft:player_explosion', 'damageScaling': 'ALWAYS', 'sound': 'minecraft:entity.player.hurt', 'deathMessageType': 'DEFAULT', 'exhaustion': 0.1 }",
+			"{ 'key': 'minecraft:sonic_boom', 'damageScaling': 'ALWAYS', 'sound': 'minecraft:entity.player.hurt', 'deathMessageType': 'DEFAULT', 'exhaustion': 0.0 }",
+			"{ 'key': 'minecraft:bad_respawn_point', 'damageScaling': 'ALWAYS', 'sound': 'minecraft:entity.player.hurt', 'deathMessageType': 'INTENTIONAL_GAME_DESIGN', 'exhaustion': 0.1 }",
+			"{ 'key': 'minecraft:outside_border', 'damageScaling': 'WHEN_CAUSED_BY_LIVING_NON_PLAYER', 'sound': 'minecraft:entity.player.hurt', 'deathMessageType': 'DEFAULT', 'exhaustion': 0.0 }",
+			"{ 'key': 'minecraft:generic_kill', 'damageScaling': 'WHEN_CAUSED_BY_LIVING_NON_PLAYER', 'sound': 'minecraft:entity.player.hurt', 'deathMessageType': 'DEFAULT', 'exhaustion': 0.0 }"
```
Thorinwasher: Probably a good idea to add this to a MethodSource instead. I think it would make stuff easier to read. You could even make that methodsource load data from a separate file. I'm not that sure about what is best,  but that's the available options (which includes what you're doing).

src/test/java/be/seeseemelk/mockbukkit/damage/DamageTypeMockTest.java
```diff
@@ -0,0 +1,79 @@
...
+	})
+	void from_GivenMinecraftExamples(String jsonString) {
+		JsonObject jsonObject = JsonParser.parseString(jsonString).getAsJsonObject();
+
+		var expectedExhaustion = jsonObject.get(DamageTypeMock.EXHAUSTION).getAsFloat();
+		var expectedDeathMessage = jsonObject.get(DamageTypeMock.DEATH_MESSAGE_TYPE).getAsString();
+		var expectedDamageScaling = jsonObject.get(DamageTypeMock.DAMAGE_SCALING).getAsString();
+		var expectedKey = jsonObject.get(DamageTypeMock.KEY).getAsString();
+		var expectedSound = jsonObject.get(DamageTypeMock.SOUND).getAsString();
```
Thorinwasher: I did not know this existed in java (hehe). Anyhow, I believe this goes against the strength of java, I've always loved how explicit it is, so please change this to explicit types.

Thorinwasher (changes requested):
Good idea to with the DamageTypeMock.from(JsonObject) function. Definittely better than using a constructor!

On that, I plan on making a new pull request that applies this on all other keyed objects (but that has to be in a different pull request)

src/test/java/be/seeseemelk/mockbukkit/damage/DamageTypeMockTest.java
4everTheOne: I agree. I initialy was thinking in using the Junit Argument Provider to load this information from the files and use it as the argument source. But i was not sure if this effort was worthed.
If you agree with it, i can update this PR to include it.

4everTheOne: This would also allows us to resolve the comment below, because it would be resolved as a POJO.
https://github.com/MockBukkit/MockBukkit/pull/973#discussion_r1499368542

Thorinwasher: You're free to decide on what you want to do here. I will approve as long as the comment below is fixed

src/main/java/be/seeseemelk/mockbukkit/damage/DamageTypeMock.java
```diff
@@ -13,13 +14,81 @@
...
+		var soundValue = data.get(SOUND).getAsString();
+		var deathMessageTypeValue = data.get(DEATH_MESSAGE_TYPE).getAsString();
+		var exhaustion = data.get(EXHAUSTION).getAsFloat();
+
+		// Parse values
+		var key = NamespacedKey.fromString(keyValue);
+		var damageScaling = DamageScaling.valueOf(damageScalingValue);
+		Sound sound = Registry.SOUNDS.get(NamespacedKey.fromString(soundValue));
+		var damageEffect = new DamageEffectMock(sound);
+		var deathMessageType = DeathMessageType.valueOf(deathMessageTypeValue);
```
Thorinwasher: I missed that you used var here as well, please remove that

4everTheOne:
I followed your suggestion , and I think the current state is a good trade-off between complexity and code readability.
This was achieved by converting the information from JSON to a record and replacing the ValueSource with MethodSource.

Thorinwasher (approved):
Looks good to me

PR #976, opened 2024-02-23
Add `Marker` mock

Seggan:
# Description

Added `MarkerMock` and added it to `WorldMock`.

# Checklist

### DIFF
--- a/src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
+++ b/src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
@@ -50,6 +50,7 @@ import be.seeseemelk.mockbukkit.entity.LeashHitchMock;
 import be.seeseemelk.mockbukkit.entity.LlamaMock;
 import be.seeseemelk.mockbukkit.entity.LlamaSpitMock;
 import be.seeseemelk.mockbukkit.entity.MagmaCubeMock;
+import be.seeseemelk.mockbukkit.entity.MarkerMock;
 import be.seeseemelk.mockbukkit.entity.MuleMock;
 import be.seeseemelk.mockbukkit.entity.MushroomCowMock;
 import be.seeseemelk.mockbukkit.entity.OcelotMock;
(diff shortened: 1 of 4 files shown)


Seggan:
I don't see error?

Seggan:
Ah, `error: SignMock is not abstract and does not override abstract method getAllowedEditor() in Sign`. Should I implement this? Not really within the PR scope

Thorinwasher:
> Ah, error: SignMock is not abstract and does not override abstract method getAllowedEditor() in Sign. Should I implement this? Not really within the PR scope

That should be in a separate PR

Thorinwasher (changes requested):
Although Marker does not have a different interface than Entity, the behavior is a bit different. Here's the difference:

It's not possible for it to have any passengers; the Marker#addPassenger(Entity) method should not add any passengers to it and always return false.

(There's one more difference, but that is not implemented in EntityMock anyhow)

Thorinwasher (changes requested):
I forgot to mention that there's a test in WorldMockTest which checks all spawnable entities (check for a huge Stream of EntityTypes and entity classes). Please add MarkerMock to that test.

src/test/java/be/seeseemelk/mockbukkit/WorldMockTest.java
```diff
@@ -1330,7 +1331,8 @@ public static Stream<Arguments> getSpawnableEntities()
 				Arguments.of(EntityType.ZOMBIFIED_PIGLIN, PigZombieMock.class),
 				Arguments.of(EntityType.TRIDENT, Trident.class),
 				Arguments.of(EntityType.SPECTRAL_ARROW, SpectralArrow.class),
-				Arguments.of(EntityType.ARROW, Arrow.class)
+				Arguments.of(EntityType.ARROW, Arrow.class),
+				Arguments.of(EntityType.MARKER, MarkerMock.class)
```
Seggan, line 1335: @Thorinwasher already done

Thorinwasher (approved):
Looks good to me!

I missed that you already added the test there, sorry about that.

PR #986, opened 2024-03-22
Make Entity#removePassenger(Entity) remove passenger

Thorinwasher:
# Description

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
@@ -960,9 +960,11 @@ public abstract class EntityMock extends Entity.Spigot implements Entity, Messag
 		{
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


Insprill (commented):
How does this compare to Paper? I'm not around to look at the vanilla logic.

Thorinwasher:
> How does this compare to Paper? I'm not around to look at the vanilla logic.

It's quite complicated in Paper for comparison. There's probably more inconstencies between our mock behavior of that. Anyhow, when I looked into the source code for paper, it seems like the equivalent of the field `vehicle` is set to null before fully dismounting the entity. I'm a not sure why they doing this, possibly for event handling?

When I think about it, this does affect event handling in comparison to paper, as;
- The vehicle will be null during the event
- The vehicle can be changed during the event, but then the event could be cancelled, this needs to be properly reverted

I therefore think this will need another modification to be done properly

thelooter (approved):

PR #997, opened 2024-04-23
Constructors for keyed classes

Thorinwasher:
# Description
1. Adds reasonable constructors for these items.
2. Depricates the all constructors with gson objects inside them; this has now been replaced with a .from method instead
3. Adds tests to all these keyed classes, which previously had not been the case

# Checklist


Thorinwasher:
Nice... at least now I know how the rebase interface works

thelooter (approved):

PR #998, opened 2024-04-23
Implementing respawn location.

svaningelgem:
# Description
Adding respawn location attribute.

Creating this as a draft because I see there is a `respawn` function available (and tested). But I don't know exactly what `.respawn()` should do: should it go to the respawn location, or to the bed? What happens exactly?

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
@@ -172,6 +172,7 @@ public class PlayerMock extends HumanEntityMock implements Player, SoundReceiver
 	private double healthScale = 20;
 	private Location compassTarget;
 	private @Nullable Location bedSpawnLocation;
+	private @Nullable Location respawnLocation;
 	private @Nullable InetSocketAddress address;
 
 	private final PlayerSpigotMock playerSpigotMock = new PlayerSpigotMock();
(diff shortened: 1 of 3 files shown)


Thorinwasher:
> should it go to the respawn location, or to the bed? What happens exactly?

I recomend looking into the Paper code for this, as our behavior in mockbukkit is meant to copy that of paper. Anyhow for your specific question; it looks like there is only one variable that defines the respawn location in Paper. The bed respawn location is that location, here's the implementation of one of the methods:
```java
    public Location getBedSpawnLocation() {
        return this.getRespawnLocation();
    }
```
I hope that clears that up for you

svaningelgem:
Thanks for the hint @Thorinwasher! I have now a copy of the papers code on my computer.
The changes are now in this branch.

Thorinwasher:
> The changes are now in this branch.

Good! Now you need to change the formatting of the code to follow the same style as everything else.

svaningelgem:
No `ruff` or `black` or anything like that for java? :'(

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -1964,20 +1964,11 @@ public int getExperiencePointsNeededForNextLevel()
...
 	}
 
-
-	@Nullable
 	@Override
-	public Location getBedSpawnLocation()
-	{
-		return bedSpawnLocation;
-	}
+	public @Nullable Location getRespawnLocation() { return this.respawnLocation; }
```
Thorinwasher: This here is what I meant reformat the code. Here's an example
```suggestion
	public @Nullable Location getRespawnLocation()
	{ 
		return this.respawnLocation; 
	}
```

svaningelgem: Ok, this should have been reformatted now.

Thorinwasher (approved):

PR #999, opened 2024-04-23
Implemented BookMetaMock::setGeneration

svaningelgem:
# Description
I needed "setGeneration" for BookMeta, so implemented this. Also saw there was no test, so added a rudimentary one for the things that I added.

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
@@ -13,7 +13,6 @@ import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.Unmodifiable;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 
@@ -236,7 +235,7 @@ public class BookMetaMock extends ItemMetaMock implements BookMeta
 	@Deprecated(since = "1.16")
 	public @NotNull List<String> getPages()
 	{
-		return pages;
+		return this.pages;
 	}
 
 	@Override
@@ -253,7 +252,8 @@ public class BookMetaMock extends ItemMetaMock implements BookMeta
 	{
 		this.pages.clear();
 
-        for (String page : pages) {
+        for (String page : pages)
+		{
             this.addPage(page);
         }
 	}
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
@@ -57,7 +57,7 @@ public class BookMetaMock extends ItemMetaMock implements BookMeta
 	{
 		final int prime = 31;
 		int result = super.hashCode();
-		result = prime * result + Objects.hash(author, pages, title);
+		result = prime * result + Objects.hash(author, pages, title, generation);
 		return result;
 	}
 
@@ -71,7 +71,7 @@ public class BookMetaMock extends ItemMetaMock implements BookMeta
 		if (!(obj instanceof BookMetaMock other))
 			return false;
 		return Objects.equals(author, other.author) && Objects.equals(pages, other.pages)
-				&& Objects.equals(title, other.title);
+				&& Objects.equals(title, other.title) && Objects.equals(generation, other.generation);
 	}
 
 	@Override
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
@@ -28,6 +28,7 @@ public class BookMetaMock extends ItemMetaMock implements BookMeta
 	private @Nullable String title;
 	private @NotNull List<String> pages = new ArrayList<>();
 	private @Nullable String author;
+	private @Nullable Generation generation = null;
 
 	/**
 	 * Constructs a new {@link BookMetaMock}.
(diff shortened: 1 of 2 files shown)


src/test/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMockTest.java
```diff
@@ -0,0 +1,83 @@
...
+		assertEquals(meta.getPageCount(), 0);
+
+		assertFalse(meta.hasAuthor());
+		assertFalse(meta.hasTitle());
+		assertFalse(meta.hasGeneration());
+		assertFalse(meta.hasPages());
+	}
+
+	@Test
+	void constructor_Clone_CopiesValues()
```
Thorinwasher, line 54: There's two ways to clone BookMetaMock, one is through the constructor (which is covered here) and another is in the Object#clone() method. Although it looks like that will work here anyhow, I think it's always a good idea to add that test

Another thing (which is an general issue not thought that much of in Mockbukkit to be real here) is that Paper almost always deep clones the objects; all mutable fields of the object has to be cloned as well (recursively). In this case it's fine implementation wise, but feel free to check if the mutable field `pages` changes reference but keeps the same contents between the two copies.

Thorinwasher (changes requested):

src/test/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMockTest.java
svaningelgem: Ok, done.

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
```diff
@@ -250,14 +252,10 @@ public void setPages(String... pages)
...
 		this.pages.clear();
-		Iterator<String> var2 = pages.iterator();
-
-		while (var2.hasNext())
-		{
-			String page = var2.next();
-			this.addPage(page);
-		}
 
+        for (String page : pages) {
```
Insprill: This causes a ConcurrentModificationException, no? We should have tests for all methods here, whether in this PR or another.

Insprill (changes requested):

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMock.java
svaningelgem: IntelliJ gave me that hint and I thought it made sense.
It shouldn't modify anything?

```
        for (String page : pages) {
            this.addPage(page);
        }
```

And in `addPage` you just overwrite the string if it's too long, and add it into the map. So nothing being modified in there as I see? (but my knowledge of Java is lacking, so I am probably wrong)

Insprill: Ahh that's the list passed into the method, not the one in the class; that naming isn't confusing at all.

It should be fine then, but should still have tests.

svaningelgem: Can you suggest which additional tests would be needed?

Thorinwasher: Here's an example of methods that can still be tested.
- hashCode
- equals (this should actually fail with the current implementation)
- hasTitle
- hasAuthor
- getPages (should probably return an copied immutable list, but that's for a different pull request)

svaningelgem: Ok @Thorinwasher : I implemented the requested tests.

src/test/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMockTest.java
```diff
@@ -0,0 +1,303 @@
...
+
+		List<String> pages = new ArrayList<String>();
+		pages.add("Page1");
+		pages.add("Page2");
+
+		meta.setPages(pages);
+		assertEquals(pages, meta.getPages());
+
+		pages.set(0, "Dummy");
+		assertNotEquals(pages, meta.getPages()); // testing that the `pages`' array modifications is not reflected in the book.
```
Insprill: It'd be clearer here to just do `assertNotSame(pages, meta.getPages())` which performs a reference equality check.

Insprill (dismissed):

src/test/java/be/seeseemelk/mockbukkit/inventory/meta/BookMetaMockTest.java
svaningelgem: Ok, didn't know that one.

I changed it now to:
```java
		meta.setPages(pages);
		assertNotSame(pages, meta.getPages());
		assertEquals(pages, meta.getPages());

		pages.set(0, "Dummy");
		assertNotSame(pages, meta.getPages()); // testing that the `pages`' array modifications is not reflected in the book.
		assertNotEquals(pages, meta.getPages()); // testing that the `pages`' array modifications is not reflected in the book.
```

ie: checking that they are NOT the same, but ARE equal.
And secondly, that they're still not the same, but now they're not equal too.

Thorinwasher (changes requested):
Almost there; found one bugg though:

You need to also check if the generation is the same for the two objects to discern if they are equal to each other in the BookMetaMock#equals(Object) method. (and a good idea to add generation to the hashcode as well)

svaningelgem:
Thanks for the review! I completely missed that. Implemented now.

Thorinwasher (approved):

PR #1001, opened 2024-04-24
Implemented BlockDataMock::getAsString(boolean)

svaningelgem:
# Description
In my code I was using the getAsString(true) method of BlockData. So, implementing it here.

# Checklist

### DIFF
--- a/src/main/resources/materials/material_data.json
+++ b/src/main/resources/materials/material_data.json
@@ -1,0 +1,3453 @@
+{
+    "minecraft:acacia_button": {
+        "face": "wall",
+        "facing": "north",
+        "powered": false
+    },
+    "minecraft:acacia_door": {
+        "facing": "north",
+        "half": "lower",
+        "hinge": "left",
+        "open": false,
+        "powered": false
+    },
+    "minecraft:acacia_fence": {
+        "east": false,
+        "north": false,
+        "south": false,
+        "waterlogged": false,
+        "west": false
+    },
+    "minecraft:acacia_fence_gate": {
+        "facing": "north",
+        "in_wall": false,
+        "open": false,
+        "powered": false
+    },
+    "minecraft:acacia_hanging_sign": {
+        "attached": false,
+        "rotation": "0",
+        "waterlogged": false
+    },
+    "minecraft:acacia_leaves": {
+        "distance": "7",
+        "persistent": false,
+        "waterlogged": false
+    },
+    "minecraft:acacia_log": {
+        "axis": "y"
+    },
+    "minecraft:acacia_planks": {},
+    "minecraft:acacia_pressure_plate": {
+        "powered": false
+    },
+    "minecraft:acacia_sapling": {
+        "stage": "0"
+    },
+    "minecraft:acacia_sign": {
+        "rotation": "0",
+        "waterlogged": false
+    },
+    "minecraft:acacia_slab": {
+        "type": "bottom",
+        "waterlogged": false
+    },
+    "minecraft:acacia_stairs": {
+        "facing": "north",
+        "half": "bottom",
+        "shape": "straight",
+        "waterlogged": false
+    },
+    "minecraft:acacia_trapdoor": {
+        "facing": "north",
+        "half": "bottom",
+        "open": false,
+        "powered": false,
+        "waterlogged": false
+    },
+    "minecraft:acacia_wall_hanging_sign": {
+        "facing": "north",
+        "waterlogged": false
+    },
+    "minecraft:acacia_wall_sign": {
+        "facing": "north",
+        "waterlogged": false
+    },
+    "minecraft:acacia_wood": {
+        "axis": "y"
+    },
+    "minecraft:activator_rail": {
+        "powered": false,
+        "shape": "north_south",
+        "waterlogged": false
+    },
+    "minecraft:air": {},
+    "minecraft:allium": {},
+    "minecraft:amethyst_block": {},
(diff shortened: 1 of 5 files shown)


svaningelgem:
Hi,

Because of the modifications I made in here, the following tests start to fail:
```
    AmethystClusterMockTest. getAsString()
    BedMockTest. getAsString()
    CampfireMockTest. getAsString()
    SlabMockTest. getAsString()
    SwitchMockTest. getAsString()
    WallSignMockTest. getAsString()
```

But as far as I can see, my implementation is more correct than before.
Could you please double check?

If you agree with me, I'll adjust the tests mentioned above to call `getAsString(true)` instead.

Thorinwasher:
> Could you please double check?

Best way for these issues (whenever the implementation seems too complex to just imitate) is to test how this works on a real instance. So simply just make a test plugin for a real paper instance (latest version) and then compare the real and the mock behavior. Alternatively you can of course just look at the paper source code, but I have a feeling that it's going to be complicated, as we're talking about serialization.

svaningelgem:
Ok, I wrote a small plugin to dump this information.

This is the result:
```
[14:36:33 INFO]: [MaterialLogger] false|minecraft:yellow_wall_banner[facing=north]
[14:36:33 INFO]: [MaterialLogger] true|minecraft:yellow_wall_banner
[14:36:33 INFO]: [MaterialLogger] false|minecraft:yellow_wool
[14:36:33 INFO]: [MaterialLogger] true|minecraft:yellow_wool
[14:36:33 INFO]: [MaterialLogger] false|minecraft:zombie_head[powered=false,rotation=0]
[14:36:33 INFO]: [MaterialLogger] true|minecraft:zombie_head
[14:36:33 INFO]: [MaterialLogger] false|minecraft:zombie_wall_head[facing=north,powered=false]
[14:36:33 INFO]: [MaterialLogger] true|minecraft:zombie_wall_head
```

svaningelgem:
I fooled a bit around and got this:
https://github.com/AvarionMC/dump_materials/releases/latest/download/material_data.json

Fully automated material generator :-). I can integrate this into MockBukkit? Or you can copy it in your own repos (I can transfer ownership)?

Thorinwasher:
> I fooled a bit around and got this:

Good that you got a test plugin up and running, I don't see any need to have this in the mockbukkit repo, but it is still good to have as comparison for this pull request.

svaningelgem:
FYI:
```
java.lang.IllegalArgumentException: Could not parse data: minecraft:chest[waterlogged=true,bumba=ABC]
Caused by: com.mojang.brigadier.exceptions.CommandSyntaxException: Block minecraft:chest does not have property 'bumba' at position 33: ...gged=true,<--[HERE]
```
==> So this means that paper can't understand invalid properties. (Good that I coded them in already 😁)

svaningelgem:
FYI:
```
[20:51:12 INFO]: [MaterialLogger] Material: minecraft:diamond
[20:51:12 ERROR]: [MaterialLogger]  --> cannot create block data!
[20:51:12 INFO]: [MaterialLogger] Material: minecraft:diamond_axe
[20:51:12 ERROR]: [MaterialLogger]  --> cannot create block data!
[20:51:12 INFO]: [MaterialLogger] Material: minecraft:diamond_block
[20:51:12 INFO]: [MaterialLogger] false|minecraft:diamond_block
[20:51:12 INFO]: [MaterialLogger] true|minecraft:diamond_block
[20:51:12 INFO]: [MaterialLogger] Material: minecraft:diamond_boots
[20:51:12 ERROR]: [MaterialLogger]  --> cannot create block data!
```
So, from some materials, you can't create blocks.

However, that's where I'm stuck at atm:
```java
	{
		MaterialData materialData = new MaterialData(Material.DIAMOND);
		endermanMock.setCarriedMaterial(materialData);

		assertEquals(materialData, endermanMock.getCarriedMaterial());
	}
```

Here a `MaterialData` is created. Which underneath creates: `carriedBlock = BlockDataMock.mock(material.getItemType());`

However, we can't create this... (see above: we can't create a diamond blockdata even in pure paper.)

But for example an enderman can drop enderpearls. But that's not a valid material to create a block from... 

I'd like some guidance for this.
(and I'll push my changes for now so you can already verify what I changed).

Thorinwasher:
> So, from some materials, you can't create blocks.

That's true, there's a method `Material#isBlock()` which you can use to determine this in mockbukkit. I hope that helps

svaningelgem:
@Thorinwasher : This PR is ready (enough) for review :)

src/main/resources/materials/material_data.json
Thorinwasher, line 1: If this is being used in production, could you make a pull request to to MetaMiner with the script you made to retrieve this. (It's in the same organisation as MockBukkit)

Thorinwasher (dismissed):
~Looks good to me (Please see the commnet below)~

I found one possible improvement

src/main/resources/materials/material_data.json
Thorinwasher, line 1: Split this into smaller files; one for each material would be good I believe (it's a lot of data here, but I'm interested if that would be better)

svaningelgem: I would suggest not to do that. There is too much data in there. And loading a json file won't take that long on modern systems. So unless there is some kind of grouping I could do, I'd like to keep it together...

svaningelgem: Done in PR1 there.

Thorinwasher (approved):

PR #1008, opened 2024-05-10
Update to 1.20.6

thelooter:
Update to 1.20.6


build.gradle.kts
```diff
@@ -17,6 +17,9 @@ repositories {
 	maven("https://hub.spigotmc.org/nexus/content/repositories/public/")
 	maven("https://repo.md-5.net/content/groups/public/")
 	maven("https://repo.papermc.io/repository/maven-public/")
+	maven(url = "https://s01.oss.sonatype.org/content/repositories/snapshots/") {
+		name = "sonatype-oss-snapshots"
+	}
```
Insprill:
```suggestion
maven("https://s01.oss.sonatype.org/content/repositories/snapshots/")

File: src/main/java/be/seeseemelk/mockbukkit/block/state/EnderChestMock.java [hunk 2]
```diff
@@ -75,4 +76,11 @@ public boolean isOpen()
 		return isOpen;
 	}
 
+	@Override
+	public boolean isBlocked()
+	{
+		// TODO Auto-generated method stub
+		throw new UnimplementedOperationException();
+	}
+
```
Insprill, line 85: Does Paper have Chest and EnterChest extend from a common class to implement this?

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -887,6 +887,7 @@ public void setFallDistance(float distance)
 	}
 
 	@Override
+	@Deprecated(forRemoval = true,since = "1.20.6")
```
Insprill:
```suggestion
	@Deprecated(forRemoval = true, since = "1.20.6")

File: src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java [hunk 4]
```diff
@@ -1283,4 +1283,6 @@ public float getSidewaysMovement()
 	}
 
 
+
+
```
Insprill:
```suggestion

**Review by Insprill — CHANGES_REQUESTED** [review]
Some preliminary nitpicks

File: build.gradle.kts (hunk 1 above)
Comment (original-codematrix) [reply build.gradle.kts]: Should not be needed anymore in the newest version due to kyori adventure-bom has been released as 2.17.0

**Review by original-codematrix — CHANGES_REQUESTED** [review]

File: src/main/java/be/seeseemelk/mockbukkit/block/state/EnderChestMock.java (hunk 2 above)
Comment (Thorinwasher) [reply src/main/java/be/seeseemelk/mockbukkit/block/state/EnderChestMock.java:85]: No, the chest implementation is different enough due to double chests.

**Comment by GrayR0ot** [comment]
Are changes done ?

**Review by GrayR0ot — APPROVED** [review]

**Comment by Insprill** [comment]
Please do not create unnecessary noise on PRs. It's not going to get it merged faster.

**Comment by GrayR0ot** [comment]
Created https://github.com/MockBukkit/MockBukkit/pull/1015 to fix build issues, they came from 1.20 rebase

File: src/main/java/be/seeseemelk/mockbukkit/services/ServerBuildInfoMock.java [hunk 5]
```diff
@@ -0,0 +1,133 @@
...
+		{
+			Properties properties = new Properties();
+			properties.load(inputStream);
+			this.minecraftVersion = properties.getProperty("minecraft.version");
+		}
+		catch (IOException e)
+		{
+			throw new RuntimeException(e);
+		}
+	}
```
Insprill: We already have a system for this that doesn't require all this overhead. Look at `BuildParameters.java.peb` and the `blossom` block in the `build.gradle.kts`.

src/main/java/be/seeseemelk/mockbukkit/services/ServerBuildInfoMock.java
```diff
@@ -0,0 +1,133 @@
...
+	public @NotNull Optional<String> gitBranch()
+	{
+		return Optional.empty();
+	}
+
+	@Override
+	public @NotNull Optional<String> gitCommit()
+	{
+		return Optional.empty();
+	}
```
Insprill: If we're not going to implement these, we shouldn't implement them. We shouldn't be returning defaults.

On that note, this would also be extremely trivial to implement with the BuildParameters template class. We already get some git info in the build file for the project version outside of actions.

src/main/java/be/seeseemelk/mockbukkit/services/ServerBuildInfoMock.java
Insprill, line 1: This class needs more proper tests, not just "does it return not null".

build.gradle.kts
```diff
@@ -91,6 +91,13 @@ tasks {
 	jacoco {
 		toolVersion = "0.8.12"
 	}
+
+	processResources {
+		filesMatching("**/version_info.properties") { expand(mapOf(
+			"minecraft_version" to project.properties["paper.api.full-version"].toString().split("-")[0],
+			"paper_api_version" to project.properties["paper.api.full-version"]
+		))}
+	}
```
Insprill: See other comment, remove this.

Insprill (changes requested):
Looks pretty good, just some concerns with the `ServerBuildInfoMock`.

src/main/java/be/seeseemelk/mockbukkit/services/ServerBuildInfoMock.java
Thorinwasher: Throwing UnimplementedOperationException is not an option as most tests in mockbukkit would be ignored.

I don't understand how the mockbukkit actions system works. But how would we get the branch and commit and buildnumber when building a new instance?

Do we need to modify actions for this?

src/main/java/be/seeseemelk/mockbukkit/services/ServerBuildInfoMock.java
Thorinwasher: Most of the properties of that class are ever changing, it therefore becomes hard to test without having to reimplement the code your testing in the test to see if it works (which makes the test invalid).

I believe this would be the case for the following methods:
- minecraftVersionId
- minecraftVersionName
- buildNumber
- buildTime
- ?gitBranch?
- gitCommit

These properties either change with every new update, or would be dependant on the branch your using, which should not be in the test, as that's also ever changing.

So the test has to be dynamic, but also actually test stuff.

src/main/java/be/seeseemelk/mockbukkit/services/ServerBuildInfoMock.java
Insprill: All you'd have to do is add new fields to Build Parameters, and set then in the blossom block in the build.gradle.kts. We already have a method or two in there was calls git, so you'd just need to add/modify them to get the info you need.

src/main/java/be/seeseemelk/mockbukkit/services/ServerBuildInfoMock.java
Insprill: I was thinking more along the lines of using regex patterns to ensure the returned values are sane, I believe we do this in the UnsafeMockTest. Should've been more clear, sorry.

Thorinwasher: Okay, that makes more sense, I will look into that.

src/main/java/be/seeseemelk/mockbukkit/services/ServerBuildInfoMock.java
Thorinwasher: Ah, using git from gradle is quite obvious when I think about it. Anyhow, what about the buildNumber method? What would be the mockbukkit equivalent of that; as I said before, we can nott throw unimplementedoperationexception

Insprill: We can get the build number from GitHub actions, IIRC it's the `GITHUB_RUN_NUMBER` we'd want (all variables can be found [here](https://docs.github.com/en/actions/learn-github-actions/variables)). If it's a local build then we can just return empty.

Insprill (approved):
LGTM!

PR #1009, opened 2024-05-10
Add sign side implementation

maxikg:
# Description
This PR implements `org.bukkit.block.Sign#getSide(..)` on the SignMock.

The existing and deprecated methods on `Sign` have been changed to use the front side of a sign. I've added parameterized tests to test both sides of a sign and kept the existing unit tests against Sign.

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
@@ -287,7 +287,7 @@ public class SignMock extends TileStateMock implements Sign
 		}
 
 		@Override
-		public void setColor(DyeColor color)
+		public void setColor(@NotNull DyeColor color)
 		{
 			Preconditions.checkNotNull(color, "Color cannot be null!");
 			this.color = color;
--- a/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
@@ -26,8 +26,8 @@ import java.util.List;
  */
 public class SignMock extends TileStateMock implements Sign
 {
-	private final SignSideImpl front;
-	private final SignSideImpl back;
+	private final SignSideMock front;
+	private final SignSideMock back;
 
 	/**
 	 * Constructs a new {@link SignMock} for the provided {@link Material}.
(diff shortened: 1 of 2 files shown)


src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
```diff
@@ -214,4 +198,99 @@ public void setWaxed(boolean waxed)
 		return new SignMock(this);
 	}
 
+	private static class SignSideImpl implements SignSide {
```
thelooter: Please rename this to `SignSideMock`

src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
```diff
@@ -63,68 +65,48 @@ protected SignMock(@NotNull Block block)
...
-		return LegacyComponentSerializer.legacySection().deserialize(getLine(index));
+		return front.line(index);
 	}
 
 	@Override
 	public void line(int index, @NotNull Component line) throws IndexOutOfBoundsException
 	{
-		Preconditions.checkNotNull(line, "Line cannot be null!");
-		this.lines[index] = LegacyComponentSerializer.legacySection().serialize(line);
+		front.line(index, line);
```
thelooter, line 87: Why was the Precondition removed?

src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
```diff
@@ -214,4 +198,99 @@ public void setWaxed(boolean waxed)
...
+		}
+
+		@Override
+		public @NotNull DyeColor getColor()
+		{
+			return color;
+		}
+
+		@Override
+		public void setColor(DyeColor color)
```
thelooter: Please add a `@NotNull` annotation here

thelooter (changes requested):

src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
maxikg: The method is calling `SignSideMock#line(int, Component)` which contains the null check. So it would be redundant keeping it there.

src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
maxikg: Done in 139fdc3093b9dd6b916cdf2f206914663559eaff

src/main/java/be/seeseemelk/mockbukkit/block/state/SignMock.java
maxikg: Done in 019dd7dbcec9e72bc984dd4b454296567e671707

thelooter (approved):

Issue #1010, opened 2024-05-12
More Inventory Mocking


fhnaumann:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

In #981 I stumbled across some difficulties/inconsistencies using the inventory mocks. This feature request aims to solve these problems as well as to add mocking of new inventories (all in the style of #972).

Each inventory will have its own mock implementation and own simulation class (which receives a mocked instance). This class then has methods like `simulatePlayerClickResult`. These methods all reduce to a call to `simulateInventoryClick` in `PlayerSimulation` with appropriate parameters.
@Thorinwasher Should the simulation implementation be in the simulation class or mock class? You implemented it in the mock class, so I'm unsure what the purpose of the simulation classes are.

Below is a list of features. Mainly I will try to copy the behaviour of `InventoryView#getSlotType`. A useful overview for the different indicies in inventories is [this resource](https://mcutils.com/inventory-slots).

- [x] Workbench
  - [x] Differentiate between matrix and result index
- [ ] Chest
- [ ] Dispenser 
- [ ] Dropper
- [ ] Furnace
  - [ ] Differentiate between input ("crafting"), fuel and result index
- [ ] Enchanting
- [ ] Brewing
  - [ ] Differentiate between fuel and ingredient
 
I will add more depending on the complexity. Right now I cannot guess how complicated everything will be and I will try to move one step at a time.

Thorinwasher:
> Should the simulation implementation be in the simulation class or mock class? You implemented it in the mock class, so I'm unsure what the purpose of the simulation classes are.

It should be in the simulation class. I moved most of code over to the simulation classes, but in some cases it was too much work, so I put it up for later (Whenever / if the simulate method in the mock class get's removed)

To clarify: I think the thoughtprocess behind simulate classes was to lessen class bloat. The simulate methods that are left in their mock classes are only left so due to backwards compatibility.

PR #1011, opened 2024-05-12
Implement more inventory mocks

fhnaumann:
See #1010 for more details.

# Description
Barebone implementation for workbench inventory (tests and documentation missing).

My general idea is to use `items` in `InventoryMock` as the "storage contents" of the inventory. ~~That means this array contains every slot where methods like `addItems` or `removeItems` have an effect. For example, adding (not setting) items to a workbench should only fill the crafting matrix but not the result slot.~~ With this implementation this behaviour does not require any changes in `InventoryMock`, because from its point of view it only knows about the existence of the crafting matrix. Additional slots that have unique behaviour (as far as I know that only the result slot in different inventories), are handled individually in their respective mock implementation. This approach is somehwat similar to what Bukkit/Spigot/Paper does. They have a default "inventory" (rather its an abstraction layer above inventory) which for a crafting inventory would be the matrix. Then they have an additional "inventory" that represents the result slot. 

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
@@ -558,9 +558,7 @@ public class InventoryMock implements Inventory
 	@Override
 	public int firstEmpty()
 	{
-		// Explicitly use items.length and not getSize() because the latter will vary if a subclass has additional slots.
-		// This means firstEmpty only checks storage slots (not result slots).
-		for (int i = 0; i < this.items.length; i++)
+		for (int i = 0; i < getSize(); i++)
 		{
 			if (items[i] == null || items[i].getType() == Material.AIR)
 			{
@@ -641,9 +639,7 @@ public class InventoryMock implements Inventory
 	@Override
 	public boolean isEmpty()
 	{
-		// Explicitly use items.length and not getSize() because the latter will vary if a subclass has additional slots.
-		// This means isEmpty only checks storage slots (not result slots).
-		for (int i = 0; i < this.items.length; i++)
+		for (int i = 0; i < getSize(); i++)
 		{
 			if (items[i] != null && items[i].getType() != Material.AIR)
 			{
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
@@ -3,6 +3,7 @@ package be.seeseemelk.mockbukkit.inventory;
 import be.seeseemelk.mockbukkit.UnimplementedOperationException;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
+import org.bukkit.Bukkit;
 import org.bukkit.Location;
 import org.bukkit.Material;
 import org.bukkit.entity.HumanEntity;
@@ -33,7 +34,7 @@ public class InventoryMock implements Inventory
 
 	private static final int MAX_STACK_SIZE = 64;
 
-	private final ItemStack @NotNull [] items;
+	protected final ItemStack @NotNull [] items;
 	private final @Nullable InventoryHolder holder;
 	private final @NotNull InventoryType type;
 
@@ -73,7 +74,17 @@ public class InventoryMock implements Inventory
 		this.holder = holder;
 		this.type = type;
 
-		items = new ItemStack[type.getDefaultSize()];
+		items = new ItemStack[inventoryType2StorageContentsSize(type)];
+	}
+
+	private static int inventoryType2StorageContentsSize(InventoryType type)
+	{
+		return switch (type)
+		{
+			// workbench has 1 result slot
+			case WORKBENCH -> type.getDefaultSize() - 1;
+			default -> type.getDefaultSize();
+		};
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


Thorinwasher (commented):
I'm mainly wondering if there could be better names for Resultable and Matrixable. I'm not the best at names here, but it's better to give the name a bit more context, MatrixableInventorySimulation maybe?

Also is this abstraction going to be used elsewhere than in WorkbenchSimulation? If not, then it's not really necesary

fhnaumann:
I can change these names, maybe to `IHasMatrixSlot` and `IHasResultSlot`. I like how concise they are, but I hate annotation interface with `I`. The abstraction is going to be used in other inventory simulations. For example, the internal player crafting would implement both interfaces. Inventories like Furnace would only implement `IHasResultSlot` (and maybe something else to represent the fuel slot).

Thorinwasher:
> I can change these names, maybe to IHasMatrixSlot and IHasResultSlot. I like how concise they are, but I hate annotation interface with I. The abstraction is going to be used in other inventory simulations. For example, the internal player crafting would implement both interfaces. Inventories like Furnace would only implement IHasResultSlot (and maybe something else to represent the fuel slot).

Let's avoid that `I` notation. I don't see the issue with having long descriptive names here, now that think about it, my suggestion should be `InventoryMatrixSimulation` (or something like that) and `InventoryResultSimulation`.

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
```diff
@@ -73,7 +74,17 @@ public InventoryMock(@Nullable InventoryHolder holder, @NotNull InventoryType ty
...
+	}
+
+	private static int inventoryType2StorageContentsSize(InventoryType type)
+	{
+		return switch (type)
+		{
+			// workbench has 1 result slot
+			case WORKBENCH -> type.getDefaultSize() - 1;
+			default -> type.getDefaultSize();
+		};
```
fhnaumann, line 92: I had to change this because using `type.getDefaultSize()` would create arrays larger than necessary, as result slots are not in this array. This also caused a unit test to fail in `InventoryMockTest` because there `getSize()` is not implemented adequately.

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
```diff
@@ -234,14 +245,14 @@ public ItemStack addItem(@NotNull ItemStack item)
...
+		for (int i = 0; i < this.getSize(); i++)
 		{
 			ItemStack oItem = items[i];
 			if (oItem == null)
 			{
 				int toAdd = Math.min(item.getAmount(), itemMaxStackSize);
-				items[i] = item.clone();
-				items[i].setAmount(toAdd);
+				this.setItem(i, item);
+				this.getItem(i).setAmount(toAdd);
```
fhnaumann: I was wrong with my initial comment. It seems like adding to an inventory can in fact also add it to the result slot. In this case `items.length==9` (matrix) and `getSize()==10` (workbench mock overrides and accounts the result slot). Subsequently calling `setItem` is necessary instead of directly assigning it to the array, because adding to the result slot has to be possible as well.

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
```diff
@@ -547,7 +558,9 @@ private int first(@NotNull ItemStack item, boolean withAmount)
 	@Override
 	public int firstEmpty()
 	{
-		for (int i = 0; i < getSize(); i++)
+		// Explicitly use items.length and not getSize() because the latter will vary if a subclass has additional slots.
+		// This means firstEmpty only checks storage slots (not result slots).
+		for (int i = 0; i < this.items.length; i++)
```
fhnaumann: This is wrong. I will change this back to getSize().

src/main/java/be/seeseemelk/mockbukkit/inventory/InventoryMock.java
```diff
@@ -628,7 +641,9 @@ public Location getLocation()
 	@Override
 	public boolean isEmpty()
 	{
-		for (int i = 0; i < getSize(); i++)
+		// Explicitly use items.length and not getSize() because the latter will vary if a subclass has additional slots.
+		// This means isEmpty only checks storage slots (not result slots).
+		for (int i = 0; i < this.items.length; i++)
```
fhnaumann: This is wrong. I will change this back to getSize().

fhnaumann:
To summarize the weird implementation of inventories with result slots:

- Some inventories extend CraftResultInventory (for example: Anvil)
  - These inventories store the result slot as a separate inventory and add items to the "correct" inventory based on the given index in `setItem`. They do not override `getStorageContents` or `getContents` and therefore searching, adding or removing (without an index, e.g. methods like `firstEmpty` or `contains`) completely ignores the result slot/inventory.
- Some inventories that have result slots do **not** extend CraftResultInventory. Notably Furnace is an example. In this case the result slot is accounted for in the search methods.
- CraftInventoryCrafting is an exception where it does not extend CraftResultInventory, has its own additional inventory for the result slot, but explicitly overrides `getContents`. Therefore the result slot in a workbench is accounted for in the search methods.

fhnaumann:
Going to postpone this until paper makes inventories more consistent (see [here](https://github.com/PaperMC/Paper/issues/10720)).

PR #1018, opened 2024-05-30
Fix Material#getKey usage

GliczDev:
# Description
Material#getKey doesn't work for legacy materials and throws an exception for messages

# Checklist


Thorinwasher (approved):
LGTM, I tested if this was an issue with any legacy material, and this was the case.

To be more clear: the issue came in any initialization of BlockDataMock when material was of Legacy type.

Another note is that it looks like the legacy materials are internally converted to non legacy materials in paper impl, we should really be doing that instead, but that's for another pull request.

PR #1021, opened 2024-06-05
Fix JitPack

GliczDev:
# Description
Changes jdk version to 21 in jitpack.yml

# Checklist


Insprill (approved):
LGTM. Just a heads up though, we don't really support JitPack anymore since every PR that gets merged pushes a new version to Maven Central which you should be using instead.

PR #1034, opened 2024-06-27
feat(loading): Implement custom config loading

thelooter:
# Description
This allows the developer to supplement a custom config to use during the test

# Checklist


build.gradle.kts
```diff
@@ -55,7 +55,11 @@ tasks {
 		options {
 			(options as? StandardJavadocDocletOptions)?.apply {
 				encoding = "UTF-8"
-
+				tags(
+					"apiNote:a:API Note:",
+					"implSpec:a:Implementation Requirements:",
+					"implNote:a:Implementation Note:"
+				)
```
Insprill, line 62: What's this?

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
```diff
@@ -25,7 +32,9 @@
 public class MockBukkit
 {
 
+	private static final Logger log = LoggerFactory.getLogger(MockBukkit.class);
 	private static @Nullable ServerMock mock = null;
+	private static final String BUKKIT_PLUGIN_SUPERCLASS = "org.bukkit.plugin.java.JavaPlugin";
```
Insprill: Is there a way for us to get this that's checked at compile time?

build.gradle.kts
thelooter: This adds support for Javadoc Tags added in Java 8

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
thelooter: What do you mean by that? Can you elaborate

seeseemelk (approved):

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
```diff
@@ -267,6 +269,117 @@ public static void loadJar(@NotNull File jarFile) throws InvalidPluginException
...
+					.getName()
+					.equals(BUKKIT_PLUGIN_SUPERCLASS))
+			{
+				throw new IllegalStateException("Invalid plugin superclass");
+			}
+			Field configField = instance
+					.getClass()
+					.getSuperclass()
+					.getSuperclass()
+					.getDeclaredField("newConfig");
```
Thorinwasher:
```suggestion
			Field configField = JavaPlugin.class.getDeclaredField("newConfig");
```
Why not do this instead?

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
```diff
@@ -267,6 +269,117 @@ public static void loadJar(@NotNull File jarFile) throws InvalidPluginException
...
+		{
+			if (!instance.
+					getClass()
+					.getSuperclass()
+					.getSuperclass()
+					.getName()
+					.equals(BUKKIT_PLUGIN_SUPERCLASS))
+			{
+				throw new IllegalStateException("Invalid plugin superclass");
+			}
```
Thorinwasher: Is there not an edgecase scenario where the plugin extends another plugin. Who knows right? This explicitly requires the class `org.bukkit.plugin.java.JavaPlugin` to be what the plugin extends. (Yes it checks the superclass of the superclass, but the initial class is a proxy class anyhow, right?)

i would remove this whole section.

Thorinwasher (changes requested):

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
Thorinwasher: @Insprill do you mean this https://github.com/MockBukkit/MockBukkit/pull/1034#discussion_r1665520972

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
thelooter: Thei issue might be that we can't get the field if JavaPlugin is not the superclass of the superclass

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
thelooter: If i recall correctly, you actually need the specific instance of the field on a specific instance of a class for this to work

Thorinwasher: Isn't there going to be only one class instance of JavaPlugin?

I don't think the class loader will create multiple duplicate instances of that class.

Insprill: Has this been tested? I do believe Thorinwasher's version should work since `getSuperclass` and `JavaPlugin.class` both just return a `Class`, and you pass in the desired instance when performing read/write operations. If this is the case (which it should be), the above check and `BUKKIT_PLUGIN_SUPERCLASS` can both be removed.

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
Insprill: I was thinking something like `private static final String BUKKIT_PLUGIN_SUPERCLASS = JavaPlugin.class.getName()` instead of storing it as a string. It should™️return the same thing (and I think it can still be final?)

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
thelooter: It seems to work, but I fear it might break if you try loading multiple Plugins
But I guess thats something we have to look at down the road

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
Insprill: It'll work regardless of how far up the inheritance chain it is, right? Can we recursively search through superclasses until we find it and only fail if it isn't anywhere in the chain?

src/main/java/be/seeseemelk/mockbukkit/MockBukkit.java
```diff
@@ -25,7 +32,9 @@
 public class MockBukkit
 {
 
+	private static final Logger log = LoggerFactory.getLogger(MockBukkit.class);
```
Insprill: Where is this used?

src/test/java/be/seeseemelk/mockbukkit/MockBukkitTest.java
```diff
@@ -220,4 +233,83 @@ private static class CustomServerMock extends ServerMock
...
+			TestPlugin plugin = MockBukkit.loadWithConfig(TestPlugin.class,inputStream);
+			assertNotNull(plugin);
+
+			FileConfiguration config = plugin.getConfig();
+			String value = config.getString("foo");
+			assertEquals("notbar", value);
+		}
+		catch (IOException e)
+		{
+			throw new RuntimeException(e);
```
4everTheOne: I would suggest replacing this with fail(e)

src/test/java/be/seeseemelk/mockbukkit/MockBukkitTest.java
```diff
@@ -220,4 +233,83 @@ private static class CustomServerMock extends ServerMock
...
+			MockBukkit.mock();
+			RuntimeException runtimeException = assertThrows(RuntimeException.class, () ->
+			{
+				MockBukkit.loadWithConfig(TestPlugin.class, inputStream);
+			});
+			assertEquals("Couldn't read config input stream", runtimeException.getMessage());
+		}
+		catch (IOException e)
+		{
+			throw new RuntimeException(e);
```
4everTheOne: I would suggest also replacing this with fail(e)

4everTheOne (commented):
Looks good to me, just have a small suggestion regarding the tests.

4everTheOne (approved):
Looks good to me

Thorinwasher (approved):

PR #1023, opened 2024-06-15
Add support for getBasePotionType and setBasePotionType

WalshyDev:
# Description
Adds support for getBasePotionType and setBasePotionType.

# Checklist
The following items should be checked before the pull request can be merged.
- [x] Code follows existing style.
- [x] Unit tests added (if applicable).

PR #1036, opened 2024-07-04
Enhance simulateConsumeItem to apply inflicted status effects gained from consumption

fhnaumann:
# Description

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
@@ -37,6 +37,7 @@ import org.bukkit.entity.memory.MemoryKey;
 import org.bukkit.event.entity.EntityDamageByEntityEvent;
 import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.event.entity.EntityDeathEvent;
+import org.bukkit.event.entity.EntityPotionEffectEvent;
 import org.bukkit.event.entity.EntityRegainHealthEvent;
 import org.bukkit.event.entity.EntityToggleSwimEvent;
 import org.bukkit.event.player.PlayerTeleportEvent;
@@ -598,37 +599,98 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 	@Override
 	public boolean addPotionEffect(@NotNull PotionEffect effect)
 	{
-		return addPotionEffect(effect, false);
+		return addPotionEffect(effect, true); // force is ignored
 	}
 
 	@Override
 	@Deprecated(since = "1.15")
 	public boolean addPotionEffect(@NotNull PotionEffect effect, boolean force)
 	{
-		AsyncCatcher.catchOp("effect add");
-		Preconditions.checkNotNull(effect, "PotionEffect cannot be null");
 		// Bukkit now allows multiple effects of the same type,
 		// the force/success attributes are now obsolete
-		activeEffects.add(new ActivePotionEffect(effect));
-		return true;
+		addPotionEffect(effect, EntityPotionEffectEvent.Cause.PLUGIN);
+		return true; // legacy behaviour always returned true
+	}
+
+	/**
+	 * Adds a potion effect with the {@link EntityPotionEffectEvent.Cause#PLUGIN} cause. If the event is canceled, no effect will be added.
+	 *
+	 * @param effect The Potion Effect to add.
+	 * @return The event containing details about adding the potion effect.
+	 * @see LivingEntityMock#addPotionEffect(PotionEffect, EntityPotionEffectEvent.Cause)
+	 */
+	public EntityPotionEffectEvent addPotionEffectWithPluginCause(@NotNull PotionEffect effect)
+	{
+		return addPotionEffect(effect, EntityPotionEffectEvent.Cause.PLUGIN);
+	}
+
+	/**
+	 * Adds a potion effect. If the event is canceled, no effect will be added.
+	 *
+	 * @param effect The Potion Effect to add.
+	 * @param cause  The cause.
+	 * @return The event containing details about adding the potion effect.
+	 * @see LivingEntityMock#addPotionEffectWithPluginCause(PotionEffect)
+	 */
+	public EntityPotionEffectEvent addPotionEffect(@NotNull PotionEffect effect, EntityPotionEffectEvent.Cause cause)
+	{
+		AsyncCatcher.catchOp("effect add");
+		Preconditions.checkNotNull(effect, "PotionEffect cannot be null");
+
+		/*
+		Applying completely new effect -> old: null, new: new effect, action: add, override: false
+		A single effects runs out (on time) (not possible via this method because @NotNull) -> old: effect that ran out, new: null, action: remove, override: true
+		Applying a new effect but effect type already active -> old: existing effect, new: new effect, action: changed, override: true
+		Clearing all effects (not possible via this method) -> old: effect that was cleared, new: null, action: clear, override: true
+
+		 Notes:
+		 If multiple effects are cleared, then for each one an EntityPotionEffectEvent is called.
+		 */
+
+		PotionEffect oldEffect = getPotionEffect(effect.getType());
+		EntityPotionEffectEvent.Action action = oldEffect == null ? EntityPotionEffectEvent.Action.ADDED : EntityPotionEffectEvent.Action.CHANGED;
+		boolean override = oldEffect != null;
+
+
+		EntityPotionEffectEvent event = new EntityPotionEffectEvent(this, oldEffect, effect, cause, action, override);
+		Bukkit.getPluginManager().callEvent(event);
+		if (!event.isCancelled())
+		{
+			activeEffects.add(new ActivePotionEffect(effect));
+		}
+		return event;
+	}
+
+	/**
+	 * Adds multiple potion effects with the {@link EntityPotionEffectEvent.Cause#PLUGIN} cause. If one event is canceled, the effect from that event won't be added.
+	 *
+	 * @param effects The Potion Effects to add.
+	 * @return A list of events containing details about adding the potion effects.
+	 * @see LivingEntityMock#addPotionEffects(Collection, EntityPotionEffectEvent.Cause)
+	 */
+	public List<EntityPotionEffectEvent> addPotionEffectsWithPluginCause(@NotNull Collection<PotionEffect> effects)
+	{
+		return addPotionEffects(effects, EntityPotionEffectEvent.Cause.PLUGIN);
+	}
+
+	/**
+	 * Adds multiple potion effects. If one event is canceled, the effect from that event won't be added.
+	 *
+	 * @param effects The Potion Effects to add.
+	 * @return A list of events containing details about adding the potion effects.
+	 * @see LivingEntityMock#addPotionEffectsWithPluginCause(Collection)
+	 */
+	public List<EntityPotionEffectEvent> addPotionEffects(@NotNull Collection<PotionEffect> effects, EntityPotionEffectEvent.Cause cause)
+	{
+		Preconditions.checkNotNull(effects, "PotionEffect cannot be null");
+		return effects.stream().map(potionEffect -> addPotionEffect(potionEffect, cause)).toList();
 	}
 
 	@Override
 	public boolean addPotionEffects(@NotNull Collection<PotionEffect> effects)
 	{
-		Preconditions.checkNotNull(effects, "PotionEffect cannot be null");
-
-		boolean successful = true;
-
-		for (PotionEffect effect : effects)
-		{
-			if (!addPotionEffect(effect))
-			{
-				successful = false;
-			}
-		}
-
-		return successful;
+		addPotionEffects(effects, EntityPotionEffectEvent.Cause.PLUGIN);
+		return true; // legacy behaviour always returned true
 	}
 
 	@Override
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
@@ -613,24 +613,11 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 	}
 
 	/**
-	 * Adds a potion effect with the {@link EntityPotionEffectEvent.Cause#PLUGIN} cause. If the event is canceled, no effect will be added.
-	 *
-	 * @param effect The Potion Effect to add.
-	 * @return The event containing details about adding the potion effect.
-	 * @see LivingEntityMock#addPotionEffect(PotionEffect, EntityPotionEffectEvent.Cause)
-	 */
-	public EntityPotionEffectEvent addPotionEffectWithPluginCause(@NotNull PotionEffect effect)
-	{
-		return addPotionEffect(effect, EntityPotionEffectEvent.Cause.PLUGIN);
-	}
-
-	/**
 	 * Adds a potion effect. If the event is canceled, no effect will be added.
 	 *
 	 * @param effect The Potion Effect to add.
 	 * @param cause  The cause.
 	 * @return The event containing details about adding the potion effect.
-	 * @see LivingEntityMock#addPotionEffectWithPluginCause(PotionEffect)
 	 */
 	public EntityPotionEffectEvent addPotionEffect(@NotNull PotionEffect effect, EntityPotionEffectEvent.Cause cause)
 	{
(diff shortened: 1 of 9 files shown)


fhnaumann:
I don't know what caused it, but for some reason tests are failing because there are compilation errors because `canUseEquipmentSlot(EquipmentSlot)` is not implemented anywhere.

src/main/resources/foods/food_properties.json
Thorinwasher, line 1: I'm not certain where you got all of this data from, but make sure there's no duplication of the `/resources/keyed/item.json` file. If any of these properties can be accessed through the ItemData class, they should do so.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -336,6 +339,25 @@ public void simulateConsumeItem(@NotNull ItemStack consumable)
 							!Bukkit.isPrimaryThread());
 			Bukkit.getPluginManager().callEvent(stopConsumeEvent);
 		}
+		else
+		{
+			// apply status effects
+			FoodConsumption foodConsumption = FoodConsumption.getFor(consumable.getType());
+			for (FoodConsumption.FoodEffect foodEffect : foodConsumption.foodEffects())
+			{
+				// TODO: how to handle foodEffect with probability?
```
Thorinwasher: Randomness and unit testing does not match well, probably a good idea to instead have a boolean switch in this method that determines the outcome.

Thorinwasher (commented):
Some prelimenary comments

Thorinwasher:
> I don't know what caused it, but for some reason tests are failing because there are compilation errors because canUseEquipmentSlot(EquipmentSlot) is not implemented anywhere.

I think paper updated their API a bit. I believe the error means that we have to implement that method. I will make a separate pr for that

src/main/resources/foods/food_properties.json
fhnaumann: `/keyed/item.json` only contains information if the item is edible, the data I extracted contains special data relevant for every time that is edible. I extract the data from `net.minecraft.world.food.Foods` class.
This is the entire code:
```java
  private JsonObject asJsonObj(FoodProperties.PossibleEffect possibleEffect) {
    JsonObject jsonObject = new JsonObject();
    jsonObject.addProperty("probability", possibleEffect.probability());
    MobEffectInstance mobEffectInstance = possibleEffect.effect();
    ResourceLocation resourceLocation = mobEffectInstance.getEffect().unwrapKey().orElseThrow().location();
    jsonObject.addProperty("name", resourceLocation.toString());
    jsonObject.addProperty("amplifier", mobEffectInstance.getAmplifier());
    jsonObject.addProperty("duration", mobEffectInstance.getDuration());
    jsonObject.addProperty("ambient", mobEffectInstance.isAmbient());
    jsonObject.addProperty("particles", mobEffectInstance.isVisible());
    jsonObject.addProperty("icon", mobEffectInstance.showIcon());
    return jsonObject;
  }

  @Override
  public JsonElement generateData() {
    JsonArray data = new JsonArray();
    try {
      //Class<?> foodsClass = Class.forName("net.minecraft.world.food.Foods");
      Class<?> foodsClass = Foods.class;
      for (Field field : foodsClass.getFields()) {
        JsonObject entry = new JsonObject();
        entry.addProperty("name", Material.matchMaterial(field.getName()).key().asString());
        FoodProperties foodProperty = (FoodProperties) field.get(null);
        entry.addProperty("nutrition", foodProperty.nutrition());
        entry.addProperty("saturationModifier", foodProperty.saturation());
        entry.addProperty("canAlwaysEat", foodProperty.canAlwaysEat());
        entry.addProperty("eatDurationTicks", foodProperty.eatDurationTicks());
        List<JsonObject> effectsAsJson = foodProperty.effects().stream().map(possibleEffect -> asJsonObj(possibleEffect)).toList();
        JsonArray effectsAsJsonArray = new Gson().toJsonTree(effectsAsJson).getAsJsonArray();
        entry.add("effects", effectsAsJsonArray);

        data.add(entry);
      }
    } catch (IllegalAccessException e) {
    }
      return data;
  }
```

fhnaumann:
Also another question: I want to call `EntityPotionEffectEvent` when simulating consumptions. Currently this event is not provided in mocks anywhere. For example `PlayerMock#addPotionEffect` does not call this event. Should I
A) call the event in the item consumption and only if not canceled, call `addPotionEffect`
B) change the behaviour of `addPotionEffect` to call the event
If B), can I make these changes here or should I create another PR for that?

Thorinwasher:
> Also another question: I want to call `EntityPotionEffectEvent` when simulating consumptions. Currently this event is not provided in mocks anywhere. For example `PlayerMock#addPotionEffect` does not call this event. Should I A) call the event in the item consumption and only if not canceled, call `addPotionEffect` B) change the behaviour of `addPotionEffect` to call the event If B), can I make these changes here or should I create another PR for that?

I think B is the better solution here, and sure you can do it in this pull request (I don't think it's that out of scope)

src/main/resources/foods/food_properties.json
Thorinwasher: Okay, good to know. Then adding all this new scrapping data is indeed necesary

src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -598,37 +599,98 @@ public void setKiller(@Nullable Player killer)
...
+	 *
+	 * @param effect The Potion Effect to add.
+	 * @return The event containing details about adding the potion effect.
+	 * @see LivingEntityMock#addPotionEffect(PotionEffect, EntityPotionEffectEvent.Cause)
+	 */
+	public EntityPotionEffectEvent addPotionEffectWithPluginCause(@NotNull PotionEffect effect)
 	{
-		Preconditions.checkNotNull(effects, "PotionEffect cannot be null");
+		return addPotionEffect(effect, EntityPotionEffectEvent.Cause.PLUGIN);
+	}
```
Thorinwasher: I don't think this new method is necessary, as there already is the `LivingEntity#addPotionEffect(PotionEffect)` method, which already does the same thing. Generally an action that has been triggered by the paper api, should be listed as if it was caused by a plugin anyways

src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -598,37 +599,98 @@ public void setKiller(@Nullable Player killer)
...
+	 * Adds multiple potion effects with the {@link EntityPotionEffectEvent.Cause#PLUGIN} cause. If one event is canceled, the effect from that event won't be added.
+	 *
+	 * @param effects The Potion Effects to add.
+	 * @return A list of events containing details about adding the potion effects.
+	 * @see LivingEntityMock#addPotionEffects(Collection, EntityPotionEffectEvent.Cause)
+	 */
+	public List<EntityPotionEffectEvent> addPotionEffectsWithPluginCause(@NotNull Collection<PotionEffect> effects)
+	{
+		return addPotionEffects(effects, EntityPotionEffectEvent.Cause.PLUGIN);
+	}
```
Thorinwasher: Same can be said here, as the comment above

src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -598,37 +599,98 @@ public void setKiller(@Nullable Player killer)
...
 
-		return successful;
+	/**
+	 * Adds multiple potion effects. If one event is canceled, the effect from that event won't be added.
+	 *
+	 * @param effects The Potion Effects to add.
```
Thorinwasher, line 654: Missing param in javadoc for `cause`

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -297,11 +300,19 @@ public boolean reconnect()
...
+	 * @param inflictPotionEffect A flag determining whether potion effects that don't have a probability of 1 are
+	 *                            inflicted on the consumer. For example, consuming {@link Material#CHICKEN} has a 0.3
+	 *                            chance to inflict the hunger effect. If this flag is true, then the hunger effect will
+	 *                            be inflicted. Otherwise, it won't be inflicted. This does not prevent effects from
+	 *                            being inflicted that have a probability of 1. These will be inflicted regardless. The
+	 *                            flag is ignored if the edible does not inflict any potion effects.
+	 * @see PlayerMock#simulateConsumeItem(ItemStack)
 	 */
-	public void simulateConsumeItem(@NotNull ItemStack consumable)
+	public void simulateConsumeItem(@NotNull ItemStack consumable, boolean inflictPotionEffect)
```
Thorinwasher:
```suggestion
	public void simulateConsumeItem(@NotNull ItemStack consumable, boolean alwaysInflictPotionEffect)
```
Somewhat more desriptive parameter name

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -297,11 +300,19 @@ public boolean reconnect()
...
+	 * probability.
 	 *
-	 * @param consumable The Item to consume
+	 * @param consumable          The Item to consume
+	 * @param inflictPotionEffect A flag determining whether potion effects that don't have a probability of 1 are
+	 *                            inflicted on the consumer. For example, consuming {@link Material#CHICKEN} has a 0.3
+	 *                            chance to inflict the hunger effect. If this flag is true, then the hunger effect will
+	 *                            be inflicted. Otherwise, it won't be inflicted. This does not prevent effects from
+	 *                            being inflicted that have a probability of 1. These will be inflicted regardless. The
+	 *                            flag is ignored if the edible does not inflict any potion effects.
```
Thorinwasher: If possible, can this description be shortened? Maybe to something like this (you're probably better at english than me though)
```
Whether to always inflict potion effect from food regardless of probability. If this is is `false` and the food 
item has a probability to inflict the effect lesser than 0, it will not do so. 
```

src/test/java/be/seeseemelk/mockbukkit/food/FoodConsumptionLoadTest.java
Thorinwasher, line 1: Maybe add an individual test, which checks whether the contents of one of the food items are correct (not all of course, that would be too much)

src/main/java/be/seeseemelk/mockbukkit/potion/MockInternalPotionData.java
```diff
@@ -104,23 +103,32 @@ public int getMaxLevel()
 		return maxLevel;
 	}
 
-	private List<PotionEffect> getPotionEffectFromData(JsonObject data)
+	// TODO probably not the right solution to make this public static, but where should this utility method be located?
+
+	/**
+	 * Internal, do not use!
+	 */
```
Thorinwasher:
```suggestion
	@ApiStatus.Internal
```
This is a thing

src/main/java/be/seeseemelk/mockbukkit/potion/MockInternalPotionData.java
```diff
@@ -104,23 +103,32 @@ public int getMaxLevel()
 		return maxLevel;
 	}
 
-	private List<PotionEffect> getPotionEffectFromData(JsonObject data)
+	// TODO probably not the right solution to make this public static, but where should this utility method be located?
```
Thorinwasher, line 107: I think it's a necessary evil to accomplish what you're trying to do here. With the above suggestion I believe it's fine

src/main/java/be/seeseemelk/mockbukkit/potion/MockInternalPotionData.java
```diff
@@ -104,23 +103,32 @@ public int getMaxLevel()
...
+		int amplifier = potionEffectDataObj.get("amplifier").getAsInt();
+		boolean ambient = potionEffectDataObj.get("ambient").getAsBoolean();
+		boolean particles = potionEffectDataObj.get("particles").getAsBoolean();
+		boolean icon = potionEffectDataObj.get("icon").getAsBoolean();
+		return new PotionEffect(potionEffectType, duration, amplifier, ambient, particles, icon);
+	}
+
+	/**
+	 * Internal, do not use!
+	 */
```
Thorinwasher:
```suggestion
	@ApiStatus.Internal
```

Thorinwasher (changes requested):
Here's some changes that are necessary from my point of view, mostly nitpicks though

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
fhnaumann: I shortened it a little, but I think adding a sentence that the flag does change the behaviour of effects with a probability of 1 as well as a sentence that the flag does not anything at all if the item does not inflict any status effects, is necessary clarification.

src/test/java/be/seeseemelk/mockbukkit/entity/PlayerMockTest.java
```diff
@@ -1847,6 +1851,57 @@ void testAssertItemConsumedWithNullItem()
...
+		);
+	}
+
+	private static ItemStack potionItemStack(PotionType potionType) {
+		ItemStack itemStack = new ItemStack(Material.POTION);
+		PotionMeta potionMeta = (PotionMeta) itemStack.getItemMeta();
+		potionMeta.setBasePotionType(potionType);
+		itemStack.setItemMeta(potionMeta);
+		return itemStack;
+	}
```
fhnaumann: This fails because `getItemMeta` requires the mocked server. Usually adding `MockBukkit#getOrCreate()` is enough, but because this test class uses its own `ServerMock` implementation, I don't know how to handle this, since I cannot just change the call there to `MockBukkit#getOrCreate()`.

Thorinwasher: I see, there's two simple approaches here (there's probably many more):
1. Use a non parameterized test (only for one item)
2. Make each argument in the methodsource be a [supplier](https://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html), which you can use to supply all the necessary data in the actuall test method

I think 1 is good enough here, as 2 feels like it becomes a bit messy

fhnaumann: I don't think 2 is too messy, will try that.

src/main/java/be/seeseemelk/mockbukkit/potion/MockInternalPotionData.java
fhnaumann, line 1: I have a question about `upgradable` and `extendable`: The way the check is written, I am interpreting it more like `upgraded` and `extended`. So do these flags indicate that the potion, generally speaking, has a strong and long variant, or do these flags indicate that the potion _is_ upgraded or extended?
This is relevant because I need to implement `PotionMetaMock#getBasePotionType` and other methods belonging to this. There I need the information if a given `PotionType` is in its upgraded or extended form.

fhnaumann: Figured it out. It really means that they are upgradable or extendable and not that the potion itself is upgraded or extended. See my comments on the relevant code.

src/main/java/be/seeseemelk/mockbukkit/potion/MockInternalPotionData.java
```diff
@@ -45,6 +47,9 @@ public MockInternalPotionData(NamespacedKey namespacedKey)
 		this.potionEffects = tempPotionEffects;
 		this.upgradeable = Registry.POTION.get(new NamespacedKey(namespacedKey.getNamespace(), "strong_" + namespacedKey.getKey())) != null;
 		this.extendable = Registry.POTION.get(new NamespacedKey(namespacedKey.getNamespace(), "long_" + namespacedKey.getKey())) != null;
+		this.upgraded = namespacedKey.getKey().startsWith("strong_");
+		this.extended = namespacedKey.getKey().startsWith("long_");
```
fhnaumann: I need these two flags so I could implement the methods in `PotionMetaMock`.

Thorinwasher: Someone else tried / is trying to do something like this, convert between potion data and potion effect type. See comment https://github.com/MockBukkit/MockBukkit/pull/1023#discussion_r1642457994 on how to do so properly.

Most notably the namespaced key supplied into the internal potion data constructor should be from a base potion effect. You can see that that is assumed in the two lines above (I think I wrote that code...).

fhnaumann: So is what I wrote fine? I dont mind if it changed later. For now I just want the tests to pass where I access it to test something else.

Thorinwasher: I do not like this implementation, as it adds functionallity to the MockInternalPotionData which it is not meant to have. Even if you say it's temporary, this is often not the case in software development. This is also getting out of scope

Even though it's not ideal, can't you just avoid this in the tests for now? For say, not use upgraded / extended potion effects.

I hope that clears up stuff / I'm trying to be more clear when reviewing, but I still have a lot to learn

fhnaumann: I can disable the tests for now. I cannot just avoid these in tests. I cannot create any potion itemstack, because setting the base potion type throws an NotImplementedError.

fhnaumann:
I think I'm done for this PR. Unless you want some changes, I'm ready for merging this.
I'll handle the other cases (effect is cleared, milk is consumed, effect runs out) where to call a EntityPotionEffectEvent in a separate PR, once this is merged.

src/main/java/be/seeseemelk/mockbukkit/potion/MockInternalPotionData.java
Thorinwasher: I see, but you can't set the potion data instead of potion effect type?

fhnaumann: Line 371 in `PlayerMock` is `PotionType.InternalPotionData internalPotionData = server.getUnsafe().getInternalPotionData(Registry.POTION.getKey(potionMeta.getBasePotionType()));`. The `potionMeta.getBasePotionType()` throws the error (I accidentally left it implemented, changing it to throwing an error right now). I don't see any way around this currently without implementing it. I'm not fully aware what the problem is and why can't can be fixed easily with another few parameters to store data. Essentially only a class variable to store and return what is set and accessed in the basePotionType methods?

Thorinwasher (approved):

Thorinwasher:
@fhnaumann Now that this has been merged, could you make a pr for this in metaminer as well?

PR #1038, opened 2024-07-06
Update 1.21

thelooter:
# Description

# Checklist


src/main/java/be/seeseemelk/mockbukkit/block/state/LecternMock.java
Thorinwasher, line 1: Very nitpicky here, but the only changes are added import statements, these new import statements probably are dormant.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -3086,6 +3087,12 @@ public void damageItemStack(@NotNull EquipmentSlot slot, int amount)
 		throw new UnimplementedOperationException();
 	}
 
+	@Override
+	public boolean canUseEquipmentSlot(@NotNull EquipmentSlot slot)
+	{
+		return false;
```
Thorinwasher: This seems like it should throw an unimplemented operation exception instead.

Thorinwasher (commented):
I don't think I have that much to add, found some minor nitpicks though

src/main/resources/potion/potion/awkward.json
Thorinwasher, line 1: Why is this in the `potion/potion` folder?

Would it not be better if it was in the `potion`folder, or am I missing something

Thorinwasher (commented):
One nitpick.

Should we also use the latest MetaMiner version which sort almost every resource alphabetically, or is it better to use that in a different pull request?

Insprill:
This needs to be merged into a new `v1.21` branch; it shouldn't be merged into `v1.20`.

Insprill (approved):
I realistically can't review all of this, but it compiles and tests pass, so you've got an OK from me.

PR #1044, opened 2024-07-18
Update username to use user token instead of username

seeseemelk:
Note that this also requires a different password, which will be set as a github secret


seeseemelk:
I'm setting the patch label so a new build will be published

PR #1047, opened 2024-07-24
Implemented simulateUseItem in PlayerMock

JamesPeters98:
# Description
Add simulateUseItem to PlayerMock
Currently only has implementations for opening Containers

The current state can be used to test plugins that manipulate the Inventory and also to test PlayerInteractEvents that cancel the Event (Since the ItemStack simulateUse isn't implemented yet) 

# Checklist


Nirbose:
The code standard isn't good.

JamesPeters98:
> The code standard isn't good.

Such as... ?

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -595,6 +603,104 @@ public void respawn()
...
+					itemInteractionResult = itemStackMock.simulateUse(this, location, hand);
+					itemStackCopy.setAmount(count);
+				} else {
+					itemInteractionResult = itemStackMock.simulateUse(this, location, hand);
+				}
+
+				if (itemInteractionResult.consumesAction()) {
+					// TODO advancements
+				}
+			}
```
Thorinwasher: This segment has become too complicated, You need to simplify this code so that it's easier to understand.
Try using methods for this instead.

Why is it complicated?

I generally have no issue with the length of this method, but there's too many nested if/else statements.

src/main/java/be/seeseemelk/mockbukkit/world/EnumInteractionResult.java
Thorinwasher, line 1: I don't think you need to call this enum `EnumInteractionResult` it's generally clear that it is an enum already. `InteractionResult` seems better.

Another note: This enum is reduntant, only one value is used: CONSUME.

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -457,4 +460,9 @@ else if (Tag.BUTTONS.isTagged(material))
 		return null;
 	}
 
+	public EnumInteractionResult simulateUse(PlayerMock playerMock, EquipmentSlot equipmentSlot)
+	{
+		throw new UnimplementedOperationException("BlockDataMock::use is not implemented for Material "+getMaterial());
+	}
```
Thorinwasher: There's only this implementation of simulateUse for anything implementing BlockData. This code is therefore reduntant // not usefull. Please remove this for now, and when/if it becomes usefull, then this might be added.

src/main/java/be/seeseemelk/mockbukkit/block/state/ContainerMock.java
```diff
@@ -121,4 +124,13 @@ public void setCustomName(@Nullable String name)
 		return ((InventoryMock) this.inventory).getSnapshot();
 	}
 
+	@Override
+	public EnumInteractionResult simulateUse(PlayerMock playerMock, EquipmentSlot equipmentSlot)
+	{
+		if (inventory != null) {
+			playerMock.openInventory(inventory);
+		}
+		return EnumInteractionResult.CONSUME;
```
Thorinwasher: Why does this have anything to do with CONSUME interaction result, you haven't determined the item yet, seems odd to say that it will always consume the item.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -595,6 +603,104 @@ public void respawn()
...
+				openInventory(containerMock.getInventory());
+			}
+		} else {
+			boolean isItemInHand = !getItemInHand().isEmpty() || !getInventory().getItemInOffHand().isEmpty();
+			boolean shouldCancelUse = isSneaking() && isItemInHand;
+			ItemStack itemStackCopy = itemInHand.clone();
+
+			if (!shouldCancelUse) {
+				if (block.getState() instanceof BlockStateMock blockStateMock) {
+					EnumInteractionResult result = blockStateMock.simulateUse(this, hand);
```
Thorinwasher: Why do you check an result here, BlockStateMock#simulateUse will only return CONSUME, or throw an UnimplementedOperationException

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -595,6 +603,104 @@ public void respawn()
...
+
+				if (getGameMode() == GameMode.CREATIVE) {
+					int count = itemStackMock.getAmount();
+					itemInteractionResult = itemStackMock.simulateUse(this, location, hand);
+					itemStackCopy.setAmount(count);
+				} else {
+					itemInteractionResult = itemStackMock.simulateUse(this, location, hand);
+				}
+
+				if (itemInteractionResult.consumesAction()) {
```
Thorinwasher: This if statement will always be true, see above comment

Thorinwasher (changes requested):
I think you are reading into the future too much, it's a bad habit as it is not possible to see what the future entails. Here's some changes I would like to see

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
JamesPeters98: I basically just replicated the logic that NMS does, and left comments in for things I didn't implement yet, I can remove them if that's what you want though. Do you want me to just extract the logic into separate methods then

src/main/java/be/seeseemelk/mockbukkit/block/state/ContainerMock.java
JamesPeters98: This is NMS logic and it consumes the interaction not the item

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
JamesPeters98: Well, my next PR would be to implement some block data logic, I just wanted a basis reviewed and committed first

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
JamesPeters98: Future BlockState's will return something other than SUCCESS when implemented?

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
Thorinwasher: I understand your reasoning there. Can't you add that into this pr though?

A lot of the current code does not make any sense without that extra implementation

src/main/java/be/seeseemelk/mockbukkit/block/state/ContainerMock.java
Thorinwasher: Oh, I understand. Maybe it would be better if it's called something else though; we don't need to copy nms code, alot of nms code is actually a mess.

What about renaming all of the interaction results to `.*_ACTION`, for example `CONSUME_ACTION`.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
Thorinwasher: > I basically just replicated the logic that NMS doe

As I wrote in another comment, a lot of nms code is a mess. And to add to that, code that has been added to prod, usually stays in prod indefinitely. I also believe mockbukkit has to have higher code standards, as we don't have as many developers as for say minecraft/microsoft. That's why we can't copy all nms code directly.

> Do you want me to just extract the logic into separate methods then

Preferably, yes. I think 3 nested if statements are in a grayzone area, while 4 should not be allowed at all. I would actually unwrap this if else statement
```java
if (block.getState() instanceof BlockStateMock blockStateMock)
{
    // Code here
}
else
{
}
```
To this:
```java
if (!(block.getState() instanceof BlockStateMock blockStateMock))
{
}
// Code here
```
Then you would at least be in the grayzone.


Also this can easily be replaced with a private method:
```java
if (itemStackCopy instanceof ItemStackMock) 
{
	itemStackMock = (ItemStackMock) itemStackCopy;
} 
else
{
	itemStackMock = new ItemStackMock(itemStackCopy);
}
```


Oh I just noticed, it's also not following code formatting standards of mockbukt.

Thorinwasher: Here's another one you can do:

This code
```java
if (getGameMode() == GameMode.CREATIVE)
{
	int count = itemStackMock.getAmount();
	itemInteractionResult = itemStackMock.simulateUse(this, location, hand);
	itemStackCopy.setAmount(count);
}
else
{
	itemInteractionResult = itemStackMock.simulateUse(this, location, hand);
}
```

Can easily be replaced with this:
```java
int count = itemStackMock.getAmount();
EnumInteractionResult itemInteractionResult = itemStackMock.simulateUse(this, location, hand);
if (getGameMode() == GameMode.CREATIVE)
(
	itemStackCopy.setAmount(count);
}
```

Like there's a lot of small things you can do to decrease the complexity of that code, this is even without using new methods.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -595,6 +603,104 @@ public void respawn()
...
+					itemStackMock = new ItemStackMock(itemStackCopy);
+				}
+
+				if (getGameMode() == GameMode.CREATIVE) {
+					int count = itemStackMock.getAmount();
+					itemInteractionResult = itemStackMock.simulateUse(this, location, hand);
+					itemStackCopy.setAmount(count);
+				} else {
+					itemInteractionResult = itemStackMock.simulateUse(this, location, hand);
+				}
```
Thorinwasher: Looked into the sourcecode for paper, and you're not replicating the behavior here, these operations should not be run on a copy on the original itemstack, but actually on the original itemstack itself. So to say, you're instansiating a local variable itemStackCopy, and then you are modifying that instance: It's doing nothing.

(Note that I'm ignoring the fact that running this throws an UnimplementedOperationException anyways)

Thorinwasher (changes requested):
Found a bug, so I added that to a new review.

I also noticed that you're not following the mockbukkit code formatting standards.

JamesPeters98:
@Thorinwasher I have added some new changes after following through the logic next to the Paper decomp. I was missing a useWithoutItem case which is implemented now.

I have also added logic for using a lever and also the trap door.

I have also renamed the InteractionResult enums

Thorinwasher:
@JamesPeters98 Sorry for the delay, I was on vacation. Will look into this now

src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java
```diff
@@ -98,6 +98,25 @@ public BlockMock(@NotNull Material material, @Nullable Location location)
...
+	{
+		if (block instanceof BlockMock blockMock)
+		{
+			return blockMock;
+		}
+		else
+		{
+			return new BlockMock(block.getLocation());
+		}
+	}
```
Thorinwasher, line 118: I think throwing an exception here is more fit, if Block is not an BlockMock instance, then there's already a huge issue. I would really just do this (not in it's separate method):
```java
BlockMock blockMock = (BlockMock) block;
```

src/main/java/be/seeseemelk/mockbukkit/block/data/LeverMock.java
```diff
@@ -0,0 +1,132 @@
...
+		BlockRedstoneEvent eventRedstone = new BlockRedstoneEvent(block, old, current);
+		playerMock.getServer().getPluginManager().callEvent(eventRedstone);
+
+		if ((eventRedstone.getNewCurrent() > 0) == (powered))
+		{
+			return InteractionResult.SUCCESS;
+		}
+
+		pull(blockState.getWorld(), location);
+		return InteractionResult.CONSUME;
```
Thorinwasher, line 122: Should there be similar behavior for ButtonMock for this // why did you separate the class SwitchMock into two classes?

src/main/java/be/seeseemelk/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -452,9 +458,19 @@ else if (Tag.WALL_SIGNS.isTagged(material))
...
 
+	public ItemInteractionResult simulateUseItemOn(BlockStateMock blockStateMock, Location location, ItemStackMock itemStack, PlayerMock playerMock, EquipmentSlot equipmentSlot, BlockFace clickedFace)
+	{
+		return ItemInteractionResult.PASS_TO_DEFAULT_BLOCK_INTERACTION;
+	}
+
+	public InteractionResult simulateUseWithoutItem(BlockStateMock blockState, Location location, PlayerMock playerMock, BlockFace clickedFace)
+	{
+		return InteractionResult.PASS;
+	}
```
Thorinwasher, line 474: Might be good to make this @ApiStatus.Internal?

Otherwise, if you want these to be part of the api, could you comment them?

src/main/java/be/seeseemelk/mockbukkit/block/state/BlockStateMock.java
```diff
@@ -420,65 +426,55 @@ else if (MaterialTags.SKULLS.isTagged(block))
...
+	}
+
+	public InteractionResult simulateUseWithoutItem(BlockStateMock blockState, Location location, PlayerMock playerMock, BlockFace clickedFace)
+	{
+		if (blockData instanceof BlockDataMock blockDataMock)
+		{
+			return blockDataMock.simulateUseWithoutItem(blockState, location, playerMock, clickedFace);
 		}
+		throw new UnimplementedOperationException();
 	}
```
Thorinwasher, line 478: Might be good to make these @ApiStatus.Internal.

Otherwise, if you want these to be part of the api, could you comment them?

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -594,6 +602,117 @@ public void respawn()
...
+				{
+					itemInHand.setAmount(count);
+				}
+
+				if (result.consumesAction())
+				{
+					// TODO advancements
+					return event;
+				}
+			}
```
Thorinwasher, line 710: I think you have to make this whole segment a separate method, even with the improvements added, it's still hard to read. 

I agree this is not an easy situation, as adding more abstraction is not necessarily good.

src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -250,6 +273,11 @@ else if (args.containsKey("meta"))
 		return result;
 	}
 
+	public InteractionResult simulateUseItemOn(PlayerMock playerMock, Location clickedPos, EquipmentSlot hand)
+	{
+		throw new UnimplementedOperationException();
+	}
```
Thorinwasher, line 279: Why are you adding a placeholder method here?

It does not add any feature to mockbukkit, It's completely unnecessary.

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -594,6 +602,117 @@ public void respawn()
...
+					}
+				}
+			}
+
+			if (!itemInHand.isEmpty() && !interactResult)
+			{
+				ItemStackMock itemStackMock = ItemStackMock.mock(itemInHand);
+				int count = itemStackMock.getAmount();
+
+				InteractionResult result = itemStackMock.simulateUseItemOn(this, location, hand);
```
Thorinwasher, line 698: This will always throw an UnimplementedOperationException. Everything inside this code segment is reduntant

src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -103,6 +108,24 @@ private ItemStackMock(@NotNull ItemType type)
...
+	{
+		if (itemStack instanceof ItemStackMock itemStackMock)
+		{
+			return itemStackMock;
+		}
+		else
+		{
+			return new ItemStackMock(itemStack);
+		}
+	}
```
Thorinwasher, line 127: This will either return a new instance of an item stack or not, this is generally going to cause issues, as you don't know whether the returned object is the same instance as the input object.

What you really need to do, is to use reflection to fetch the item stack mock field from item stack. Something like this when the item stack is not a mocked instance:
```java
Field field = ItemStack.class.getField("craftDelegate");
field.setAccessible(true);
return (ItemStackMock) field.get(itemStack);
```

If it's not clear, ItemStack is now just a wrapper class for either an ItemStackMock instance (when mocking) or a CraftItemStack instance (in production) // accessing the class it's delegating will yield the same result as if is the same instance (for most cases).

Thorinwasher (changes requested):
There you go, some changes are still necessary

src/main/java/be/seeseemelk/mockbukkit/block/data/LeverMock.java
JamesPeters98: Because the logic is slightly different for them, and they are separated in NMS. The button can be different variants of a material, so those need to be implemented with (net.minecraft.world.level.block.state.properties.BlockSetType) so that the correct sound can be played for the buttons.

I pulled out the Lever because that only needs to play the Lever sound effect, it's not dependent on material.

I will probably go through and create a separate PR to implement these BlockSetType's for buttons, doors, trapdoors etc.
image

Thorinwasher: Alright, I can accept that explanation, it's probably better // yeah the interaction is different when clicking a lever and a button.

Thorinwasher:
@JamesPeters98 Sorry if I have been a bit inactive here. How is this pull request going along?

I just saw that you added a comment a week ago, I have now answered it. (see https://github.com/MockBukkit/MockBukkit/pull/1047#discussion_r1715182081 ).

Mostly some nitpicks left:
- Try to make that huge code block more readable
- Add comments to new methods that are not overridden, or simply add the @ApiStatus.Internal annotation
- Some issues with __future to be implemented code__, you need to choose where the complete code should be added: In this pull request or in another pull request. Just to add my own argument to why I believe this is a reasonable approach is that:
    1. We don't know where exactly how those future pull requests will look like, it's unnecessary to prepare for them to this extent, things might change implementation wise anyhow.
    2. We don't know when/if these future pull requests will be merged. At the moment the code will just be there to do nothing
    3. It's confusing: someone could mistake this for fully implemented code. At the moment this could confuse other future developers -> more code requires more maintenance, which is acceptable when the code does something, but not so much otherwise.

PR #1058, opened 2024-08-04
Add more tests to item stack class with fixes to those tests

### LINKED ISSUE #1028: Adopt paper 1.21 ItemStack delegation

### Describe the feature

Paper 1.21 will have all [ItemStack](https://jd.papermc.io/paper/1.20.6/org/bukkit/inventory/ItemStack.html) methods simply call a delegate [ItemStack](https://jd.papermc.io/paper/1.20.6/org/bukkit/inventory/ItemStack.html) instance, which holds a CraftItemStack.

This enables ItemStack to have direct access to a server internal impl and is the first step to moving [ItemStack](https://jd.papermc.io/paper/1.20.6/org/bukkit/inventory/ItemStack.html) over to an interface.

As such, the constructor for [ItemStack](https://jd.papermc.io/paper/1.20.6/org/bukkit/inventory/ItemStack.html) 
will either make use of ItemType (and its respective implementation) or a new method `UnsafeValues#createEmptyStack`.

To properly support plugins creating ItemStacks, MockBukkit will need to mock [ItemType#createItemStack(int)](https://jd.papermc.io/paper/1.20.6/org/bukkit/inventory/ItemType.html#createItemStack(int)) and the previously mentioned `UnsafeValues#createEmptyStack`.

See: https://github.com/PaperMC/Paper/pull/10852.

### LINKED ISSUE #1111: Incorrect behavior with ItemStackMock#getMaxStackSize()

### Minecraft Version

1.21

### Describe the bug

In 1.21, an item stack's max stack size can be changed in `ItemMeta`, but this change is not reflected in [`ItemStackMock#getMaxStackSize()`](https://github.com/MockBukkit/MockBukkit/blob/f7980147d593f2e14047c0fdd9c26aeb62e992af/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java#L159), which still returns the item type's default max stack size.

### Reproducible Test

```java

        ItemStack stack = new ItemStack(Material.STICK, 65);
        ItemMeta meta = stack.getItemMeta();
        meta.setMaxStackSize(99);
        stack.setItemMeta(meta);
        assertEquals(99, stack.getMaxStackSize());
```

### Anything else?

[Bukkit's implementation](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/browse/src/main/java/org/bukkit/inventory/ItemStack.java#254)

Thorinwasher:
# Description
See conversation in issue #1028 

Uses data generated from MockBukkit/MetaMiner

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
@@ -146,7 +146,7 @@ public class ItemStackMock extends ItemStack
 	@Override
 	public int getMaxStackSize()
 	{
-		return this.type.getMaxStackSize();
+		return this.itemMeta.hasMaxStackSize() ? this.itemMeta.getMaxStackSize() : this.type.getMaxStackSize();
 	}
 
 	@Override
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
@@ -210,6 +210,21 @@ public class ItemStackMock extends ItemStack
 		return isSimilar(bukkit) && this.amount == bukkit.getAmount();
 	}
 
+	@Override
+	public int hashCode()
+	{
+		if (type == ItemType.AIR)
+		{
+			return EMPTY.hashCode();
+		}
+		else
+		{
+			int hash = type.hashCode();
+			hash = hash * 31 + this.getAmount();
+			return hash;
+		}
+	}
+
 	private static @Nullable ItemMeta findItemMeta(Material material)
 	{
 		if (!material.isItem() || material == Material.AIR)
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
@@ -17,6 +17,7 @@ import javax.annotation.Nullable;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 
 public class ItemStackMock extends ItemStack
 {
@@ -24,7 +25,7 @@ public class ItemStackMock extends ItemStack
 	private ItemType type = ItemTypeMock.AIR;
 	private int amount = 1;
 	private ItemMeta itemMeta = new ItemMetaMock();
-	private short durability;
+	private short durability = -1;
 
 	private static final ItemStackMock EMPTY = new ItemStackMock((Void) null);
 	private static final String ITEMMETA_INITIALIZATION_ERROR = "Failed to instanciate item meta class ";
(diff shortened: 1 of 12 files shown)


src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -217,6 +193,42 @@ public boolean equals(Object obj)
...
+	{
+		if (type == ItemType.AIR)
+		{
+			return EMPTY.hashCode();
+		}
+		else
+		{
+			int hash = type.hashCode();
+			hash = hash * 31 + this.getAmount();
+			return hash;
```
Cybermaxke, line 249: Also use durability and item meta to calculate the hash?

Thorinwasher: I will investigate that, does sound reasonable to do, I do believe I tested on a real instance where I changed the item meta: This had no effect on item equality (which always should mean the hash needs to be the same)

Thorinwasher:
When writing this, I have realised one of our tests are not testing anything. This pr added some logic into itemstack that slightly increased the test coverage for that test which therefore now fails. This pr can not build until #1076 is finished and merged.

src/main/java/be/seeseemelk/mockbukkit/inventory/ItemTypeMock.java
```diff
@@ -76,8 +76,6 @@ public static ItemType from(JsonObject jsonObject)
 								|| metaClassAsString.equals("BlockDataMeta")
 								|| metaClassAsString.equals("EnchantmentStorageMeta")
 								|| metaClassAsString.equals("MusicInstrumentMeta")
-								|| metaClassAsString.equals("OminousBottleMeta")
-								|| metaClassAsString.equals("BundleMeta")
```
4everTheOne, line 80: Was this intentionaly removed?

src/main/java/be/seeseemelk/mockbukkit/inventory/meta/OminousBottleMetaMock.java
```diff
@@ -42,4 +44,33 @@ public void setAmplifier(int amplifier)
...
 		return clone;
 	}
+
+	@Override
+	protected void deserializeInternal(@NotNull Map<String, Object> args)
+	{
+		super.deserializeInternal(args);
+		if (args.containsKey("amplifier"))
+		{
+			this.amplifier = (Integer) args.get("amplifier");
```
4everTheOne: I would just suggest moving the "amplifier" string to a constant. (This will reduce sonar issues)

src/main/java/be/seeseemelk/mockbukkit/inventory/ItemTypeMock.java
Thorinwasher: Those item meta classes were already implemented, so doing this allows using those item metas

4everTheOne (approved):
Looks good to me!

PR #1059, opened 2024-08-06
feat: add support for getBasePotionType and setBasePotionType (new)

ybw0014:
# Description

This PR adds support for base potion type.

Part of this PR is from https://github.com/MockBukkit/MockBukkit/pull/1023, I am trying to finish his PR because he seems very busy.

Partial credit to: @WalshyDev

# Checklist


thelooter:
Is there a way to dynamically initialize those bimaps with potions? This seems really prone to breaking on subsequent versions

ybw0014:
> Is there a way to dynamically initialize those bimaps with potions? This seems really prone to breaking on subsequent versions

they are just simple keyed enum elements, I can't think of a good idea.

Thorinwasher:
>Is there a way to dynamically initialize those bimaps with potions? This seems really prone to breaking on subsequent versions

In this case I ?think? it would be fine as potion data is a legacy feature. Otherwise, we could generate a json file for this in MockBukkit/MetaMiner

thelooter:
> > Is there a way to dynamically initialize those bimaps with potions? This seems really prone to breaking on subsequent versions
> 
> In this case I ?think? it would be fine as potion data is a legacy feature. Otherwise, we could generate a json file for this in MockBukkit/MetaMiner

I would prefer this. We should have as little state as possible hardcoded.
Realistically this is a style decision, so I'm willing to accept this solution if I'm outvoted

ybw0014:
How do I use MetaMiner to generate json? Do i need to open a new PR for a new generator, or you will do this?
Or I just keep the current state and wait for this to be merged?

ybw0014:
I've used json to store these mappings.

Thorinwasher:
> How do I use MetaMiner to generate json?

I think you already managed to do this. Anyhow you need to scan a real paper instance using meta miner (you can then run `gradle runServer` and that should start a server and do everything automatically for you)

> Do i need to open a new PR for a new generator, or you will do this?

Yes please do that, you've already generated those files, seems to be a waste to reimplement that. I will merge when this pr is merged.

> I've used json to store these mappings.

Alright. Will look into this now (going away on the weekend though)

Braekpo1nt:
This would be super handy :)

Thorinwasher (approved):
Nice! Looks good to me.

I was initially going to comment something on the nullability of base potion data, but this seems to have changed in later versions of paper i.e null is allowed, so all good.

PR #1063, opened 2024-08-10
Feat/1.21.1

thelooter:
# Description
Depends on #1062 

# Checklist


src/main/java/be/seeseemelk/mockbukkit/block/banner/PatternTypeMock.java
```diff
@@ -0,0 +1,46 @@
...
+	private final NamespacedKey key;
+	private final String identifier;
+
+	public PatternTypeMock(NamespacedKey key,String name, int ordinal,String identifier)
+	{
+		super(name, ordinal,key);
+		this.identifier = identifier;
+		this.key = key;
+	}
+
```
Thorinwasher: Probably best to add the @ApiStatus.Internal annotation to this // from my point of view this method should not be used in any way except internally

src/main/resources/keyed/attribute.json
Thorinwasher, line 1: Seems like meta miner needs to sort everything even more alphabetically. I will self assign myself that task

Thorinwasher (approved):
Looks good to me

Thorinwasher:
I forgot one comment requested a change, anyhow if we merge this, it doesn't matter that much from my point of view

PR #1069, opened 2024-08-14
Merge MetaMiner into the MockBukkit repository

Thorinwasher:
# Description
As time goes on with MockBukkit, more generated files have been necessary to produce. From the constant updating of the generator of these files `MetaMiner` I have realised that it becomes a hassle to keep the two projects synhronized. That's why I think it's a better idea to completely merge these projects into one repository.

This pull request does exactly that, MetaMiner has now been moved into the `extra/MetaMiner` folder in the MockBukkit repository. I also added a simple script that moves the generated files over to the mocbukkit project.

# Checklist


Thorinwasher:
Just going to comment, that it's probably a good idea to merge the changes from metaminer before merging this into the main branch

thelooter (changes requested):
Please convert it to a git submodule

Thorinwasher:
> Please convert it to a git submodule

I dissagree with that, with the stated reasons in my description, i.e. we need to keep these projects completely synched. Anyhow, I think this might be a somewhat heated topic, so we probably need to have a voice call about this.

thelooter (approved):

PR #1072, opened 2024-08-15
Added mocks for Illager and Pillager

4everTheOne:
# Description
Implemented mocks for [Illager](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/Illager.html) and [Pillager](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/Pillager.html).

A small refactor in InventoryMock was required, to support this change. In Paper, the Inventory does not have a restriction in sizes, only in CraftServer.
To resolve this, the validation was moved to ChestInventory.

# Checklist

# Relates

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/IllagerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/IllagerMock.java
@@ -1,0 +1,28 @@
+package be.seeseemelk.mockbukkit.entity;
+
+import be.seeseemelk.mockbukkit.ServerMock;
+import org.bukkit.entity.Illager;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.UUID;
+
+/**
+ * Mock implementation of a {@link Illager}.
+ *
+ * @see RaiderMock
+ */
+public abstract class IllagerMock extends RaiderMock implements Illager
+{
+
+	/**
+	 * Constructs a new {@link RaiderMock} on the provided {@link ServerMock} with a specified {@link UUID}.
+	 *
+	 * @param server The server to create the entity on.
+	 * @param uuid   The UUID of the entity.
+	 */
+	protected IllagerMock(@NotNull ServerMock server, @NotNull UUID uuid)
+	{
+		super(server, uuid);
+	}
+
+}
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/IllagerMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/IllagerMock.java
@@ -15,7 +15,7 @@ public abstract class IllagerMock extends RaiderMock implements Illager
 {
 
 	/**
-	 * Constructs a new {@link RaiderMock} on the provided {@link ServerMock} with a specified {@link UUID}.
+	 * Constructs a new {@link IllagerMock} on the provided {@link ServerMock} with a specified {@link UUID}.
 	 *
 	 * @param server The server to create the entity on.
 	 * @param uuid   The UUID of the entity.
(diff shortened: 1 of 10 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/IllagerMock.java
Thorinwasher, line 1: Mostly a question, there's no different implementation for illagers?

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
```diff
@@ -90,7 +91,7 @@
 import org.spigotmc.event.player.PlayerSpawnLocationEvent;
 
 import javax.imageio.ImageIO;
-import java.awt.Graphics2D;
+import java.awt.*;
```
Thorinwasher: I often do this as well, anyhow please avoid wildcard imports. You can change this in your IDE settings in such a way that it does not autoformat to this.

src/main/java/be/seeseemelk/mockbukkit/entity/PillagerMock.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Mock implementation of a {@link Pillager}.
+ *
+ * @see IllagerMock
+ */
+public class PillagerMock extends IllagerMock implements Pillager, MockRangedEntity<PillagerMock>
+{
+
+	private final Inventory inventory = new InventoryMock(this, 5, InventoryType.CHEST);
```
Thorinwasher, line 25: I don't know that much about inventories, but from looking into the paper source code I found that the main hand should have a crossbow item in it for pillagers.

Thorinwasher (changes requested):
I found one small inconsistency between the paper implementation and this implementation (To be honest here, that one was kind of hard to find, but I'm always suspicious when the implementation is too similar to the super class)

For more info: I think `net.minecraft.world.entity.Entity#finalizeSpawn()` is always triggered when spawning an entity, this method essentially adds a crossbow to the main hand equipment slot of the entity. The thing is that these differences are only visible in mojang code.

src/test/java/be/seeseemelk/mockbukkit/ServerMockTest.java
4everTheOne: Thanks for the heads up. For some reason in this project I still had the configuration to covert the "java.awt" to a wildcard import. I've removed this and updated the PR.

src/main/java/be/seeseemelk/mockbukkit/entity/IllagerMock.java
Thorinwasher: I could not find anything special in the mojang implementation, so I assume not

4everTheOne: In Paper/Bukkit implemention the Illager (CraftIllager) is working like an abstract class that will be extended by [Evoker](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/Evoker.html), [Illusioner](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/Illusioner.html), [Pillager](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/Pillager.html), [Spellcaster](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/Spellcaster.html), [Vindicator](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/Vindicator.html).

src/main/java/be/seeseemelk/mockbukkit/entity/PillagerMock.java
4everTheOne: Good catch! I totally forgot about the Crossbow.
PR has been updated with it.

Thorinwasher (approved):
Looks good to me

PR #1077, opened 2024-08-16
Plain text component language codes

### LINKED ISSUE #1040: Adventure's PlainTextComponentSerializer does not translate Minecraft Language Codes

### Minecraft Version

1.20

### Describe the bug

When running on a server, using [Minecraft Language Codes](https://gist.github.com/sppmacd/82af47c83b225d4ffd33bb0c27b0d932#:~:text=argument.range.empty) inside translatable Adventure Components is handled automatically. For example, if I have this plugin:

```java

public class BukkitTestPlugin extends JavaPlugin {
    @Override
    public void onEnable() {
        getLogger().info(getMessageString());
    }

    public Component getMessageComponent() {
        return Component.translatable("argument.range.empty");
    }

    public String getMessageString() {
        return PlainTextComponentSerializer.plainText().serialize(getMessageComponent());
    }
}
```

The following message appears in the console:

```log
[BukkitTestPlugin] Expected value or range of values
```

However, when running tests through MockBukkit, these components are serialized as their translation key, in this case `argument.range.empty`.

### Reproducible Test

```java
public class ComponentTest {
    @Test
    void test() {
        MockBukkit.mock();
        BukkitTestPlugin plugin = MockBukkit.load(BukkitTestPlugin.class);

        Assertions.assertEquals("Expected value or range of values", plugin.getMessageString());
        
        MockBukkit.unmock();
    }
}

ComponentTest.test:13 expected: <Expected value or range of values> but was: <argument.range.empty>
```

### Anything else?

I originally asked about this on the [Paper Discord](https://discord.com/channels/289587909051416579/555462289851940864/1259993507658076351). I thought it might be possible to register the translation keys myself into Adventure's [GlobalTranslator](https://docs.papermc.io/paper/dev/component-api/i18n#globaltranslator). The Paper team mentioned that the `en_us.json` file should be present in the server jar, so I tried loading translations like so:

```java
private void setupTranslations() {
	TranslationRegistry registry = TranslationRegistry.create(Key.key("minecraft:english"));

	Class<?> nmsClass;
	try {
		nmsClass = Class.forName("net.minecraft.server.Language");
	} catch (ClassNotFoundException e) {
		// In this case, the code stops here
		throw new IllegalStateException("Could not load translations!", e);
	}

	// I could never find this resource
	try (InputStream input = nmsClass.getResourceAsStream("/assets/minecraft/lang/en_us.json")) {
		if (input == null) throw new IllegalStateException("Could not load translations! InputStream was null.");
		JsonObject jsonObject = new Gson().fromJson(new InputStreamReader(input, StandardCharsets.UTF_8), JsonObject.class);

		for (Map.Entry<String, JsonElement> entry : jsonObject.entrySet()) {
			registry.register(entry.getKey(), Locale.US, new MessageFormat(entry.getValue().getAsString()));
		}
	} catch (IOException e) {
		throw new IllegalStateException("Could not load translations!", e);
	}

	GlobalTranslator.translator().addSource(registry);
}
```

However, I was never able to find the `en_us.json` file, and it seems plausible that this file is simply not included in the classpath when running with MockBukkit.

The Paper team also mentioned that 'PaperAdventure usually hooks into the default Minecraft translations' without using GlobalTranslator (https://discord.com/channels/289587909051416579/555462289851940864/1259995435964039198), so maybe that's a pathway MockBukkit can use. I suppose something about the PlainTextComponentSerializer provided by MockBukkit vs Paper is missing this part.

Thorinwasher:
# Description
Fixes  #1040

~~I might need to add more tests to this. But I honestly don't understand how translatable components work~~ (got one more test from [willkroboth](https://github.com/willkroboth))

# Checklist


src/main/java/be/seeseemelk/mockbukkit/adventure/PlainTextComponentProviderImpl.java
```diff
@@ -0,0 +1,110 @@
...
+				.build();
+	}
+
+	@Override
+	public @NotNull Consumer<PlainTextComponentSerializer.Builder> plainText()
+	{
+		return builder -> builder.flattener(BASIC_COMPONENT_FLATTENER);
+	}
+
+	private static void processTranslatable(TranslatableComponent translatable, Consumer<Component> consumer)
```
thelooter, line 41: Please split this method into multiple methods to reduce cognitive complexity

thelooter (changes requested):

src/main/java/be/seeseemelk/mockbukkit/adventure/PlainTextComponentProviderImpl.java
Thorinwasher: Okay, moved some of the code into a method

src/main/java/be/seeseemelk/mockbukkit/adventure/Language.java
```diff
@@ -0,0 +1,34 @@
+package be.seeseemelk.mockbukkit.adventure;
+
+public interface Language
```
4everTheOne, line 3: Should we add Javadocs to this interface for both "class" and method?

src/main/java/be/seeseemelk/mockbukkit/adventure/Languages.java
```diff
@@ -0,0 +1,46 @@
...
+		return instance;
+	}
+
+	public static void setInstance(Language instance)
+	{
+		Preconditions.checkNotNull(instance);
+		Languages.instance = instance;
+	}
+
+	public static Language loadLanguage(Language.LanguageType language)
```
4everTheOne, line 43: Up for discussion: at the moment a developer using the API only has access to the languages specified in LanguageType. Should we provide a way for users do load their own files?

For example separating this method in 2, one to load from input streams and other to read from resources.

src/main/resources/translations/en-us.json
```diff
@@ -0,0 +1,6886 @@
+{
```
4everTheOne, line 1: Where does this file come from? Is it available in paper?

4everTheOne (approved):

src/main/java/be/seeseemelk/mockbukkit/adventure/Language.java
Thorinwasher: Hmm, I'm not quite sure, should we:
1. Declare this as part of the api
2. Don't declare it as api

Is there any unit tests that would gain something from using this class?

src/main/java/be/seeseemelk/mockbukkit/adventure/Languages.java
Thorinwasher: > Should we provide a way for users do load their own files?

It's kind of possible by using the JsonBackedLanguage class, you mainly have to deserialize the json from file firstly. But we could also add the possibility to read inputstreams (given that they are in the right json format)

src/main/resources/translations/en-us.json
Thorinwasher: Generated by metaminer (legal grayzone), ~~let me make a pr for that actually~~ already done apparently

src/main/java/be/seeseemelk/mockbukkit/adventure/Language.java
Thorinwasher: I wrote some docs, seems like it was necessary

src/main/java/be/seeseemelk/mockbukkit/adventure/Languages.java
Thorinwasher: There we go, added one static methoid for that as well

4everTheOne (approved):

thelooter (approved):

PR #1080, opened 2024-08-17
Implemented methods in EntityMock

4everTheOne:
# Description
This PR implements some methods in Entity that were throwing a UnimplementedOperationException.

# Checklist


src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -646,8 +655,17 @@ protected EntitySubType getSubType()
 	@Override
 	public boolean isOnGround()
 	{
-		// TODO Auto-generated constructor stub
-		throw new UnimplementedOperationException();
+		return this.onGround;
```
Thorinwasher, line 657: Probably up for discussion whether this is best as a boolean switch. To be real, making a real implementation for this would be a nightmare to maintain due to new block types, so I would say this probably is the best choice.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -1262,8 +1272,17 @@ public boolean isInWater()
 	@Override
 	public boolean isInWorld()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return this.inWorld;
```
Thorinwasher: Probably better to check if location.getWorld() is null. 

Generally the entity will be in a world if there is one, see lines 149-152 entity mock

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
Thorinwasher, line 1: The following behaviors are tick dependant:
- frozen ticks
- ticks lived

I would try to link these to the scheduler. What would happen if you added a repeating task like this in the entitty mock constructor?
```java
Bukkit.getScheduler().runTaskRepeatedly(() -> {
    if(isDead){
         return;
     }
    if(isFreezing){
        freezingTick++;
    }
    ticksLived++;
}); // run this task every tick, including the 0th tick  from the moment this entity has spawned
```
I believe every task in the scheduler *should* be cleared whenever the method MockBukkit::unMock() is called. So it should not be an issue for memmory?

This is really up to discussion, as this is getting close to being an only mockbukkit implementation

Thorinwasher (commented):
Here's some comments. I'm being a bit vague here as it's up for discussion

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
4everTheOne: In the beginning of development i was wondering if I should implement it based on the real implementation or not, but then dropped the idea.
One of the reasons behind it, was that identifying if the entity is standing on the block is not trivial. We need to take in consideration:
- Entity location
- How big is the entity? Is the block below the boundary box?
- Is a full block? Does it have small collision boxes (Glass pane)? Is transparent?

From a developer that's trying to test the application using MockBukkit I would prefer just have full control over the entity without needing to make changes to the world.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
4everTheOne: Thanks for the feedback. I will try that approach

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
4everTheOne: That can be interesting, but if done in a way that the developer can control the game time.
Something similar to what exists in the game in "/tick" command.

Lets say for example the developer using MockBukkit is developing a plugin with the following requirement: When a player is hit by a snow ball it should take damage for 5 seconds.

With this "wizard ball" he can "simulate" 5 seconds of the entity being alive and then check the current snapshot of the game.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
Thorinwasher: I agree with that sentiment

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
Thorinwasher: > That can be interesting, but if done in a way that the developer can control the game time.
Something similar to what exists in the game in "/tick" command.

There's `BukkitSchedulerMock#performOneTick()` and `BukkitSchedulerMock#performTicks(long)`. And yeah, I guess one could include damage taken from freezing? (let's not, that seems like a different pr)

4everTheOne: The "ticking" of ticksLived in Mojang code is being done at a server level. Should we implement something similar? For example, each entity having a "tick" method that would be called to increment the ticks lived.

What do you think?
```java
public void tick() {
    if(isDead){
         return;
     }
    if(isFreezing){
        freezingTick++;
    }
    ticksLived++;
}
```

Thorinwasher: That works as well, it's probably better

4everTheOne: I have updated the EntityMock to support the _ticksLived_ counter. After looking into the Mojang source code, it seems that the logic for freezing the entity is in _LivingEntity_ class on method "aiStep" (L3530 - L3538). After this information and if you agree, i will not include this increment in this PR.

Thorinwasher: Sure, that works

Thorinwasher (approved):

Issue #550, opened 2022-07-16
PlayerMock#assertSaid not working correctly with BungeeCord Components


PeachesMLG:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Are you using the latest version of MockBukkit?

- [X] I am using the latest version of MockBukkit.

### Minecraft Version

1.19

### Describe the bug

Looks like my component isnt being asserted correctly, looks like its messing up due to the colour or something.

### Reproducible Test

```java
    @Test
    public void test(){
        PlayerMock playerMock = serverMock.addPlayer();
        TextComponent previousButton = new TextComponent(ChatColor.translateAlternateColorCodes('&', "&c<<"));
        playerMock.spigot().sendMessage(previousButton);
        playerMock.assertSaid(ChatColor.translateAlternateColorCodes('&',"&c<<"));
    }
```

### Anything else?

Side note, since the update we are serialising messages into json chat components which makes failing tests alot harder to debug, would be cool if we could find a way to make it easier to see the differences
![image](https://user-images.githubusercontent.com/26843204/179371441-27e92cf2-b638-4475-9329-acdbe77bd455.png)

Thorinwasher:
Serialize to minimessage first before printing?

4everTheOne:
Hi @PeachesMLG could you please validate if https://github.com/MockBukkit/MockBukkit/pull/1082 fixed the issue?
Thanks in advance.

4everTheOne:
I'll mark this issue as resolved. If someone identifies that it was not resolved yet we can re-open it

PR #1082, opened 2024-08-18
Refactor messaging system 

4everTheOne:
# Description
According to Paper implementation, the only entity that implements sendMessage methods is the Player.
The code was refactored to comply with this behavior, taking into account the least possible impact on the user.
Additionally to this, the support for Conversation API was added.

# Checklist

# Fixes
- https://github.com/MockBukkit/MockBukkit/issues/550


4everTheOne:
Unfortunately I was not able to find a way to properly unit test the _ConversationTracker_ . If someone has an ideia in how to do it, please let me know.

(The implementation is the same as in Paper)

src/main/java/be/seeseemelk/mockbukkit/entity/PlayerMock.java
```diff
@@ -3354,33 +3396,63 @@ public class PlayerSpigotMock extends Player.Spigot
...
+
+		@Override
+		@Deprecated
+		public void sendMessage(@NotNull ChatMessageType position, @Nullable UUID sender, @NotNull BaseComponent... components)
+		{
+			Preconditions.checkNotNull(position, "Position must not be null");
+			Preconditions.checkNotNull(components, "Component must not be null");
+			Component comp = BungeeComponentSerializer.get().deserialize(components);
+			String serialized = LegacyComponentSerializer.legacySection().serialize(comp);
+			comp = LegacyComponentSerializer.legacySection().deserialize(serialized);
```
4everTheOne, line 3454: This is the fix for the issue mentioned in the description.
After deserialize using Bungeecord API we convert it to a String and back to a component to match the expected behaviour.

Thorinwasher:
> Unfortunately I was not able to find a way to properly unit test the _ConversationTracker_ . If someone has an ideia in how to do it, please let me know.
> 
> (The implementation is the same as in Paper)

Yes, as the implementation is exactly the same as paper, it might not be necessary to test this for now. The only reason to add tests here would be if we later on specifically refactored this code, I honestly don't see that happening, and to be real here, I don't think everything needs to be tested. I would just add some simple tests for this that just tests if a class instance is in the right state after given input and say it's good enough.

src/main/java/be/seeseemelk/mockbukkit/conversations/ConversationTracker.java
```diff
@@ -0,0 +1,83 @@
...
+import org.bukkit.Bukkit;
+import org.bukkit.conversations.Conversation;
+import org.bukkit.conversations.ConversationAbandonedEvent;
+import org.bukkit.conversations.ManuallyAbandonedConversationCanceller;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.logging.Level;
+
+public class ConversationTracker
```
Thorinwasher, line 14: Maybe add the @ApiStatus.Internal annotation to this?

Who knows if it needs to be refactored, it's the perfect excuse for us maintainers to not provide backwards compatibility for this

4everTheOne:
This PR has been updated to draft, because of the following:
Comparing CommandSender and Player there's a diference in implementation, the method:
```java
@Override
@Deprecated(forRemoval = true)
public void sendMessage(final @NotNull Identity source, final @NotNull Component message, final @NotNull MessageType type)
{
    Preconditions.checkNotNull(message, "input");
    this.messages.add(message);
}
```
This might be the root cause of the issue, i need to analyse it further.

4everTheOne:
Did not find any other alternative way. This seems the best approach at the moment.

By removing the method mentioned in the previous comment, we lose the information regarding the clickEvent.

src/test/java/be/seeseemelk/mockbukkit/conversations/FakeConversable.java
Thorinwasher, line 1: Can't you use the player mock class instead of this?

This is a mock-mock-implementation (hehe). It's probably better in this case to not test the mock of a mock (whilst mocking), as not doing this would cover more of our code. Or what is the reason behind this?

Thorinwasher (approved):

thelooter:
@4everTheOne could you possibly join the Discord Server? We have a few questions we would like to talk about (don't worry nothing bad ^^)

PR #1083, opened 2024-08-21
Implemented entity FallingBlockMock

4everTheOne:
# Description
Implemented mocks for [Falling Block](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/FallingBlock.html).

# Checklist

# Relates

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/FallingBlockMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/FallingBlockMock.java
@@ -1,0 +1,152 @@
+package be.seeseemelk.mockbukkit.entity;
+
+import be.seeseemelk.mockbukkit.ServerMock;
+import be.seeseemelk.mockbukkit.block.state.BlockStateMock;
+import com.google.common.base.Preconditions;
+import org.bukkit.Material;
+import org.bukkit.block.BlockState;
+import org.bukkit.block.data.BlockData;
+import org.bukkit.entity.FallingBlock;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.UUID;
+
+public class FallingBlockMock extends EntityMock implements FallingBlock
+{
+
+	private final BlockState blockState = new BlockStateMock(Material.SAND);
+
+	private boolean dropItem = true;
+	private boolean canHurtEntities;
+	private boolean autoExpire;
+	private boolean cancelDrop;
+
+	private float damagePerBlock;
+	private int maximumDamage;
+
+	/**
+	 * Constructs a new {@link FallingBlock} on the provided {@link ServerMock} with a specified {@link UUID}.
+	 *
+	 * @param server The server to create the entity on.
+	 * @param uuid   The UUID of the entity.
+	 */
+	public FallingBlockMock(@NotNull ServerMock server, @NotNull UUID uuid)
+	{
+		super(server, uuid);
+	}
+
+	@Override
+	public @NotNull Material getMaterial()
+	{
+		return getBlockData().getMaterial();
+	}
+
+	@Override
+	public @NotNull BlockData getBlockData()
+	{
+		return getBlockState().getBlockData();
+	}
+
+	@Override
+	public void setBlockData(@NotNull BlockData blockData)
+	{
+		Preconditions.checkArgument(blockData != null, "blockData");
+		getBlockState().setBlockData(blockData);
+	}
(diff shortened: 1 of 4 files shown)


thelooter:
Is there a reason this doesn't contain World spawning logic?

4everTheOne:
> Is there a reason this doesn't contain World spawning logic?

I forgot to add that before opening the PR. That was the reason for converting it into a draft.
Now it's ready for review :)

src/main/java/be/seeseemelk/mockbukkit/entity/FallingBlockMock.java
```diff
@@ -0,0 +1,164 @@
...
+	public @NotNull BlockData getBlockData()
+	{
+		return getBlockState().getBlockData();
+	}
+
+	@Override
+	public void setBlockData(@NotNull BlockData blockData)
+	{
+		Preconditions.checkArgument(blockData != null, "blockData");
+		getBlockState().setBlockData(blockData);
```
Thorinwasher: Here's a scenario that would cause weird behavior with this implementation:
- The blockstate is an instance of BarrelMock
- You set the blockdata to, for say to something that is not a BarrelMock related material

After doing this, the blockstate would still be an BarrelMock, and it would now have the wrong type of block data.

~~From looking at the paper implementation, it looks like they are storing the mojang equavelent of BlockData in the falling block.~~

src/main/java/be/seeseemelk/mockbukkit/entity/FallingBlockMock.java
Thorinwasher, line 1: BlockData and BlockState are both mutable. For every getter / setter related to those you need to copy them, so as to not mutate the behavior of FallingBlockMock, whenever one of those classes get's mutated, i.e. this should not be an behavior:
```java
Directional blockData = (Directional) Material.OAK_WALL_SIGN.createBlockData();
fallingBlock.setBlockData(blockData);
blockData.setFacing(BlockFace.UP);
System.out.println(((Directional) fallingBlock.getBlockData()).getFacing());
```
Unwantede output:
```
UP
```
The default facing is expected as an output here

Just to be clear, whenever paper converts from bukkit to mojang classes this exact behavior is replicated, as the conversion means the creation of a new class instance.

Thorinwasher (changes requested):

src/main/java/be/seeseemelk/mockbukkit/entity/FallingBlockMock.java
4everTheOne: If i'm not mistaken that's what happening here, but please let me know if i'm missing something:

I tested using this test and it succeeded:
```java
@Test
void test2() {
  Directional blockData = (Directional) Material.OAK_WALL_SIGN.createBlockData();
  fallingBlock.setBlockData(blockData);

  assertEquals(BlockFace.NORTH, ((Directional) fallingBlock.getBlockData()).getFacing());
  blockData.setFacing(BlockFace.EAST);
  assertEquals(BlockFace.NORTH, ((Directional) fallingBlock.getBlockData()).getFacing());
}
```

4everTheOne: I've updated the unit tests to include an assertion for this

src/main/java/be/seeseemelk/mockbukkit/entity/FallingBlockMock.java
4everTheOne: Can you validate if commit [bc27d7c](https://github.com/MockBukkit/MockBukkit/pull/1083/commits/bc27d7c6d433ae8d07bbf9584e1eaba157324fc1) fixes the issue?
Thanks

src/main/java/be/seeseemelk/mockbukkit/entity/FallingBlockMock.java
Thorinwasher: > If i'm not mistaken that's what happening here, but please let me know if i'm missing something

Yep, that's right. I honestly just missed that

Thorinwasher (approved):
Looks good to me

src/main/java/be/seeseemelk/mockbukkit/entity/FallingBlockMock.java
Thorinwasher: That probably fixes the issue

PR #1089, opened 2024-08-25
feat(entities): Implement endersignal entity

thelooter:
# Description

# Checklist


src/main/java/be/seeseemelk/mockbukkit/entity/EnderSignalMock.java
```diff
@@ -0,0 +1,108 @@
...
+	@Override
+	public void setTargetLocation(@NotNull Location location, boolean update)
+	{
+		Preconditions.checkArgument(location != null, "Location cannot be null");
+		Preconditions.checkArgument(this.getWorld().equals(location.getWorld()), "Cannot target EnderSignal across worlds");
+		this.targetLocation = location;
+
+		if (update)
+		{
+			//ignored
```
4everTheOne, line 61: Should we implement this part of the code?
- Update life back to 0
- Set the value for "survive after death" to false given a 1/5 chance

src/main/java/be/seeseemelk/mockbukkit/entity/EnderSignalMock.java
```diff
@@ -0,0 +1,108 @@
...
+	@Override
+	public void setItem(@Nullable ItemStack item)
+	{
+		if (item == null)
+		{
+			this.item = new ItemStack(Material.ENDER_EYE);
+			return;
+		}
+
+		this.item = item;
```
4everTheOne: Being picky here:
We should store a copy of the ItemStack to keep the behavior similar to `CraftItemStack.asNMSCopy(item)`

src/main/java/be/seeseemelk/mockbukkit/entity/EnderSignalMock.java
```diff
@@ -0,0 +1,108 @@
...
+	{
+		this.setTargetLocation(location, true);
+	}
+
+	@Override
+	public void setTargetLocation(@NotNull Location location, boolean update)
+	{
+		Preconditions.checkArgument(location != null, "Location cannot be null");
+		Preconditions.checkArgument(this.getWorld().equals(location.getWorld()), "Cannot target EnderSignal across worlds");
+		this.targetLocation = location;
```
4everTheOne: Looking into the implementation, seems like only the block coordinates are stored. Maybe we should do a toBlockLocation before:
```suggestion
		this.targetLocation = location.toBlockLocation();
```

src/test/java/be/seeseemelk/mockbukkit/entity/EnderSignalMockTest.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Test
+    void testGetItemDefault()
+    {
+        assertEquals(Material.ENDER_EYE, enderSignal.getItem().getType());
+    }
+
+    @Test
+    void  testSetItem()
```
4everTheOne: Just a small detail:
```suggestion
    void testSetItem()
```

src/main/java/be/seeseemelk/mockbukkit/entity/EnderSignalMock.java
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	public EnderSignalMock(@NotNull ServerMock server, @NotNull UUID uuid)
+	{
+		super(server, uuid);
+	}
+
+	@Override
+	public @NotNull Location getTargetLocation()
+	{
+		Preconditions.checkState(targetLocation != null, "Target location has not been set!");
```
4everTheOne: This precondition does not seem to match the current implementation in Bukkit. Maybe we could initialize the variable target location like this:
```java
private Location targetLocation = new Location(null, 0, 0, 0);
```

What do you think?

4everTheOne (changes requested):
Overall looks good! Just have some suggestions, and please let me know what do you think.

src/main/java/be/seeseemelk/mockbukkit/entity/EnderSignalMock.java
thelooter: I didn't implement this yet because we still need to set a proper policy on handling randomness

src/main/java/be/seeseemelk/mockbukkit/entity/EnderSignalMock.java
thelooter: Yes this doesn't match the implementation exactly, but it is meant as a notice to the user that this is infact user error and a Mockbukkit quirk. I should probably write a clearer Error message

src/main/java/be/seeseemelk/mockbukkit/entity/EnderSignalMock.java
Thorinwasher: > we still need to set a proper policy on handling randomness

I think the policy should be to not have any randomness. In this case I don't see the property `survive after death` to be affecting our mock anyhow, I therefore don't think it matters in this scenario, but... here's some examples of a policy
#### Example 1
Here's an example which should be easy to understand / use
```java
void setSurviveAfterDeathUpdateValue(boolean surviveAfterDeathUpdateValue)  {
    this.surviveAfterDeathUpdateValue = surviveAfterDeathUpdateValue;
}
```
```java
if(update) {
    // other stuff
    this.surviveAfterDeath = surviveAfterDeathUpdateValue;
}
```

#### Example 2
Make an interface which can provide a generic value (we could really just use the `Supplier` interface here). In comparison to example 1 this one allows for more complicated constructions, you could for example have a cyclic behavior, but I have no clue if there's any use for that.
```java
private DeterministicValue<Boolean> surviveAfterDeathUpdateValue = DeterministicValue.of(false);

void setSurviveAfterDeathUpdateValue(DeterministicValue<Boolean> surviveAfterDeathUpdateValue) {
    this.surviveAfterDeathUpdateValue = surviveAfterDeathUpdateValue;
}
```
```java
if(update) {
    // other stuff
    this.surviveAfterDeath = surviveAfterDeathUpdateValue.fetch();
}

thelooter: I like the second one a whole lot more tbh

Thorinwasher: Okay, then that probably goes into a separate pr. Where should the policy be written btw, or just keep it informally?

4everTheOne (approved):
Looks good to me!

PR #1106, opened 2024-09-02
Life cycle event manager implementation

### LINKED ISSUE #1064: Implement MockLifecycleEventManager.registerEventHandler

### Describe the feature

I'm working on a command framework and would like to mock the registration of commands using the `LifecycleEventManager` as described here https://docs.papermc.io/paper/dev/commands

### LINKED ISSUE #1352: Add support for the MessageComponentSerializer

### Describe the feature

The [MessageComponentSerializer](https://jd.papermc.io/paper/1.20.6/io/papermc/paper/command/brigadier/MessageComponentSerializer.html) was added to the paper api in 1.20.6 and I'd like a mock for it to properly test my [custom argument types](https://docs.papermc.io/paper/dev/command-api/basics/custom-arguments).

See #586

Thorinwasher:
# Description
Fixes #1064

One note currently is that I don't know exactly how to test this implementation, this pr is therefore a draft at the moment.

# Checklist


Thorinwasher:
I just realized how nested the lifecycle event stuff is. I will put this pr on hold until I become smart enough to figure this out (If anyone has a clue here, just make a pr to this pr or something)

EnriqueGambra:
Any updates on this PR? I would love to see this get merged in eventually :)

Thorinwasher:
> Any updates on this PR? I would love to see this get merged in eventually :)

The logic for life cycle event stuff is just so nested. It will probably take a while until I get back to this, I honestly can't find the full implementation

CronixZero:
It probably isn't possible to implement this for at least the command registration, right? Currently, Paper Plugins using this functionality aren't able to be tested

Thorinwasher:
> It probably isn't possible to implement this for at least the command registration, right?

Implementing lifecycle event command registration is/was the main purpose of this pull request. That's where I kind of gave up on it with the specified reason. I will go back and take a look on it sooner or later.

src/main/java/org/mockbukkit/mockbukkit/command/brigadier/DummyCommandSender.java
```diff
@@ -0,0 +1,147 @@
...
+import org.bukkit.plugin.Plugin;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.mockbukkit.mockbukkit.MockBukkit;
+
+import java.util.Set;
+import java.util.UUID;
+
+public class DummyCommandSender implements CommandSender
+{
```
4everTheOne, line 18: Can you include a comment in the empty methods please?

src/main/java/org/mockbukkit/mockbukkit/command/brigadier/bukkit/BukkitBrigadierForwardingMapMock.java
```diff
@@ -0,0 +1,397 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.Spliterator;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+public class BukkitBrigadierForwardingMapMock extends HashMap<String, Command>
```
4everTheOne: Should we extend the HashMap? Can't we use more generic type like AbstractMap or Map interface?

src/main/java/org/mockbukkit/mockbukkit/exception/PaperTagListenerManagerMock.java
4everTheOne, line 1: Is this file needed? Looks like it's empty

src/main/java/org/mockbukkit/mockbukkit/plugin/PluginMock.java
```diff
@@ -65,7 +66,7 @@ private Builder()
 		 * @param onLoad What to run on load
 		 * @return This builder
 		 */
-		public Builder withOnLoad(@NotNull Runnable onLoad)
```
4everTheOne, line 68: This change will break the retro compatibility, due to method signature change. Should we keep a copy from the old function signature (marked as deprecated) and redirect to the new method?

4everTheOne (changes requested):
Additionally to the previous comments, can you please take a look at the issues reported by sonar with severity Blocker and High? Thank you 🙂

src/main/java/org/mockbukkit/mockbukkit/command/brigadier/bukkit/BukkitBrigadierForwardingMapMock.java
Thorinwasher: Seems like the only methods that needs to be implemented when switching from hashmap is the equals method and to string method. Should we switch?

4everTheOne: I think so :)

4everTheOne (commented):
Sorry for the delay in the response.
Seems good to me! Just looks like we have a unit test failing

Thorinwasher:
> Sorry for the delay in the response. Seems good to me! Just looks like we have a unit test failing

No problem there, we all have other missions in life as well 😛. Seems like the test were failing because of an API behavior change. I also changed bukkit brigadier forwarding map to now inherit from AbstractMap

Thorinwasher:
Another note is that I have tested this locally on a plugin for the command system, it seems to function properly. But that got me thinking, this command executor that is currently used follows the behavior of a paper server, i.e. when a command fails, it sends the exception message only to the command executor if it is an instance of a brigadier exception (that I've forgotten the name of).

There should actually be some usecases with implementing our own command executor, that forwards the exception to the test, where you can simply do simple tests like this:
```java
registerCommand();
assertThrows(MyBrigadierCommandException.class, () -> executor.performCommand("invalid command"));
assertDoesNotThrow(() -> executor.performCommand("valid command"));
```
That's definitely for a different pr though

4everTheOne (approved):

PR #1107, opened 2024-09-02
Create new `Build` workflow without unused scripts

Thorinwasher:
# Description
I just saw that the readme `build` sign does not show any relevant information at all; it just shows the status of the latest workflow run which is run on every pr change. I think these changes would make it clearer, if even make the pull requests compile slightly faster as I removed some steps for the `Build` workflow

# Checklist


4everTheOne:
I like this approach, it makes the pipelines cleaner!

Edit: The only downside related with this, is that we have repeated steps in both the pipeline, but I don't think it's an issue

Thorinwasher:
> I like this approach, it makes the pipelines cleaner!
> 
> Edit: The only downside related with this, is that we have repeated steps in both the pipeline, but I don't think it's an issue

Yep, I had to repeat some things there, as I'm scared of doing any real changes on the workflow, and it's probably also a good idea to run tests when pushing to main as well anyhow.

Extra: there's also the good old DRY(UINTM) ((don't repeat yourself unless it's not that much)). We could technically reduce the size of the script that initiates gradle stuff, ~~but that means actually touching the workflow, which is a big nono~~ meh let's do that (there already exist premade step that does this)

Thorinwasher:
There we go, I decreased code duplication by using an already premade step for gradle caches. I noticed that the cache does not seem to cache the gradle distribution, with this new current build pipeline. This might be because the caches can by default only be modified by pipelines triggered from the main branch.

thelooter (approved):

PR #1112, opened 2024-09-04
Implemented collidesAt and wouldCollideUsing

4everTheOne:
# Description
Implements the collision detection in entity.

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
@@ -1399,15 +1399,13 @@ public abstract class EntityMock extends Entity.Spigot implements Entity, Messag
 	@Override
 	public boolean collidesAt(@NotNull Location location)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return getBoundingBox().contains(location.toVector());
 	}
 
 	@Override
 	public boolean wouldCollideUsing(@NotNull BoundingBox boundingBox)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return getBoundingBox().overlaps(boundingBox);
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -1399,15 +1399,13 @@ public boolean isInPowderedSnow()
 	@Override
 	public boolean collidesAt(@NotNull Location location)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
```
Thorinwasher, line 1403: I think the paper implementation has some more details to it. For example if entity is instance of armorstand and this paper config property is false `entities.armorStands.doCollisionEntityLookups`, return false. (Do we even have an paper config mock?)

I think there's some more things to this, but I honestly can't comprehend it, seems tideous to mock properly. Something something about checking against blocks and world border.

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
```diff
@@ -1399,15 +1399,13 @@ public boolean isInPowderedSnow()
...
+		return getBoundingBox().contains(location.toVector());
 	}
 
 	@Override
 	public boolean wouldCollideUsing(@NotNull BoundingBox boundingBox)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return getBoundingBox().overlaps(boundingBox);
```
Thorinwasher, line 1408: Same can be said about this implementation as with the previous comment

src/main/java/be/seeseemelk/mockbukkit/entity/EntityMock.java
4everTheOne: That's correct, but at the moment we don't have a mock for _PaperConfigurations_, only for _ServerConfiguration_.
How should we address this?

Thorinwasher: > That's correct, but at the moment we don't have a mock for _PaperConfigurations_, only for _ServerConfiguration_. How should we address this?

Let's  not address this then // just ignore that behavior.

Thorinwasher (approved):

PR #1115, opened 2024-09-07
Implemented logic for launch projectile in living entity

4everTheOne:
# Description
Implemented logic to throw a projectile from a living entity

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
@@ -1678,8 +1678,9 @@ public class WorldMock implements World
 	@Override
 	public <T extends Entity> @NotNull T createEntity(@NotNull Location location, @NotNull Class<T> aClass)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		EntityMock entity = this.mockEntity(location, aClass, false);
+		Preconditions.checkState(aClass.isInstance(entity), "Entity {} is not an instance of {}.", entity.getClass().getName(), aClass.getName());
+		return aClass.cast(entity);
 	}
 
 	@Override
--- a/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
@@ -143,7 +143,6 @@ import org.bukkit.WorldCreator;
 import org.bukkit.WorldType;
 import org.bukkit.block.Biome;
 import org.bukkit.block.Block;
-import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
 import org.bukkit.block.data.BlockData;
 import org.bukkit.block.data.Waterlogged;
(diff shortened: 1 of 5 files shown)


src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
```diff
@@ -1118,41 +1123,42 @@ else if (clazz == Firework.class)
...
+//				break;
+//			}
+//			if (spawnFace == BlockFace.SELF)
+//			{
+//				spawnFace = BlockFace.SOUTH;
+//			}
+//			spawnFace = spawnFace.getOppositeFace();
+//			// TODO: Spawn entities here.
+//			throw new UnimplementedOperationException();
+//		}
```
4everTheOne: This code was the only part that depended on the location directly, and it seems incomplete. I only see UnimplementedOperationException being trown. I commented it out, since it's important code but that was not finished yet

Thorinwasher: Generally commenting out code is a bad practice. I'm not quite sure why, It might be because it gets harder to read the code that is actually doing things. (I will look into this pull request now)

src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -318,19 +319,38 @@ public void registerAttribute(@NotNull Attribute attribute)
...
 	{
 		Preconditions.checkNotNull(projectile, "Projectile cannot be null");
-		T entity = launchProjectile(projectile);
+
+		World world = getWorld();
+		Preconditions.checkNotNull(world, "World cannot be null when launching a projectile");
+
+		// The throwable spawn location is slightly lower than the eye level (See: ThrowableProjectile constructor)
+		Location spawnLocation = getEyeLocation().subtract(0, 0.10000000149011612D, 0);
+		T entity = world.createEntity(spawnLocation, projectile);
```
Thorinwasher, line 341: Just to be sure here, can every class that extends `Projectile` be spawned in the world using createEntity?

src/main/java/be/seeseemelk/mockbukkit/WorldMock.java
Thorinwasher: Seems like paper has moved over to have a full data-based implementation now. When I look at it, this method in mockbukkit is about 450 lines of if statements, this should be fixed for our own sake. I will think about how to do this. But that's a separate pr

Thorinwasher (changes requested):
I think it's best to remove commented out code completely. If we need to recover it, then we can just do some git magic to see how it was

src/main/java/be/seeseemelk/mockbukkit/entity/LivingEntityMock.java
4everTheOne: I will update the unit tests to include this, but some of them are not implemented yet:
image

This is probably related with https://github.com/MockBukkit/MockBukkit/issues/825

Thorinwasher: Ah, that looks alright, I was mostly wondering if the "abstract" classes were included, like `AbstractArrow`, which should redirect to another class.

Anyhow, that is for a separate pr if that's an issue anyhow.

Thorinwasher (approved):

PR #1126, opened 2024-09-16
Implemented mocks for Breeze, WindCharge and BreezeWindCharge

4everTheOne:
# Description
Implemented mocks for [Breeze](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/Breeze.html), [WindCharge](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/WindCharge.html) and [BreezeWindCharge](https://jd.papermc.io/paper/1.21.1/org/bukkit/entity/BreezeWindCharge.html).

# Checklist

# Relates

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/BreezeWindChargeMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/BreezeWindChargeMock.java
@@ -35,6 +35,7 @@ public class BreezeWindChargeMock extends AbstractWindChargeMock implements Bree
 		}
 
 		getWorld().createExplosion(this, getLocation(), 3.0F, false);
+		remove();
 	}
 
 	@Override
--- a/src/main/java/be/seeseemelk/mockbukkit/entity/BreezeWindChargeMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/entity/BreezeWindChargeMock.java
@@ -1,0 +1,46 @@
+package be.seeseemelk.mockbukkit.entity;
+
+import be.seeseemelk.mockbukkit.ServerMock;
+import org.bukkit.entity.BreezeWindCharge;
+import org.bukkit.entity.EntityType;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.UUID;
+
+/**
+ * Mock implementation of a {@link BreezeWindCharge}.
+ *
+ * @see AbstractWindChargeMock
+ */
+public class BreezeWindChargeMock extends AbstractWindChargeMock implements BreezeWindCharge
+{
+
+	/**
+	 * Constructs a new {@link BreezeWindChargeMock} on the provided {@link ServerMock} with a specified {@link UUID}.
+	 *
+	 * @param server The server to create the entity on.
+	 * @param uuid   The UUID of the entity.
+	 */
+	public BreezeWindChargeMock(@NotNull ServerMock server, @NotNull UUID uuid)
+	{
+		super(server, uuid);
+	}
+
+	@Override
+	public void explode()
+	{
+		if (!isInWorld())
+		{
+			return;
+		}
+
+		getWorld().createExplosion(this, getLocation(), 3.0F, false);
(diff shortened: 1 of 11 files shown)


src/main/java/be/seeseemelk/mockbukkit/entity/BreezeWindChargeMock.java
```diff
@@ -0,0 +1,46 @@
...
+	@Override
+	public void explode()
+	{
+		if (!isInWorld())
+		{
+			return;
+		}
+
+		getWorld().createExplosion(this, getLocation(), 3.0F, false);
+	}
```
Thorinwasher: I think you forgot to remove the entity and trigger an entity remove event with reason `DISCARD`. See line 15 in `CraftAbstractWindCharge`

src/main/java/be/seeseemelk/mockbukkit/entity/WindChargeMock.java
```diff
@@ -0,0 +1,46 @@
...
+	@Override
+	public void explode()
+	{
+		if (!isInWorld())
+		{
+			return;
+		}
+
+		getWorld().createExplosion(this, getLocation(), 1.2F, false);
+	}
```
Thorinwasher: Same here as with the above comment

src/test/java/be/seeseemelk/mockbukkit/entity/BreezeWindChargeMockTest.java
Thorinwasher, line 1: ~~Possibly add an test for the exploding the entity?~~ (Edit: the explosion of the entity method is not implemented, I will add a different comment for that (edit 2: seems fine, the implementation just return false by some reason))

Like for example if it will trigger an entity explode event (should it?)

src/test/java/be/seeseemelk/mockbukkit/entity/WindChargeMockTest.java
Thorinwasher, line 1: Same here as above comment

Thorinwasher (changes requested):

src/test/java/be/seeseemelk/mockbukkit/entity/BreezeWindChargeMockTest.java
Thorinwasher: But... Could be good to see if entity remove event get's triggered whenever the explosion is triggered.

4everTheOne: I'm not sure how to create a test for this, while the explosion is not implemented yet. This will result in a UnimplementedOperationException in the meantime.

Thorinwasher: Oh, it doesn't just return a boolean value without doing anything more?

src/main/java/be/seeseemelk/mockbukkit/entity/WindChargeMock.java
```diff
@@ -0,0 +1,47 @@
...
+	@Override
+	public void explode()
+	{
+		if (!isInWorld())
+		{
+			return;
+		}
+
+		getWorld().createExplosion(this, getLocation(), 1.2F, false);
+		remove();
```
Thorinwasher: Seems like this method does not trigger a EntityRemoveEvent. This is a separate issue though

4everTheOne: I have updated the method to call the EntityRemoveEvent.

src/test/java/be/seeseemelk/mockbukkit/entity/BreezeWindChargeMockTest.java
4everTheOne: Do you mean the explode method? It does not return anything.

Thorinwasher (approved):

PR #1128, opened 2024-09-24
Fix getEnchantmentLevel

Intybyte:
# Description
Fix getEnchantmentLevel and hashCode

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
@@ -17,6 +17,7 @@ import javax.annotation.Nullable;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 
 public class ItemStackMock extends ItemStack
 {
@@ -306,6 +307,11 @@ public class ItemStackMock extends ItemStack
 				}
 			}
 		}
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(type, amount, itemMeta, durability);
 	}
 
 }
--- a/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
@@ -17,7 +17,6 @@ import javax.annotation.Nullable;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Locale;
 import java.util.Map;
-import java.util.Objects;
 
 public class ItemStackMock extends ItemStack
 {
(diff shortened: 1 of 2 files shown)


4everTheOne:
Hi @Intybyte , thank you for your first contribution!

Could you add some unit tests to cover this methods?

src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -297,4 +309,9 @@ private static void handleLegacyEnchantmentsForDeserialization(@NotNull Map<Stri
 		}
 	}
 
+	@Override
+	public int hashCode() {
+		return Objects.hash(type, amount, itemMeta, durability);
+	}
```
4everTheOne: We should also override equals method

Intybyte: after a bit of talk on the discord  #1058 will cover that part, as it did so way beter than my case, i think i will revert the hashCode change for better merging later on

src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -184,6 +184,17 @@ public void addUnsafeEnchantment(@NotNull Enchantment ench, int level)
 		}
 	}
 
+	@Override
+	public int getEnchantmentLevel(Enchantment ench)
+	{
+		Preconditions.checkArgument(ench != null, "Enchantment cannot be null");
+
+		final ItemMeta meta = this.getItemMeta();
+		Preconditions.checkNotNull(meta, "Meta must not be null");
```
Thorinwasher, line 193: I think the paper implementation just defaults to 0 if there's no meta available. Although that's the case, we have a policy to throw exceptions whenever something is used improperly, so I think this is fine.

src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -184,6 +184,17 @@ public void addUnsafeEnchantment(@NotNull Enchantment ench, int level)
 		}
 	}
 
+	@Override
+	public int getEnchantmentLevel(Enchantment ench)
```
Thorinwasher, line 188: Although there's little to test here, could you just add a test that triggers this method, and checks if it does anything? (in ItemStackMockTest)

Thorinwasher (changes requested):

src/main/java/be/seeseemelk/mockbukkit/inventory/ItemStackMock.java
Intybyte: Alright

Thorinwasher (approved):

PR #1139, opened 2024-09-30
Add a builder for MockPlugin

Thorinwasher:
# Description
I thought, maybe it could possibly be good to add custom onEnable / onDisable / onLoad functionality for the mock plugin. It's at least useful internally. Something like this would be necessary for pull request #1106 as something would need to execute before the plugin is fully enabled (although yeah, you could just make a new plugin for every individual test, but that's not ideal). Please just tell me if this a bad idea, and I will just close this and make a custom class instead.

# Checklist

### DIFF
--- a/src/main/java/be/seeseemelk/mockbukkit/MockPlugin.java
+++ b/src/main/java/be/seeseemelk/mockbukkit/MockPlugin.java
@@ -3,9 +3,11 @@ package be.seeseemelk.mockbukkit;
 import org.bukkit.plugin.PluginDescriptionFile;
 import org.bukkit.plugin.java.JavaPlugin;
 import org.bukkit.plugin.java.JavaPluginLoader;
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
 import java.io.File;
+import java.util.Optional;
 
 /**
  * A simple plugin that does nothing.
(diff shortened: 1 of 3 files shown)


src/main/java/be/seeseemelk/mockbukkit/MockPlugin.java
```diff
@@ -33,4 +35,101 @@ protected MockPlugin(@NotNull JavaPluginLoader loader, @NotNull PluginDescriptio
...
+		private String pluginName = "MockPlugin";
+		private String pluginVersion = "1.0.0";
+
+		private Builder()
+		{
+		}
+
+		public Builder withOnLoad(Runnable onLoad)
+		{
+			this.onLoad = Optional.of(onLoad);
```
4everTheOne, line 68: It seems that we also support null here, if so we should use Optional.ofNullable here, otherwise a NPE is thrown.

src/main/java/be/seeseemelk/mockbukkit/MockPlugin.java
```diff
@@ -33,4 +35,101 @@ protected MockPlugin(@NotNull JavaPluginLoader loader, @NotNull PluginDescriptio
...
+			return this;
+		}
+
+		public Builder withOnEnable(Runnable onEnable)
+		{
+			this.onEnable = Optional.of(onEnable);
```
4everTheOne, line 78: It seems that we also support null here, if so we should use Optional.ofNullable here, otherwise a NPE is thrown.

src/main/java/be/seeseemelk/mockbukkit/MockPlugin.java
```diff
@@ -33,4 +35,101 @@ protected MockPlugin(@NotNull JavaPluginLoader loader, @NotNull PluginDescriptio
...
+			return this;
+		}
+
+		public Builder withOnDisable(Runnable onDisable)
+		{
+			this.onDisable = Optional.of(onDisable);
```
4everTheOne, line 88: It seems that we also support null here, if so we should use Optional.ofNullable here, otherwise a NPE is thrown.

src/main/java/be/seeseemelk/mockbukkit/MockPlugin.java
```diff
@@ -33,4 +35,101 @@ protected MockPlugin(@NotNull JavaPluginLoader loader, @NotNull PluginDescriptio
...
+			return this;
+		}
+
+		public Builder withPluginName(String name)
+		{
+			this.pluginName = name;
```
4everTheOne, line 99: Should we add a Precondition to the name argument and validate that is not null and has text?

src/main/java/be/seeseemelk/mockbukkit/MockPlugin.java
```diff
@@ -33,4 +35,101 @@ protected MockPlugin(@NotNull JavaPluginLoader loader, @NotNull PluginDescriptio
...
+			return this;
+		}
+
+		public Builder withPluginVersion(String version)
+		{
+			this.pluginVersion = version;
```
4everTheOne, line 110: Should we add a Precondition to the version argument and validate that is not null and has text?

4everTheOne (commented):
Just have some small suggestions

4everTheOne:
Also, seems that some tests are failing

thelooter (approved):

PR #1153, opened 2024-10-12
Feature/boss mock

Intybyte:
# Description
Add a boss abstract class and add Wither and EnderDragon to mockable entities

Progresses  #825

# Checklist


thelooter:
Hey, Can you please rebase this and check coverage of your tests?

src/main/java/org/mockbukkit/mockbukkit/entity/EnderDragonMock.java
```diff
@@ -0,0 +1,190 @@
...
+	private Phase phase = Phase.HOVER;
+	private final DragonBattle battle;
+
+	/**
+	 * Constructs a new {@link MonsterMock} on the provided {@link ServerMock} with a specified {@link UUID}.
+	 *
+	 * @param server The server to create the entity on.
+	 * @param uuid   The UUID of the entity.
+	 */
+	public EnderDragonMock(@NotNull ServerMock server, @NotNull UUID uuid)
```
Thorinwasher, line 29: Most of these methods in this anonymous class are not really implemented, yes they don't throw unimplemented operation exception, but their behavior does follow paper behavior.

Either implement them or don't

src/main/java/org/mockbukkit/mockbukkit/entity/EnderDragonMock.java
```diff
@@ -0,0 +1,190 @@
...
+	@Override
+	public int getDeathAnimationTicks()
+	{
+		throw new UnimplementedOperationException();
+	}
+
+	@Override
+	public @NotNull Location getPodium()
+	{
+		return podium;
```
Thorinwasher, line 65: This can return a null value if setPodioum has not been triggered. There should be a default value for this, like for example (0, 0, 0) on the world the entity is in

src/main/java/org/mockbukkit/mockbukkit/entity/WitherMock.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	@Override
+	public boolean canTravelThroughPortals()
+	{
+		return canPortal;
+	}
+
+	@Override
+	public void setCanTravelThroughPortals(boolean b)
```
Thorinwasher: Use proper parameter names

src/main/java/org/mockbukkit/mockbukkit/entity/WitherMock.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	@Override
+	public int getInvulnerabilityTicks()
+	{
+		return invulnerable_ticks;
+	}
+
+	@Override
+	public void setInvulnerabilityTicks(int i)
```
Thorinwasher: Use proper parameter names, for example `invulnerableTicks `

src/main/java/org/mockbukkit/mockbukkit/entity/EnderDragonMock.java
```diff
@@ -0,0 +1,190 @@
...
+			}
+
+			@Override
+			public boolean hasBeenPreviouslyKilled()
+			{
+				return false;
+			}
+
+			@Override
+			public void setPreviouslyKilled(boolean b)
```
Thorinwasher: Use proper parameter names, for example `previouslyKilled`

src/main/java/org/mockbukkit/mockbukkit/entity/EnderDragonMock.java
```diff
@@ -0,0 +1,190 @@
...
+			}
+
+			@Override
+			public @Nullable Location getEndPortalLocation()
+			{
+				return null;
+			}
+
+			@Override
+			public boolean generateEndPortal(boolean b)
```
Thorinwasher: Use proper parameter names, for example  `shouldGenerateEndPortal`

src/main/java/org/mockbukkit/mockbukkit/entity/WitherMock.java
```diff
@@ -0,0 +1,130 @@
...
+import org.mockbukkit.mockbukkit.exception.UnimplementedOperationException;
+import org.mockbukkit.mockbukkit.world.WorldMock;
+
+import java.util.UUID;
+
+public class WitherMock extends AbstractBossMock implements Wither
+{
+
+	private final LivingEntity[] headsTarget = new LivingEntity[] {null, null, null}; // DATA_TARGETS for each head
+	private int invulnerable_ticks = 0; // DATA_ID_INV
```
Thorinwasher: Wrong naming convention, use camel case

src/main/java/org/mockbukkit/mockbukkit/entity/WitherMock.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	@Override
+	public int getInvulnerableTicks()
+	{
+		return invulnerable_ticks;
+	}
+
+	@Override
+	public void setInvulnerableTicks(int i)
```
Thorinwasher: Use proper parameter names

src/main/java/org/mockbukkit/mockbukkit/entity/data/EntityData.java
Thorinwasher, line 1: Reformat this file

src/main/java/org/mockbukkit/mockbukkit/entity/WitherMock.java
Thorinwasher, line 1: Reformat this file

Thorinwasher (changes requested):
Take a look at [coverage](https://app.codecov.io/gh/MockBukkit/MockBukkit/pull/1153/blob/src/main/java/org/mockbukkit/mockbukkit/entity/EnderDragonMock.java) and try to test every method you have implemented, compare it with the paper implementation to see what to expect approximately. Some examples is to test nullability and index out of bounds stuff.

Other than that, I have added some more requested changes, mostly style things. I'm trying to do as explicit a review as possible to decrease the amount cycles until this gets merged, therefore the repeated comments.

src/main/java/org/mockbukkit/mockbukkit/entity/EnderDragonMock.java
Intybyte: Where can i implement a class to handle it properly instead of creating an anonymius object on the fly?

Thorinwasher: > Where can i implement a class to handle it properly instead of creating an anonymius object on the fly?

You can just make an implementation in the entity package, just add mock to the end of the name, and we're set 😄

src/main/java/org/mockbukkit/mockbukkit/entity/DragonBattleMock.java
```diff
@@ -0,0 +1,163 @@
...
+	public @Nullable Location getEndPortalLocation()
+	{
+		return portalLocation;
+	}
+
+	@Override
+	public boolean generateEndPortal(boolean withPortals)
+	{
+		return false;
+	}
```
Thorinwasher: This is the paper implementation:
```java
@Override
    public boolean generateEndPortal(boolean withPortals) {
        if (this.handle.portalLocation != null || this.handle.findExitPortal() != null) {
            return false;
        }

        this.handle.spawnExitPortal(withPortals);
        return true;
    }
```
I recommend just throwing unimplemented operation exception there

src/main/java/org/mockbukkit/mockbukkit/entity/DragonBattleMock.java
```diff
@@ -0,0 +1,163 @@
...
+	@Override
+	public void setPreviouslyKilled(boolean previouslyKilled)
+	{
+		this.previouslyKilled = previouslyKilled;
+	}
+
+	@Override
+	public void initiateRespawn()
+	{
+		respawnPhase = RespawnPhase.START;
```
Thorinwasher, line 78: This does not follow paper behavior, look at `net.minecraft.world.level.dimension.end.EndDragonFight#tryRespawn(BlockPos)`

src/main/java/org/mockbukkit/mockbukkit/entity/DragonBattleMock.java
```diff
@@ -0,0 +1,163 @@
...
+		{
+
+			if (portalLocation == null)
+			{
+				portalLocation = new Location(enderDragonMock.getWorld(), 0, 0, 0);
+			}
+
+			if (enderCrystalCollection != null)
+			{
+				respawnCrystals = enderCrystalCollection.stream().filter(Objects::nonNull).toList();
```
Thorinwasher: Should also filter away any endercrystals that are not in the same world as the dragon battle.

src/main/java/org/mockbukkit/mockbukkit/entity/DragonBattleMock.java
```diff
@@ -0,0 +1,163 @@
...
+	@Override
+	public DragonBattle.RespawnPhase getRespawnPhase()
+	{
+		return respawnPhase;
+	}
+
+	@Override
+	public boolean setRespawnPhase(@NotNull DragonBattle.RespawnPhase respawnPhase)
+	{
+		this.respawnPhase = respawnPhase;
```
Thorinwasher, line 120: Does not follow paper implementation, for example, there's a argument check whether phase is not null and that it is not the `NONE` phase. Other than that, the other logic modifies something that is not currently implemented, i.e. respawn stage. So we can just ignore that

src/main/java/org/mockbukkit/mockbukkit/entity/DragonBattleMock.java
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		gateways--;
+		return true;
+	}
+
+	@Override
+	public void spawnNewGateway(@NotNull Position position)
+	{
+		//this seems to just add a new portal
```
Thorinwasher: Probably a better idea to throw a unimplemented operation exception, as one probably expects there to be structure generation or something

src/main/java/org/mockbukkit/mockbukkit/entity/WitherMock.java
```diff
@@ -0,0 +1,131 @@
...
+	@Override
+	public void enterInvulnerabilityPhase()
+	{
+		this.setInvulnerableTicks(220);
+		this.bossBarMock.setProgress(0.0F);
+		this.setHealth(this.getMaxHealth() / 3.0F);
+	}
+
+	@Override
+	public void rangedAttack(@NotNull LivingEntity livingEntity, float v)
```
Thorinwasher: I think I missed these parameters here, please rename the parameter `v` to something else

src/main/java/org/mockbukkit/mockbukkit/entity/WitherMock.java
```diff
@@ -0,0 +1,131 @@
...
+	{
+		throw new UnimplementedOperationException();
+	}
+
+	@Override
+	public void setChargingAttack(boolean b)
```
Thorinwasher: Parameter name

Thorinwasher (changes requested):
I found some implementation differences between your code and paper code, please take a look at that. Being a bit thorough here, as we don't want the mock implementation to stray away too much from the real implementation.

Other than that, I think you forgot to add tests for dragon battle mock. Code coverage is therefore still a bit too low

Thorinwasher (approved):

PR #1157, opened 2024-10-14
Fix unloadWorld returning false when WorldUnloadEvent is not cancelled

benwoo1110:
# Description

A boolean mishap when checking if WorldUnloadEvent is cancelled during server#unloadWorld

# Checklist


4everTheOne (approved):

PR #1177, opened 2024-11-02
Implementation of `InventoryViewMock::getItem` and `InventoryViewMock::setItem`

svaningelgem:
# Description
Implementation of `InventoryViewMock::getItem` and `InventoryViewMock::setItem`
Closing this would deprecate #1179 (and vice versa)

# Checklist
The following items should be checked before the pull request can be merged.
- [x] Code follows existing style.
- [x] Unit tests added (if applicable).

PR #1179, opened 2024-11-02
Implementation of `InventoryViewMock::getItem` and `InventoryViewMock::setItem` (to 4.0 branch)

svaningelgem:
# Description
As @Thorinwasher mentioned, seems like 4.0 is now the main branch. So pushing my changes to there to.
Closing this would deprecate #1177 (and vice versa)

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
@@ -146,32 +146,29 @@ public abstract class InventoryViewMock implements InventoryView
 	@Override
 	public void setItem(int slot, @Nullable ItemStack item)
 	{
-		int topSize = getTopInventory().getSize();
+		ItemStack stack = item == null ? null : item.clone();
+
 		if (slot < 0)
 		{
-			if (item != null)
+			if (stack != null)
 			{
-				Location location = bottomInventory.getLocation(); // should be the player
-				if (location != null)
-				{
-					location.getWorld().dropItemNaturally(location, item);
-					return;
-				}
+				getPlayer().getWorld().dropItemNaturally(getPlayer().getLocation(), stack);
 			}
 
-			throw new IndexOutOfBoundsException(slot);
+			return;
 		}
 
+		int topSize = getTopInventory().getSize();
 		if (slot < topSize)
 		{
-			getTopInventory().setItem(slot, item);
+			getTopInventory().setItem(slot, stack);
 			return;
 		}
 
 		slot -= topSize;
 		if (slot < getBottomInventory().getSize())
 		{
-			getBottomInventory().setItem(slot, item);
+			getBottomInventory().setItem(slot, stack);
 			return;
 		}
 
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
@@ -145,14 +145,52 @@ public abstract class InventoryViewMock implements InventoryView
 	@Override
 	public void setItem(int slot, @Nullable ItemStack item)
 	{
-		// TODO Auto-generated method stub
+		int topSize = getTopInventory().getSize();
+		if (slot < 0)
+		{
+			// TODO : ???
+			throw new UnimplementedOperationException();
+		}
+
+		if (slot < topSize)
+		{
+			getTopInventory().setItem(slot, item);
+			return;
+		}
+
+		slot -= topSize;
+		if (slot < getBottomInventory().getSize())
+		{
+			getBottomInventory().setItem(slot, item);
+			return;
+		}
+
+		// TODO: ???
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
 	public @Nullable ItemStack getItem(int slot)
 	{
-		// TODO Auto-generated method stub
+		int topSize = getTopInventory().getSize();
+		if (slot < 0)
+		{
+			// TODO : ???
+			throw new UnimplementedOperationException();
+		}
+
+		if (slot < topSize)
+		{
+			return getTopInventory().getItem(slot);
+		}
+
+		slot -= topSize;
+		if (slot < getBottomInventory().getSize())
+		{
+			return getBottomInventory().getItem(slot);
+		}
+
+		// TODO: ???
 		throw new UnimplementedOperationException();
 	}
 
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
@@ -148,8 +148,14 @@ public abstract class InventoryViewMock implements InventoryView
 		int topSize = getTopInventory().getSize();
 		if (slot < 0)
 		{
-			// TODO : ???
-			throw new UnimplementedOperationException();
+			if (item != null)
+			{
+				// This should throw the item on the ground, but for that we need a location. Which is unimplemented for
+				//   now in 'topInventory'...
+				throw new UnimplementedOperationException();
+			}
+
+			return;
 		}
 
 		if (slot < topSize)
@@ -165,8 +171,7 @@ public abstract class InventoryViewMock implements InventoryView
 			return;
 		}
 
-		// TODO: ???
-		throw new UnimplementedOperationException();
+		throw new IndexOutOfBoundsException("Index " + slot + " out of bounds for length " + (topSize + getBottomInventory().getSize()));
 	}
 
 	@Override
@@ -175,8 +180,7 @@ public abstract class InventoryViewMock implements InventoryView
 		int topSize = getTopInventory().getSize();
 		if (slot < 0)
 		{
-			// TODO : ???
-			throw new UnimplementedOperationException();
+			return null;
 		}
 
 		if (slot < topSize)
@@ -190,8 +194,7 @@ public abstract class InventoryViewMock implements InventoryView
 			return getBottomInventory().getItem(slot);
 		}
 
-		// TODO: ???
-		throw new UnimplementedOperationException();
+		throw new IndexOutOfBoundsException("Index " + slot + " out of bounds for length " + (topSize + getBottomInventory().getSize()));
 	}
 
 	@Override
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
@@ -148,14 +148,9 @@ public abstract class InventoryViewMock implements InventoryView
 		int topSize = getTopInventory().getSize();
 		if (slot < 0)
 		{
-			if (item != null)
-			{
-				// This should throw the item on the ground, but for that we need a location. Which is unimplemented for
-				//   now in 'topInventory'...
-				throw new UnimplementedOperationException();
-			}
-
-			return;
+			// This should throw the item on the ground, but for that we need a location. Which is unimplemented for
+			//   now in 'topInventory'...
+			throw new IndexOutOfBoundsException("Index " + slot + " out of bounds for length " + (topSize + getBottomInventory().getSize()));
 		}
 
 		if (slot < topSize)
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
@@ -1,5 +1,6 @@
 package org.mockbukkit.mockbukkit.inventory;
 
+import org.bukkit.Location;
 import org.mockbukkit.mockbukkit.exception.UnimplementedOperationException;
 import com.google.common.base.Preconditions;
 import org.bukkit.entity.HumanEntity;
@@ -148,9 +149,17 @@ public abstract class InventoryViewMock implements InventoryView
 		int topSize = getTopInventory().getSize();
 		if (slot < 0)
 		{
-			// This should throw the item on the ground, but for that we need a location. Which is unimplemented for
-			//   now in 'topInventory'...
-			throw new IndexOutOfBoundsException("Index " + slot + " out of bounds for length " + (topSize + getBottomInventory().getSize()));
+			if (item != null)
+			{
+				Location location = bottomInventory.getLocation(); // should be the player
+				if (location != null)
+				{
+					location.getWorld().dropItemNaturally(location, item);
+					return;
+				}
+			}
+
+			throw new IndexOutOfBoundsException(slot);
 		}
 
 		if (slot < topSize)
@@ -166,7 +175,7 @@ public abstract class InventoryViewMock implements InventoryView
 			return;
 		}
 
-		throw new IndexOutOfBoundsException("Index " + slot + " out of bounds for length " + (topSize + getBottomInventory().getSize()));
+		throw new IndexOutOfBoundsException(slot);
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
```diff
@@ -145,14 +145,52 @@ public void setCursor(@Nullable ItemStack item)
 	@Override
 	public void setItem(int slot, @Nullable ItemStack item)
 	{
-		// TODO Auto-generated method stub
+		int topSize = getTopInventory().getSize();
+		if (slot < 0)
+		{
+			// TODO : ???
+			throw new UnimplementedOperationException();
```
Thorinwasher: Just from a hunch, this should probably throw an IllegalArgumentException.

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
```diff
@@ -145,14 +145,52 @@ public void setCursor(@Nullable ItemStack item)
...
 
 	@Override
 	public @Nullable ItemStack getItem(int slot)
 	{
-		// TODO Auto-generated method stub
+		int topSize = getTopInventory().getSize();
+		if (slot < 0)
+		{
+			// TODO : ???
+			throw new UnimplementedOperationException();
```
Thorinwasher: Same here

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
```diff
@@ -145,14 +145,52 @@ public void setCursor(@Nullable ItemStack item)
...
+
+		slot -= topSize;
+		if (slot < getBottomInventory().getSize())
+		{
+			getBottomInventory().setItem(slot, item);
+			return;
+		}
+
+		// TODO: ???
 		throw new UnimplementedOperationException();
```
Thorinwasher: Illegal argument exception?

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
```diff
@@ -145,14 +145,52 @@ public void setCursor(@Nullable ItemStack item)
...
+		}
+
+		slot -= topSize;
+		if (slot < getBottomInventory().getSize())
+		{
+			return getBottomInventory().getItem(slot);
+		}
+
+		// TODO: ???
 		throw new UnimplementedOperationException();
```
Thorinwasher: Illegal argument exception?

src/test/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMockTest.java
Thorinwasher, line 1: Check for out of bounds behavior // assertThrows

Thorinwasher (changes requested):
I think an illegal argument exception should be thrown whenever the slot is out of bounds. I have not checked this at all against paper code, but this seems like a reasonable guess.

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
svaningelgem: Just tested: it returns null

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
svaningelgem: IndexOutOfBoundsException it seems

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
svaningelgem: Just tested: it returns null

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
svaningelgem: IndexOutOfBoundsException it seems

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
svaningelgem: This is weird: -1 drops the itemstack
```java
    @EventHandler
    public void onInvOpen(InventoryOpenEvent event) {
        event.getView().setItem(-1, new ItemStack(Material.DIAMOND_SWORD));
    }
```

Thorinwasher: Thats hillariously wrong. Let's just throw an index out of bounds exception then, as this is unintended behavior

svaningelgem: @Thorinwasher : I would need some guidance here...

What do I do?
```java
			if (item != null) {
				Location location = topInventory.getLocation();
				if (location != null) {
					// Drop the item in the world at the inventory's location
					location.getWorld().dropItem(location, item);
				}
			}
```

But the getLocation() for the inventory isn't implemented. And if I start that, it's going to get me down the rabbit hole.
What do I do?
I'd suggest to add a comment refering this PR + what needs to be done, but still throw `UnimplementedOperationException`?

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
Thorinwasher: Alright, let's be more consistent on the mockbukkit side and just throw index out of bounds exception.

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
svaningelgem: > Thats hillariously wrong. Let's just throw an index out of bounds exception then, as this is unintended behavior

Maybe, but it's what MC does...
So we should implement it as closely as possible in my opinion.

My 2cts

svaningelgem: Actually, it's correct behaviour:
https://hub.spigotmc.org/javadocs/spigot/org/bukkit/inventory/InventoryView.html#setItem(int,org.bukkit.inventory.ItemStack)

However, they state "it is not required"

Thorinwasher: I totally see your point of view there. But don't forget mockbukkit is to test plugins. For debugging purposes, throwing an exception would help more

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
```diff
@@ -145,15 +146,59 @@ public void setCursor(@Nullable ItemStack item)
...
-		throw new UnimplementedOperationException();
+		int topSize = getTopInventory().getSize();
+		if (slot < 0)
+		{
+			if (item != null)
+			{
+				Location location = bottomInventory.getLocation(); // should be the player
+				if (location != null)
+				{
+					location.getWorld().dropItemNaturally(location, item);
```
Thorinwasher: From looking at the paper code this happens:
```java
this.player.getHandle().drop(stack, false);
```
The item is thrown out of the players head, confirmed from testing. Retain ownership is probably whether it can be picked up instantly, don't know exactly. Anyhow, the items seem to also have a velocity which is relative to the players looking direction.

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
```diff
@@ -145,15 +146,59 @@ public void setCursor(@Nullable ItemStack item)
 	@Override
 	public void setItem(int slot, @Nullable ItemStack item)
```
Thorinwasher, line 147: Item needs to be copied before setting it in the inventory to retain paper parity

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
```diff
@@ -145,15 +146,59 @@ public void setCursor(@Nullable ItemStack item)
...
+		{
+			getBottomInventory().setItem(slot, item);
+			return;
+		}
+
+		throw new IndexOutOfBoundsException(slot);
 	}
 
 	@Override
 	public @Nullable ItemStack getItem(int slot)
```
Thorinwasher, line 179: Just going to note that the item should not be copied here, this implementation is fine

src/test/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMockTest.java
```diff
@@ -90,18 +98,129 @@ void getType_TypeSet_SameReturned()
...
+	}
+
+	@Test
+	void setItemNegativeSlot_WithNullItem()
+	{
+		Player player = server.addPlayer();
+		InventoryMock chest = new ChestInventoryMock(null, 9);
+		view = new PlayerInventoryViewMock(player, chest);
+
+		assertThrows(
```
Thorinwasher: No index out of bounds exception is thrown when the item is null on a negative index. Nothing happens when this is done

src/test/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMockTest.java
Thorinwasher, line 1: Add test to check if the item is copied when setting it in the inventory view. Add test that checks that it is not when getting from the inventory view.

Thorinwasher (changes requested):
If you're going full parity, let's just do it completely full parity. I added some comments that should help with this.

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryViewMock.java
svaningelgem: From what i can see from occasional testing (remember that we have a chest open, so there shouldn't be any velocity).
I'll have a look as well, but from my tests it looks like "DropNatural"

Thorinwasher: This code throws the item from the players head though. 
```java
@EventHandler
void onPlayerInteract(PlayerInteractEvent event) {
    event.getPlayer().getOpenInventory().setItem(- 1, new ItemStack(Material.DIAMOND));
}
```
Also, remember to test from paper if you are not already, eh and yeah there might be a difference depending on inventoryview type, but from a quick search, this does not look like it's the case.

Thorinwasher (approved):

Thorinwasher:
Don't worry about this huge pile of merge issues, you can use the `--onto` flag when rebasing later on

svaningelgem:
> Don't worry about this huge pile of merge issues, you can use the `--onto` flag when rebasing later on

Ok, just ping me when I can start the rebase.

thelooter:

svaningelgem:
@thelooter : done

thelooter:
Please rebase this PR

svaningelgem:
Hi @thelooter , could you explain me why?

I noticed that all PRs are squash merged into main, and I've already merged the latest main branch into this PR to resolve any conflicts. Given this workflow, I'm curious what specific concerns you're hoping to address with a rebase? The squash merge will create a single clean commit on top of v1.21 regardless of the intermediate history.

With a squash merge:
- All commits in this PR will be combined into a single commit anyway
- The commit history of my branch becomes irrelevant since it won't be preserved
- The final squashed commit will be based on the latest v1.21, regardless of how I integrated it into my branch

When I did the merge into my PR branch:
- My branch is already up-to-date with the main branch
- I have already resolved any conflicts
- The PR can be tested in its current state

If you want to review the actual changes, I invite you to click the "Files changed" tab on top instead of reviewing the actual intermediate commits. There you'll see all the changes between the v1.21 & my branch, independent of the intermediate history.

Let me know if you're seeing any particular issues I should address differently.

Thorinwasher:
> I noticed that all PRs are squash merged into main, and I've already merged the latest main branch into this PR to resolve any conflicts. Given this workflow, I'm curious what specific concerns you're hoping to address with a rebase? The squash merge will create a single clean commit on top of v1.21 regardless of the intermediate history.

That's true, the history will not be messed up, but now the pr looks borked. I also bet that this would have taken much lesser time to achieve if you would have just rebased initially.

Anyhow with that, @thelooter , should we merge?

svaningelgem:
Actually, from experience: a merge takes much less time than a rebase. For a rebase it's going to replay every commit on top of the target branch. And I need to fix the merge conflicts for *every single commit*... For a merge I only need to do this once.
Maybe it's just IntelliJ's implementation.

But at least in my opinion that doesn't matter when you squash merge, and you just use the "files changed" review feature in github...

Thorinwasher:
> Actually, from experience: a merge takes much less time than a rebase. For a rebase it's going to replay every commit on top of the target branch. And I need to fix the merge conflicts for _every single commit_... For a merge I only need to do this once. Maybe it's just IntelliJ's implementation.
> 
> But at least in my opinion that doesn't matter when you squash merge, and you just use the "files changed" review feature in github...

No, the process would have been automatic, at least if you would have used the `--onto` flag, which I recommended. I have done this for other pr's that had this exact same situation, and it was completely automatic. I can agree with the sentiment that rebases are not necessary to do all the time, but in this scenario it would have been a much more superior alternative.

thelooter:
> Actually, from experience: a merge takes much less time than a rebase. For a rebase it's going to replay every commit on top of the target branch. And I need to fix the merge conflicts for _every single commit_... For a merge I only need to do this once. Maybe it's just IntelliJ's implementation.
> 
> But at least in my opinion that doesn't matter when you squash merge, and you just use the "files changed" review feature in github...

The main issue is local commit logs. Some people e.g. use automatic scripts to automate some stuff about working with git. Having a non clean history on a branch will majorly impact the usefulness of these scripts. A rebase when done properly, should not take much longer than a merge, and usually results in a more conscious decision about changes.

We don't have this written down as a required process yet, but it's been an unwritten rule for quite a while. 

Also, given the nature of this project, its fairly common to go back to PRs to check what changes have been made. Having the individual commits and their commit message can help greatly understanding the thought process and reasoning of the author. Having a cluttered / unclean history like in this PR currently, greatly impairs the ability to do so efficiently.

svaningelgem:
Thanks for taking the time to write this!
I now better understand why you would want a rebase.

Anyway, I tried it via the IDE, and lo and behold... *every*. *single*. *commit* I had to do a merge conflict resolution. So I stopped after 2. I recreated the branch and did a cherry pick. No conflicts and everything still seemed correct.
I additionally tried via the cmdline with --onto, but that seemed to fail too (but no conflict resolutions) because the branch history didn't seem to reflect my own commits at all.

I see that the build started to fail now, so I'll look into that one.

Thorinwasher:
> Thanks for taking the time to write this! I now better understand why you would want a rebase.
> 
> Anyway, I tried it via the IDE, and lo and behold... _every_. _single_. _commit_ I had to do a merge conflict resolution. So I stopped after 2. I recreated the branch and did a cherry pick. No conflicts and everything still seemed correct. I additionally tried via the cmdline with --onto, but that seemed to fail too (but no conflict resolutions) because the branch history didn't seem to reflect my own commits at all.
> 
> I see that the build started to fail now, so I'll look into that one.

That's unfortunate to hear. For more more information, if you use the onto flag, you will disregard all the commits before the selected commit including the selected commit during rebase. I suspect you took the latest commit from your own history, which will disregard all of your work, but I might be wrong.

Anyhow with that, it looks like we are ready to merge

Thorinwasher (approved):
Looks good to me

PR #1183, opened 2024-11-03
Fix data races in scheduler, plugin manager and player list

frengor:
# Description
This fixes data races and race conditions in the scheduler, plugin manager and player list.

As discussed on Discord, this also makes internal-only methods in `ScheduledTask` and `RepeatingTask` classes `protected` and `@ApiStatus.Internal`.

# Checklist
- [x] Unit tests added (not really applicable here I think).

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/PlayerListMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/PlayerListMock.java
@@ -1,5 +1,6 @@
 package org.mockbukkit.mockbukkit;
 
+import com.google.common.collect.Iterators;
 import org.mockbukkit.mockbukkit.ban.IpBanListMock;
 import org.mockbukkit.mockbukkit.ban.ProfileBanListMock;
 import org.mockbukkit.mockbukkit.entity.OfflinePlayerMock;
@@ -15,12 +16,12 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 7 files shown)


frengor:
~I don't know what's wrong with GitHub Actions.~
Edit: fixed now, thanks!

Thorinwasher:
Looks good to me, just going to wait with this until we completely switch to 4.0.

I'm clueless here, but is there any way to test this. Raceconditions seems from my point of view not possible to test using unit tests

frengor:
> I'm clueless here, but is there any way to test this. Raceconditions seems from my point of view not possible to test using unit tests

I don't think there is any sane way to test it. The best I can think of is to call the methods from a bunch of different threads and hope that if anything's wrong it fails the test &mdash; which may not even happen.

thelooter:
> > I'm clueless here, but is there any way to test this. Raceconditions seems from my point of view not possible to test using unit tests
> 
> I don't think there is any sane way to test it. The best I can think of is to call the methods from a bunch of different threads and hope that if anything's wrong it fails the test — which may not even happen.

You can somewhat test for it by using `@RepeatedTest`

frengor:
> You can somewhat test for it by using `@RepeatedTest`

Yes, but not really IMO. `@RepeatedTest` is basically equivalent to running the tests multiple times manually. It would surely help to try the tests with different scheduling, however that is not the main issue when testing race conditions.

Thorinwasher:
Don't worry about this huge pile of merge issues, you can use the `--onto` flag when rebasing later on

Thorinwasher:
> Yes, but not really IMO. @RepeatedTest is basically equivalent to running the tests multiple times manually. It would surely help to try the tests with different scheduling, however that is not the main issue when testing race conditions

Yeah, it's not a reliable unit test, as race conditions will always have this randomness to it. I agree, this is more for manual testing, where you stress test methods by repeating the same operation thousands of times on two different threads. We could actually possibly make a specialized repository for this, that tests mockbukkit automatically, but with a much larger test-time, like multiple minutes or hours.

thelooter:

thelooter:
Please rebase this PR

frengor:
> Please rebase this PR

Done!

Thorinwasher (approved):

PR #1184, opened 2024-11-04
Make OSSRH token username secret

Thorinwasher:
# Description
It's currently accessible in the workflow, I'd say it's better practice to make this username secret. We have already exposed the current one, but we will completely change domain and all.

# Checklist
- [n/a] Code follows existing style.


Thorinwasher:
This will already be fixed in #1185 , I will close this

PR #1188, opened 2024-11-07
Patches for ItemMetaMock and ItemStackMock

alexcrea:
# Description
Various patches for classes related to ItemMetaMock and ItemStackMock

- Ensure meta is copied on setItemMeta and getItemMeta
- Renamed EnchantedBookMetaMock to EnchantmentStorageMetaMock
- ~~Recreated EnchantedBookMetaMock and mark it as deprecated for removal~~ now deleted again
- Removed EnchantmentStorageMeta from unimplemented meta list 
(so ENCHANTED_BOOK Item can have EnchantmentStorageMetaMock as meta and not the default ItemMetaMock)
- Replaced use of LeatherArmorMetaMock by ColorableArmorMetaMock 
(and added equal and hashcode to ColorableArmorMetaMock)

changes added after setting this pr to draft:
- Make ItemMetaMock damage nullable to implement hasDamageValue
- Only allow positive damage value
- Avoid stack overflow on ItemStackMock hashcode and return -1 durrability for AIR item
- Make ItemStackMock durrability start at 0 
(bukkit concept of ItemStacks durrability is "used durrability" (aka damage) and not "durrability remaining")
- Made ItemMetaMock damage impact ItemStackMock durrability and viceverca
- Keep maximum common data of the ItemMeta when item type change or setItemMeta 
- Fixed FixItemFactory not giving correct Meta type
- Implement LeatherArmorMeta.isDyed, ItemMetaMock.removeEnchantments
- Fixed/splited some tests

# Checklist


src/main/java/org/mockbukkit/mockbukkit/inventory/meta/EnchantedBookMetaMock.java
thelooter, line 1: We should either fully remove this class with 4.0 or leave the logic intact

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/EnchantmentStorageMetaMock.java
```diff
@@ -0,0 +1,191 @@
...
+	{
+		super(meta);
+
+		this.storedEnchantments = new HashMap<>(meta.getStoredEnchants());
+	}
+
+	@Override
+	public int hashCode()
+	{
+		final int prime = 31;
```
thelooter: Where does this magic number come from? Please add explanatory comment

src/test/java/org/mockbukkit/mockbukkit/entity/ThrownPotionMockTest.java
```diff
@@ -104,7 +104,7 @@ void getPotionMeta_GivenValidPotion(Material input)
 
 		@NotNull PotionMeta actual = potion.getPotionMeta();
 		assertNotNull(actual);
-		assertSame(actual, potion.getPotionMeta());
+		assertEquals(actual, potion.getPotionMeta());
```
thelooter: If these two shouldn't be the same, that should also be checked with `assertNotSame`

thelooter (approved):

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/EnchantmentStorageMetaMock.java
alexcrea: The class was renamed from EnchantedBookMetaMock in https://github.com/MockBukkit/MockBukkit/pull/1188/commits/5ef13bce771aa90247c24e74c9481890445f24dc. And probably shown as new here as I later recreated EnchantedBookMetaMock
This function was untouched by me from EnchantedBookMetaMock

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/EnchantedBookMetaMock.java
alexcrea: EnchantedBookMetaMock is simply not applied as meta of enchanted book currently.
So deleting it would only break the logic if someone do not expect an enchanted storage meta for a enchanted book (why ?) or instantiate EnchantedBookMetaMock for their testing. 
Deleting it would be something acceptable for a major release. Even if I had preferred to warn users with deprecated annotation before.

thelooter:

src/test/java/org/mockbukkit/mockbukkit/inventory/ItemStackMockTest.java
```diff
@@ -317,4 +323,237 @@ void getEnchantmentLevel_GivenCustomEnchantment()
...
+		ItemMeta meta3 = itemStack.getItemMeta();
+		assertNotEquals(meta1, meta3);
+	}
+
+	@Test
+	void getItemMeta_DamageEmpty(){
+		ItemStackMock itemStack = new ItemStackMock(Material.DIAMOND_PICKAXE);
+		ItemMeta meta = itemStack.getItemMeta();
+
+		assertTrue(meta instanceof Damageable);
```
Thorinwasher: There's apparently `assertInstanceOf` which will give a more informative test failure. Here's a regex 😛  `assertTrue\(.+ instanceof .+\)`

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -127,15 +145,43 @@ public boolean isEmpty()
 	@Override
 	public boolean setItemMeta(@org.jetbrains.annotations.Nullable ItemMeta itemMeta)
```
Thorinwasher: Seems like we have imported javax nullable anotations, mostly a note to myself to make an issue for this, we should use the jetbrains annotations as far as I know for  everything. This is out of scope of this pr though

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -153,13 +199,19 @@ public int getMaxStackSize()
...
+
 		return (short) Math.max(this.durability, 0);
 	}
 
 	@Override
 	public void setDurability(short durability)
 	{
+		short oldDurability = this.durability;
 		this.durability = (short) Math.min(Math.max(durability, 0), this.type.getMaxDurability());
+		if(this.itemMeta != null && (this.itemMeta instanceof Damageable damageable) && this.durability != oldDurability){
```
Thorinwasher: You don't need to make a not null check here, that operation is included in the instanceof operation. For example:
```java
null instanceof Object
```
Will return false

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -212,8 +264,8 @@ public static ItemStackMock empty()
 		ItemStackMock clone = new ItemStackMock(this.type);
 
 		clone.setAmount(this.amount);
+		clone.setItemMeta(this.itemMeta == null ? null : this.itemMeta);
```
Thorinwasher: Yout don't need to do a ternary statement here
```suggestion
clone.setItemMeta(this.itemMeta);
```

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/EnchantmentStorageMetaMock.java
Thorinwasher, line 1: This is technically a breaking change. But now that I think about it, this is necessary for the correct serialization. @thelooter do we have policy for breaking changes? Like this one is necessary, I believe. Well the alternative is to just keep the old definition `EnchantedBookMetaMock`, but also have a new definition `EnchantmentStorageMetaMock`.

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/FireworkMetaMock.java
```diff
@@ -35,11 +36,14 @@ public FireworkMetaMock()
...
 	 */
-	public FireworkMetaMock(@NotNull FireworkMeta meta)
+	public FireworkMetaMock(@NotNull ItemMeta meta)
 	{
 		super(meta);
 
-		this.effects.addAll(meta.getEffects());
+		if(meta instanceof FireworkMeta fireworkMeta)
+		{
+			this.effects.addAll(fireworkMeta.getEffects());
```
Thorinwasher, line 45: This is not copying the power of the firework, is this intended?

src/test/java/org/mockbukkit/mockbukkit/util/UnsafeValuesTest.java
```diff
@@ -196,48 +196,76 @@ private void insertTrivialData(ItemMeta meta) throws InvocationTargetException,
...
+			String methodName = method.getName();
+			if("wait".contentEquals(methodName)) continue;
+			if(methodName.startsWith("get")) continue;
+			if(methodName.startsWith("remove")) continue;
+
+			// This method cause error with trivial value. So we set convenient value
+			if(methodName.contentEquals("setAmplifier")){
+				method.invoke(meta, 4);
 				continue;
 			}
```
Thorinwasher, line 208: Good catch here, did not think about this at all. I'm guessing this is what caused the unreliability issues when testing

Thorinwasher (changes requested):
I could not find anything major in a pr of this scope, very nice.

So mostly some nitpicks that I have for you, and there's also the question of backwards compatibility when changing the name of things (although in this case, it is necessary).

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/EnchantmentStorageMetaMock.java
Thorinwasher: oh, we had a discussion about this already, my bad. I guess this is fine then?

src/test/java/org/mockbukkit/mockbukkit/inventory/ItemStackMockTest.java
alexcrea: I did know of that, but I followed previous test style. I will resolve that

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
alexcrea: Oh I did not know that thanks !

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/EnchantmentStorageMetaMock.java
alexcrea: Either way is ok for me

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/FireworkMetaMock.java
alexcrea: It was not intended, thanks !

Thorinwasher (approved):

PR #1204, opened 2024-11-14
Fix equals and implement missing ItemStackMock methods

alexcrea:
# Description
Another fix for ItemStackMock
- Fix equals function
- Implement most part or throw exception on method referencing delegated ItemStack. (null on ItemStackMock)
- Also ensure AIR item has not ItemMeta after setItemMeta

Also implemented PersistentDataContainerViewMock as it is required for getPersistentDataContainer in ItemStack

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
@@ -1,20 +1,23 @@
 package org.mockbukkit.mockbukkit.inventory;
 
-import org.bukkit.inventory.meta.Damageable;
-import org.mockbukkit.mockbukkit.exception.ItemMetaInitException;
-import org.mockbukkit.mockbukkit.inventory.meta.ItemMetaMock;
 import com.google.common.base.Preconditions;
+import io.papermc.paper.persistence.PersistentDataContainerView;
 import org.bukkit.Bukkit;
 import org.bukkit.Material;
 import org.bukkit.NamespacedKey;
 import org.bukkit.Registry;
 import org.bukkit.enchantments.Enchantment;
+import org.bukkit.entity.LivingEntity;
 import org.bukkit.inventory.ItemFlag;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.ItemType;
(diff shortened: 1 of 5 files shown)


src/test/java/org/mockbukkit/mockbukkit/persistence/PersistentDataContainerViewMockTest.java
Thorinwasher, line 1: This file does not follow our style guidlines

Thorinwasher (changes requested):
It looks good to me except for code style in one file. Sorry for the large delay in reviewing this by the way.

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -282,7 +356,7 @@ public boolean equals(Object obj)
 		}
 		if (stack instanceof ItemStackMock bukkit)
 		{
-			return isSimilar(bukkit) && this.amount == bukkit.getAmount() && this.durability == bukkit.durability && Objects.equals(this.getLore(), bukkit.getLore()) && Objects.equals(this.getEnchantments(), bukkit.getEnchantments());
+			return isSimilar(bukkit) && this.amount == bukkit.getAmount() && this.durability == bukkit.durability && Objects.equals(this.itemMeta, bukkit.getItemMeta());
```
Insprill, line 359: Does this match Paper's impl? I remember it being weird and not actually checking if *everything* is equal, not sure if they changed that at some point.

Thorinwasher: I checked this, and it seems to check everything now

alexcrea: Thanks for checking !

src/test/java/org/mockbukkit/mockbukkit/persistence/PersistentDataContainerViewMockTest.java
alexcrea: oups

Thorinwasher (approved):

PR #1206, opened 2024-11-19
Change coverage workflow name

Thorinwasher:
# Description

# Checklist

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -3,6 +3,8 @@ name: Coverage check
 on:
   pull_request:
     branches: [ v1.* ]
+  push:
+    branches: [ v1.* ]
 
 jobs:
   coverage:
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,10 +1,8 @@
-name: Build
+name: Coverage check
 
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yaml
```diff
@@ -1,10 +1,8 @@
-name: Build
+name: Coverage check
 
 on:
   pull_request:
     branches: [ v1.* ]
-  push:
-    branches: [ v1.*]
```
thelooter: This will make it not run on v1.21 , which is required for codecov to actually display our current coverage and coverage over time

thelooter (changes requested):

4everTheOne:
What do you think about renaming this action to:
> Coverage / check

Thorinwasher:
> What do you think about renaming this action to:
> 
> > Coverage / check

Like this?
image

4everTheOne:
I was thinking more like this:
image

Thorinwasher:
> I was thinking more like this: image

I think that would mean renaming the workflow file from `coverage.yml` to `check.yml`. I think it's better to have a more descriptive file name, such as `coverage.yml`.

4everTheOne (approved):

Issue #123, opened 2020-10-05
FileSystemNotFoundException for Tags


TheBusyBiscuit:
Unfortunately Pull Request #118 introduced a slight error.
When it tries to read the `/tags/` category it throws a `FileSystemNotFoundException` as these resources are somehow not carried over properly, I am looking into it already, just wanted to give a heads up about it.

https://pastebin.com/uPXCxwJK

PR #1214, opened 2024-11-27
fix(workflow): update tag message to include commit ID and PR number

thelooter:
# Description
This PR fixes PR Number parsing for tags

> [!WARNING]  
> This relies on having the PR number as part of the commit message e.g. `fizz buzz (#123)`. Changing the format of these commit messages breaks this workflow and requires adaptations to be made to accomodate the new format

# Checklist

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           token: ${{ secrets.BUILD_TOKEN }}
-      
+
       - name: Get merged pull request
         uses: actions-ecosystem/action-get-merged-pull-request@v1.0.1
         if: ${{ github.event_name == 'push' }}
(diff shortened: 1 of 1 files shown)


4everTheOne:
What if instead of parsing the values from the commit message, we just include the commit message as it is? Would it be an option?

.github/workflows/tag.yml
```diff
@@ -43,9 +43,18 @@ jobs:
           current_version: ${{ steps.get-latest-tag.outputs.tag }}
           level: ${{ steps.release-label.outputs.level }}
 
+      - name: Get commit ID for the new tag
+        id: get-commit-id
+        run: echo "commit_id=$(git rev-parse HEAD)" >> $GITHUB_ENV
+
+      - name: Extract PR number from commit message
```
Thorinwasher: Just by looking at the old workflow,  why not use the `github.event.pull_request.number` variable instead. This step feels redundant and unstable

.github/workflows/tag.yml
```diff
@@ -43,9 +43,18 @@ jobs:
           current_version: ${{ steps.get-latest-tag.outputs.tag }}
           level: ${{ steps.release-label.outputs.level }}
 
+      - name: Get commit ID for the new tag
```
Thorinwasher: I'm not in a situation right now to look it up, but there's no variable, for say something like this `github.event.pull_request.commit_id`?

Thorinwasher (changes requested):

.github/workflows/tag.yml
thelooter: Because that didn't work. That's the whole point of this PR to fix this

.github/workflows/tag.yml
thelooter: Because that didn't work. That's the whole point of this PR to fix this

Thorinwasher: Ah, my bad. What about this environment variable instead:
```
steps.pr.outputs.number
```
We have the `actions-ecosystem/action-get-merged-pull-request@v1.0.1` workflow, which should already do this for us

.github/workflows/tag.yml
Thorinwasher: Ah, my bad. What about this environment variable instead:
```
steps.pr.outputs.number
```
We have the `actions-ecosystem/action-get-merged-pull-request@v1.0.1` workflow, which should already do this for us

.github/workflows/tag.yml
Thorinwasher: Okay, my bad. Then this should be fine

Thorinwasher: Actually, why not just use `github.sha`?

Thorinwasher (approved):
Worst case scenario is that the workflow fails whenever a patch or more is merged, which might break other workflows as well. Gong to approve, as I don't see any other way to test this. You only live once

PR #1215, opened 2024-11-29
Add BlockStateMetaMock, adding support for creating ItemStacks which hold TileState meta (chests with inventories, shulker boxes, more)

ens-gijs:
# Description
Adds support for shulker box and chest temStack's with inventory.

# Checklist


ens-gijs:
The following test is not currently working (or included in the PR). `item.getItemMeta()` returns `ItemMetaMock` so clearly there's a hook somewhere I still need to update.

```java
	@Test
	void testInitializedAsPartOfItemStackCreation() {
		ItemStack item = ItemStack.of(Material.BLUE_SHULKER_BOX);
		assertInstanceOf(BlockStateMetaMock.class, item.getItemMeta());
		assertInstanceOf(ContainerStateMock.class, ((BlockStateMetaMock) item.getItemMeta()).getBlockState());
	}
```

ens-gijs:
Ready for review.

Note: I had to hand edit setType.json - I know this file is generated but I don't know what needs updating to make my changes permanent.

src/main/java/org/mockbukkit/mockbukkit/block/state/BlockStateMock.java
```diff
@@ -334,8 +337,27 @@ public boolean isCollidable()
...
+			}
+
+			throw new NoSuchMethodException(
+					"Cannot find an BlockState constructor for the class \"" + target.getName() + "\"");
+		}
+		catch (SecurityException | InstantiationException | IllegalAccessException | InvocationTargetException
+			   | NoSuchMethodException e)
+		{
+			throw new BlockStateInitException(e);
+		}
```
4everTheOne: Personally, I'm not a big fan of using reflection here. I would suggest manually calling the constructor, since it makes it easier to identify usages:
```java
return new BlockStateMock(this);
```

src/main/java/org/mockbukkit/mockbukkit/block/state/CalibratedSculkSensorStateMock.java
```diff
@@ -0,0 +1,26 @@
+package org.mockbukkit.mockbukkit.block.state;
+
+import org.bukkit.Material;
+import org.bukkit.block.Block;
+import org.bukkit.block.CalibratedSculkSensor;
+import org.jetbrains.annotations.NotNull;
+
```
4everTheOne, line 7: Javadoc missing:
```suggestion

/**
 * Mock implementation of a {@link CalibratedSculkSensor}.
 *
 * @see SculkSensorStateMock
 */
```

src/main/java/org/mockbukkit/mockbukkit/block/state/FurnaceStateMock.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.bukkit.Material;
+import org.bukkit.block.BlastFurnace;
+import org.bukkit.block.Block;
+import org.bukkit.block.BlockState;
+import org.bukkit.block.Furnace;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Mock implementation of a {@link BlastFurnace}.
```
4everTheOne:
```suggestion
 * Mock implementation of a {@link Furnace}.
```

4everTheOne (commented):
Thank you for this contribution, I can see you had a lot of work with this :)

I just have some small comments that are open to discussion

src/main/java/org/mockbukkit/mockbukkit/block/state/BlockStateMock.java
ens-gijs: This creates one of the specific state mocks
- BannerStateMock
- BarrelStateMock
- BlastFurnaceStateMock
- BrewingStandStateMock
- ChestStateMock
- DispenserStateMock
- DropperStateMock
- FurnaceStateMock
- HopperStateMock
- LecternStateMock
- ShulkerBoxStateMock
- SmokerStateMock

4everTheOne: Yes, but personally i would prefer to use a [Prototype Constructor](https://refactoring.guru/design-patterns/prototype) to create the items instead of using reflection. Basically something similar to what was done in  #1224.

This is my personal opinion, but i would like to know your opinions on this topic: @ens-gijs , @Thorinwasher and/or @thelooter

src/main/java/org/mockbukkit/mockbukkit/block/state/CalibratedSculkSensorStateMock.java
```diff
@@ -0,0 +1,31 @@
...
+	public CalibratedSculkSensorStateMock(@NotNull Material material)
+	{
+		super(material);
+	}
+
+	protected CalibratedSculkSensorStateMock(@NotNull SculkSensorStateMock state)
+	{
+		super(state);
+	}
+
```
4everTheOne, line 30: [Question] This class does not implement the _getSnapshot_ method. Was this on purpose?

src/test/java/org/mockbukkit/mockbukkit/inventory/meta/BlockStateMetaMockTest.java
```diff
@@ -0,0 +1,177 @@
...
+				.map(Map.Entry::getKey)
+				.map(Arguments::of);
+	}
+
+
+	public static Stream<Arguments> all_MaterialsAndExpectedMetaTypes()
+	{
+		return BlockStateMetaMock.BLOCK_STATE_MATERIALS.entrySet().stream()
+				.map(e -> Arguments.of(e.getKey(), e.getValue()));
+	}
```
4everTheOne: Is this method in use?

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/BlockStateMetaMock.java
```diff
@@ -0,0 +1,323 @@
...
+				{
+					int slot = (int) slotData.getOrDefault("slot", -1);
+					if (slot >= 0)
+					{
+						inventory.setItem(slot, ItemStack.deserialize((Map<String, Object>) slotData.get("item")));
+					}
+				}
+			}
+			// TODO: deserialize other TileStates
+		}
```
Thorinwasher: Probably a good idea to unwrap this, such that it's easier to read. Like this, for say
```suggestion
		if (!args.containsKey("container"))
		{
			return;
		}
		if (material != null)
		{
			blockState = getBlockState();
		}
		if (!(blockState instanceof ContainerStateMock container))
		{
			// TODO: deserialize other TileStates (Other function call probably)
			return;
		}
		Inventory inventory = container.getInventory();
		List<Map<String, Object>> containerData = (List<Map<String, Object>>) args.get("container");
		for (Map<String, Object> slotData : containerData)
		{
			int slot = (int) slotData.getOrDefault("slot", -1);
			if (slot >= 0)
			{
				inventory.setItem(slot, ItemStack.deserialize((Map<String, Object>) slotData.get("item")));
			}
		}
```
I'd say that's more readable, at least for the moment. Probably a good idea to make a separate function for each different state

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ShieldMetaMock.java
```diff
@@ -28,52 +38,256 @@ public ShieldMetaMock()
...
+	public @NotNull BlockState getBlockState()
+	{
+		if (banner != null) return banner.copy();
+		return getBlockState(null);
+	}
+
+	@Override
+	public void setBlockState(@NotNull BlockState blockState)
+	{
+		Preconditions.checkArgument(blockState instanceof Banner, "Invalid blockState");
```
Thorinwasher, line 241: For clarity, probably a good idea to add a separate null check (that throws illegal argument exception), as null values would give the message IllegalArgumentException: Invalid blockstate.

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ShieldMetaMock.java
```diff
@@ -28,52 +38,256 @@ public ShieldMetaMock()
...
+		this.banner = (Banner) blockState;
+	}
+
+	private static Banner getBlockState(DyeColor color)
+	{
+		return new BannerStateMock(dyeColorToMaterial(color));
+	}
+
+	static Material dyeColorToMaterial(DyeColor color)
```
Thorinwasher: Why is this package-private, and not just private?

Thorinwasher (changes requested):
I also found some things in this pr. Looks really good otherwise, I haven't compared to paper implementation though, so there might be some inconsistency. If you want to, I can take a proper look, or @4everTheOne can also do it (I believe in you).

Oh, just going to add, that unwrapping if statements are usually a good idea if you get too many of them nested into each other. Avoids abstraction, and makes stuff easier to read / comprehend

src/main/java/org/mockbukkit/mockbukkit/block/state/BlockStateMock.java
Thorinwasher: From my point of view it's fine. Probably a bit harder to comprehend, but that comes with more automation

ens-gijs: > Yes, but personally i would prefer to use a [Prototype Constructor](https://refactoring.guru/design-patterns/prototype) to create the items instead of using reflection. Basically something similar to what was done in https://github.com/MockBukkit/MockBukkit/pull/1224.

That exists and it's what the reflection finds and delegates to. This `.copy()` pattern is Papers doing not MockBukkits.

The easy way to get rid of reflection here is to require all `BlockStateMock` child classes to provide a `.copy()` implementation. Currently `BlockStateMock` is not abstract and is used a the fallback impl when a more specialized `BlockState` isn't available.

IMO the right way to do this would be to make `BlockStateMock` abstract and not provide a default `copy` impl but this may break existing tests and would require creating a possibly awkward `DefaultBlockStateMock` impl.

Without making this class abstract I could guard against bad copy calls by changing this impl to the following and implement `.copy()` for all existing concretions which would just defer to their own copy constructor.

```java
	@Override
	public @NotNull BlockState copy()
	{
		if (this.getClass() != BlockStateMock.class)
		{
			throw new UnimplementedOperationException(this.getClass().getSimpleName() +
					" does not provide a .copy() implementation!");
		}
		return new BlockStateMock(this);
	}
```

Given the tradeoffs I don't have any particular feelings about which approach is best - well I feel making BlockStateMock abstract is best, but I can't judge if its worth the negative impact to users without doing it as part of a MockBukkit major version bump where people should expect compilation may break.

All that said - I don't really see a negative to requiring concretions to provide a .copy method, using the snip above, and failing the BlockStateMock::copy call if a concretion doesn't provide one. When adding a new concretion it's just as likely that someone will forget to add a .copy override as it is they will forget to create a copy constructor that would be required by the reflection impl. It'd just change the error thrown from "class does not have a copy constructor" to "class does not provide a .copy() impl".

src/main/java/org/mockbukkit/mockbukkit/block/state/CalibratedSculkSensorStateMock.java
ens-gijs: good catch - fixed

src/test/java/org/mockbukkit/mockbukkit/inventory/meta/BlockStateMetaMockTest.java
ens-gijs: nope, removed

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ShieldMetaMock.java
ens-gijs: done

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ShieldMetaMock.java
ens-gijs: Made private.

src/main/java/org/mockbukkit/mockbukkit/block/state/BlockStateMock.java
ens-gijs: implemented non-abstract, non-reflection, solution.

4everTheOne (approved):
Thank you for the improvements! It looks good to me!

Thorinwasher (approved):

Issue #1173, opened 2024-10-30
Support Direct placement of BlockMocks in Mock worlds.


XDMAN500:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

As part of a unit test, I'd like to test against a block with mocked drops. The current [BlockMock](https://github.com/MockBukkit/MockBukkit/blob/219186b03fd1b2410627c9caf06423d3cdeafa6f/src/main/java/be/seeseemelk/mockbukkit/block/BlockMock.java#L580-L591) class does not have an implementation for `Block#getDrops()` and there is no way to add a specialized implementation of the BlockMock to a MockWorld.

The interface for this functionality can be an extension of the existing  [WorldMock#createBlock](https://github.com/MockBukkit/MockBukkit/blob/v1.21/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java#L519C28-L519C62) method with another parameter for the BlockMock instance.

`WorldMock#createBlock(Coordinate, BlockMock)`

Thorinwasher:
It's not better to just add something like this to the BlockMock class?:
```java
public void setBlockDrops(@NotNull Collection<ItemStack> blockDrops) {
    this.blockDrops = blockDrops.deepCopy();
}

@Override
public Collection<ItemStack> getDrops() {
    return this.blockDrops.deepCopy();
}
```

XDMAN500:
That solves the immediate issue I'm having. Since this is a mocking framework, I don't see why the state of the WorldMock needs to be protected. Adding the createBlockmethod would still add flexibility.

Thorinwasher:
> Since this is a mocking framework, I don't see why the state of the WorldMock needs to be protected. Adding the createBlockmethod would still add flexibility.

I don't think I understand what you mean exactly here. Do you mean that it should be possible to completely create new block materials which has custom block drops and so on, or some alternative global modification?

XDMAN500:
No, not a global modification to all blocks.  As a MockBukkit users I want to create a test case with a block that emulates the behavior of a vanilla block.  I should be able to define the behavior of that mock block for that position in the world to write my tests.  Defining custom block mock and placing it directly into the world mock is one way to solve this problem.

PR #1222, opened 2024-12-02
Add custom BlockMock implementations  to WorldMock

XDMAN500:
# Description

Allows the addition of a custom implementation of BlockMock to a WorldMock so that testing against BlockStates, BlockData, and item drops are easier.

Addresses: #1173

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
@@ -554,6 +554,23 @@ public class WorldMock implements World
 		return block;
 	}
 
+	public @NotNull void setBlock(@NotNull BlockMock block)
+	{
+		if (block.getY() >= maxHeight)
+		{
+			throw new ArrayIndexOutOfBoundsException("Y larger than max height");
+		}
+		else if (block.getY() < minHeight)
+		{
+			throw new ArrayIndexOutOfBoundsException("Y smaller than min height");
+		}
+		if (!this.equals(block.getWorld())) {
+			throw new IllegalArgumentException("Block world does not match current world");
+		}
+		blocks.put(new Coordinate(block.getX(), block.getY(), block.getZ()), block);
+	}
+
+
 	@Override
 	public boolean isVoidDamageEnabled()
 	{
--- a/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
@@ -554,7 +554,7 @@ public class WorldMock implements World
 		return block;
 	}
 
-	public @NotNull void setBlock(@NotNull BlockMock block)
+	public void setBlock(@NotNull BlockMock block)
 	{
 		if (block.getY() >= maxHeight)
 		{
--- a/src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
+++ b/src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
@@ -42,7 +42,6 @@ import org.bukkit.inventory.ItemStack;
 import org.bukkit.metadata.FixedMetadataValue;
 import org.bukkit.util.BoundingBox;
 import org.bukkit.util.Consumer;
-import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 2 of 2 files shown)


src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -554,6 +554,23 @@ else if (c.y <= grassHeight)
 		return block;
 	}
 
+	public @NotNull void setBlock(@NotNull BlockMock block)
+	{
+		if (block.getY() >= maxHeight)
+		{
+			throw new ArrayIndexOutOfBoundsException("Y larger than max height");
```
4everTheOne, line 561: I would replace this "Index Out Of Bounds" with the Illegal Argument.

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -554,6 +554,23 @@ else if (c.y <= grassHeight)
...
+		}
+		else if (block.getY() < minHeight)
+		{
+			throw new ArrayIndexOutOfBoundsException("Y smaller than min height");
```
4everTheOne, line 565: I would replace this "Index Out Of Bounds" with the Illegal Argument.

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -554,6 +554,23 @@ else if (c.y <= grassHeight)
 		return block;
 	}
 
+	public @NotNull void setBlock(@NotNull BlockMock block)
```
4everTheOne: Please remove the "NonNull" annotation from the return type since it's void

4everTheOne (commented):
I just have left some small nitpicking comments

Thorinwasher (commented):
What is the usecase for custom block mocks?

Why should this be implemented?

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
XDMAN500: I also prefer the IllegalArgumentException for this situation but I decided to keep this method consistent with the [WorldMock#createBlock](https://github.com/MockBukkit/MockBukkit/blob/a915aa376ec359aca097256f791fc22cb67b99fe/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java#L529-L536)

XDMAN500:
> What is the usecase for custom block mocks?
> 
> Why should this be implemented?

The original issue has the context https://github.com/MockBukkit/MockBukkit/issues/1173.  

As a test writer, I want to be able to create test scenarios that run against blocks that return specific types of BlockState, BlockData, and other exposed data in the Block interface. We could add setters for this data in BlockMock but setting a specialized version of the BlockMock is the most flexible option.

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
XDMAN500: Removed.

Thorinwasher:
> The original issue has the context #1173.
> 
> As a test writer, I want to be able to create test scenarios that run against blocks that return specific types of BlockState, BlockData, and other exposed data in the Block interface. We could add setters for this data in BlockMock but setting a specialized version of the BlockMock is the most flexible option.

Okay, I looked through all fields in block mock. all of them are exposed with getters, and you can also set them, I still don't understand the necessity for this at all. I think you need to provide an exact example for me to understand.

About specialized classes: Instead of having a specialized class for this, why not just have a static function that creates your desired block objects? Like if your class does not provide any new behavior, what's the necessity of it?

XDMAN500:
> > The original issue has the context #1173.
> > As a test writer, I want to be able to create test scenarios that run against blocks that return specific types of BlockState, BlockData, and other exposed data in the Block interface. We could add setters for this data in BlockMock but setting a specialized version of the BlockMock is the most flexible option.
> 
> Okay, I looked through all fields in block mock. all of them are exposed with getters, and you can also set them, I still don't understand the necessity for this at all. I think you need to provide an exact example for me to understand.
> 
> About specialized classes: Instead of having a specialized class for this, why not just have a static function that creates your desired block objects? Like if your class does not provide any new behavior, what's the necessity of it?

The static factory method approach does not work since the specialized block mock will be accessed and modified using the World Interface. The purpose of the specialized BlockMock class IS to add behavior that IS NOT supported in the existing BlockMock object. It is not uncommon to define a mock objects with specialized behavior for a specific unit test in other code under test.

Here are some scenarios where specialized BlockMocks are useful.
1) A PlayerInteractEvent listener  that uses fence gate blocks and expects the BlockData to be an instance of Openable. The event listener will access the block through the World but the test will fail since the default BlockMock BlockData is not an instance of Openable.

2) A ProjectileHitEvent listener that destroys and drops the items of the block the projectile hit with modifications. The listener expects a getDrops() call that does not throw an exception. The unit test will cover multiple branches of the block drop modification logic by defining drops in the mock.

3) A PlayerInteractEvent that moves the player through walls on a right click. The unit test will populate an area in a world with passable an impassable blocks and verify that the player is moved correctly between points separated by a wall of impassable blocks.

Thorinwasher:
> The static factory method approach does not work since the specialized block mock will be accessed and modified using the World Interface. 

?
```java
@Test
void test(){
    initializeBlock(world.getBlock(x,y,z));
}

static void initializeBlock(Block block){
    block.setBlockData(specializedData);
    BlockState state = block.getState();
    // do stuff with state
    state.update();
}
```

> The purpose of the specialized BlockMock class IS to add behavior that IS NOT supported in the existing BlockMock object. It is not uncommon to define a mock objects with specialized behavior for a specific unit test in other code under test.

You gave 3 examples:
1. Can be done through a static method call, which I showed you
2. Is from a lack of features, from what I can understand. I can understand that you would like to have more features that might not be supported at the moment, why not make a pr of that feature instead? That way more people will get to use and analyse it.
3. Do you mean that you tag the blocks through is instance checks or something? You have already tagged the block with a material, so this is not necessary

XDMAN500:
Closing this pull request since this approach is not the correct solution. I'll create another pull request that adds the functionality I need to the BlockMock class.

PR #1228, opened 2024-12-10
Feature/fix container inventory not copied

ens-gijs:
# Description
- fixed ContainerStateMock was creating a shallow copy of inventories
- adds equals and hashCode to ContainerStateMock concretions to be able to use assertEquals in tests
- added toString impls to ContainerStateMock concretions to improve client test debugging

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/block/state/AbstractFurnaceStateMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/block/state/AbstractFurnaceStateMock.java
@@ -198,4 +198,15 @@ public abstract class AbstractFurnaceStateMock extends ContainerStateMock implem
 		return Objects.hash(super.hashCode(), burnTime, cookTime, cookTimeTotal, cookSpeedMultiplier);
 	}
 
+	@Override
+	public String toString()
+	{
+		return "AbstractFurnaceStateMock{" +
+				"burnTime=" + burnTime +
+				", cookTime=" + cookTime +
+				", cookTimeTotal=" + cookTimeTotal +
+				", cookSpeedMultiplier=" + cookSpeedMultiplier +
+				"} " + super.toString();
+	}
+
 }
--- a/src/main/java/org/mockbukkit/mockbukkit/block/state/AbstractFurnaceStateMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/block/state/AbstractFurnaceStateMock.java
@@ -199,14 +199,11 @@ public abstract class AbstractFurnaceStateMock extends ContainerStateMock implem
 	}
 
 	@Override
-	public String toString()
-	{
-		return "AbstractFurnaceStateMock{" +
-				"burnTime=" + burnTime +
+	protected String toStringInternal() {
+		return super.toStringInternal() +
+				", burnTime=" + burnTime +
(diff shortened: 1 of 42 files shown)


src/main/java/org/mockbukkit/mockbukkit/block/state/AbstractFurnaceStateMock.java
```diff
@@ -179,4 +180,33 @@ public FurnaceInventory getSnapshotInventory()
...
+	@Override
+	public int hashCode()
+	{
+		return Objects.hash(super.hashCode(), burnTime, cookTime, cookTimeTotal, cookSpeedMultiplier);
+	}
+
+	@Override
+	public String toString()
+	{
+		return "AbstractFurnaceStateMock{" +
```
Thorinwasher: Won't this give a confusing toString response?
```
SomethingStateMock{<states originating from this class>} SuperStateMock{<states originating from super class>}
```
I would be confused by this, as it would actually seem like it's two separate block states // one SomethingStateMock and one SuperStateMock. This might be fine when printing one block state, would still be confusing from my point of view, but it would get even more confusing when printing multiple states, let's say I write this code:
```java
for(BlockState state : states) {
    System.out.print(state + " ")
}
```
^yes this would become an extremely confusing mess (also bad practise as there's other one line alternatives).

Long story short: You need to have all the components of the super state inside of the brackets of the child state for the return value of the `toString` method call.

Just going to note that another point against this is that the string can be recursively extended, i.e. the super class can also have a super class that is a BlockState. So the behavior of the toString method becomes even more complicated.

Thorinwasher (changes requested):
I think this needs to change, I'm open to debates on this though. I do think it would be confusing as is.

Other than that, looks good to me

src/main/java/org/mockbukkit/mockbukkit/block/state/AbstractFurnaceStateMock.java
ens-gijs: Yeah, I'm not a huge fan of this pattern either. It's what Intellij _suggested_. I've never really like the `Type{properties}` format that is typical to java because of this kind of awkwardness dealing with inheritance. There aren't any examples to follow in MockBukkit for this situation so I just went with Intellij's suggestion figuring it was roughly java-standard-sop.

Do you have any opinions on how you want it to look?

1. Could move the last curly to the end
```
SomethingStateMock{<states originating from this class>, SuperStateMock{<states originating from super class>}}
```

2a. Could unwrap `super.toString()` states using fun string manipulation to find the first curly and skip past it (and not emit an extra close curly)
```
SomethingStateMock{<states originating from this class>, <states originating from super class>}
```

2b. Same result as 2a but don't call `super.toString()` and make every concretion responsible for printing all parent properties - increased maintenance but clean impl.

3. Could say forget the Type{} part making `super.toString()` usage simple and uncomplicated.
```
<states originating from this class>, <states originating from super class>
```

4. other?

ens-gijs: Eh, thought of a much more obvious way to implement no.2 - updating PR

Thorinwasher (approved):

PR #1230, opened 2024-12-13
Add ZombieVillagerMock

Intybyte:
# Description
Add ZombieVillagerMock

# Checklist
- [x] Code follows existing style. (sadly)

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
@@ -73,7 +73,7 @@ public class ZombieVillagerMock extends ZombieMock implements ZombieVillager
 	}
 
 	@Override
-	public Villager.Profession getVillagerProfession()
+	public @NotNull Villager.Profession getVillagerProfession()
 	{
 		return profession;
 	}
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
@@ -1,0 +1,68 @@
+package org.mockbukkit.mockbukkit.entity;
+
+import org.bukkit.OfflinePlayer;
+import org.bukkit.entity.EntityType;
+import org.bukkit.entity.Villager;
+import org.bukkit.entity.ZombieVillager;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.mockbukkit.mockbukkit.ServerMock;
+import org.mockbukkit.mockbukkit.exception.UnimplementedOperationException;
+
+import java.util.UUID;
+
+public class ZombieVillagerMock extends ZombieMock implements ZombieVillager
+{
+	private Villager.Type type = Villager.Type.PLAINS;
+	private OfflinePlayer conversionStarter = null;
+
+	//VillagerType.PLAINS, VillagerProfession.NONE, 1
+	/**
+	 * Constructs a new {@link ZombieMock} on the provided {@link ServerMock} with a specified {@link UUID}.
+	 *
+	 * @param server The server to create the entity on.
+	 * @param uuid   The UUID of the entity.
+	 */
+	public ZombieVillagerMock(@NotNull ServerMock server, @NotNull UUID uuid)
+	{
+		super(server, uuid);
+	}
+
+	@NotNull
+	@Override
+	public Villager.Type getVillagerType()
+	{
+		return type;
+	}
+
+	@Override
+	public void setVillagerType(@NotNull Villager.Type type)
+	{
+		this.type = type;
+	}
+
+	@Override
+	public @Nullable OfflinePlayer getConversionPlayer()
+	{
+		return conversionStarter;
+	}
+
+	@Override
+	public void setConversionPlayer(@Nullable OfflinePlayer offlinePlayer)
+	{
+		this.conversionStarter = offlinePlayer;
+	}
+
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.UUID;
+
+public class ZombieVillagerMock extends ZombieMock implements ZombieVillager
+{
+	private Villager.Type type = Villager.Type.PLAINS;
+	private Villager.Profession profession = Villager.Profession.NONE;
+	private OfflinePlayer conversionStarter = null;
+
+	//VillagerType.PLAINS, VillagerProfession.NONE, 1
```
thelooter: Can you elaborate on the need for this comment?

Intybyte: so i wouldnt forget when copying it over 🤣 🤣 🤣

src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
```diff
@@ -0,0 +1,80 @@
...
+	@Override
+	public @NotNull EntityType getType()
+	{
+		return EntityType.ZOMBIE_VILLAGER;
+	}
+
+	@Override
+	public void setVillagerProfession(Villager.@NotNull Profession profession)
+	{
+		this.profession = profession;
```
4everTheOne, line 75: Small detail missing here:
```java
		Preconditions.checkArgument(profession != null, "Villager.Profession cannot be null");
		this.profession = profession;
```

src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
```diff
@@ -0,0 +1,80 @@
...
+	@Override
+	public Villager.Type getVillagerType()
+	{
+		return type;
+	}
+
+	@Override
+	public void setVillagerType(@NotNull Villager.Type type)
+	{
+		this.type = type;
```
4everTheOne, line 43: Small detail missing here:
```java
		Preconditions.checkArgument(type != null, "Villager.Type cannot be null");
		this.type = type;
```

src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
```diff
@@ -0,0 +1,80 @@
...
+	@Override
+	public @Nullable OfflinePlayer getConversionPlayer()
+	{
+		return conversionStarter;
+	}
+
+	@Override
+	public void setConversionPlayer(@Nullable OfflinePlayer offlinePlayer)
+	{
+		this.conversionStarter = offlinePlayer;
```
4everTheOne, line 56: To mimic the paper behavior we should update this to match:
```java
        if (this.isConverting())
        {
             this.conversionStarter = offlinePlayer;
        }
```

4everTheOne (commented):
Thank you for this contribution! Just have some small suggestions to match the paper behavior :)

src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
Intybyte: converting can't be set and defaults to false, if I were to do this it would not work at all

4everTheOne:
Could you update the unit test class to remove the "public" modifiers to solve the following sonar issues?
https://sonarcloud.io/project/issues?id=MockBukkit_MockBukkit&pullRequest=1230&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

Thanks in advance.

src/test/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMockTest.java
Thorinwasher, line 1: Some parts of zombiemock is not being tested. A really simple test is to test for nullability. For more info look at the code coverage report.

Just going to note that we (should) only use this report as a check whether something has been missed that can have some effect on the behaviour (unimplemented methods does not need to be tested, and really toString methods as well, from my point of view). We also have a policy on getters and setters to test them, I think it's because the implementation can change to something complicated, and it's good to test some simple behavior or something.

Intybyte: Done, only thing that needs null testing technically is conversionStarter, however I don't think i can make a testing for the setter due to offline player not being implemented anywhere that I can see.

src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
4everTheOne: I think the Preconditions.checkArgument was a better solution like suggested in the previous comment. Because this matches the paper behavior.
```java
Preconditions.checkArgument(type != null, "Villager.Type cannot be null");
```

src/main/java/org/mockbukkit/mockbukkit/entity/ZombieVillagerMock.java
4everTheOne: I think the Preconditions.checkArgument was a better solution like suggested in the previous comment. Because this matches the paper behavior.
```java
Preconditions.checkArgument(profession != null, "Villager.Profession cannot be null");
```

Thorinwasher (approved):

PR #1232, opened 2024-12-15
Refactor WorldMock code to improve code readability

4everTheOne:
# Description
This PR refactors the logic to create the entity, and moves the creation logic to `EntitTypes` class.
This make the behaviour more close to what Paper implements, and also makes it easier to work on.

# Checklist


4everTheOne:
This should only be merged after https://github.com/MockBukkit/MockBukkit/pull/1230, so that we can also include the ZombieVillagerMock.

src/main/java/org/mockbukkit/mockbukkit/entity/EntityTypesMock.java
```diff
@@ -0,0 +1,304 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.function.BiFunction;
+
+@ApiStatus.Internal
+public final class EntityTypesMock
+{
+
+	private static final Map<Class<? extends Entity>, EntityData<? extends Entity, ? extends EntityMock>> BUKKIT_TO_MOCK_DATA = new HashMap<>();
```
Thorinwasher: Being a bit picky, and this does not matter that much. But could it be possible to use an unmodifiable map instead here?

By notation, every field that is in capital snakecase should be a constant, in other words static and unmodifiable.

Thorinwasher (commented):
Gonna leave it as a comment, as it's not essential

src/main/java/org/mockbukkit/mockbukkit/entity/EntityTypesMock.java
4everTheOne: Thank you for the feedback! I've update the code to use a build pattern, so that when all the items are registered it changes to a immutable map.

Thorinwasher: Nice! I hope you learned something from me, I mostly left this for you :)

Thorinwasher (approved):
Looks good to me. I know this is still a draft pr though

4everTheOne:
This PR is now ready for review

PR #1234, opened 2024-12-15
Implemented set game time in WorldMock

4everTheOne:
# Description
Implemented set game time and get game time in WorldMock.

# Checklist


src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -3450,8 +3452,19 @@ public void setSendViewDistance(int viewDistance)
...
-		throw new UnimplementedOperationException();
+		return this.gameTime;
+	}
+
+	/**
+	 * Sets the full in-game time on this world since the world generation.
+	 *
+	 * @param time The current absolute time since the world generation.
+	 */
+	@ApiStatus.Internal
```
Thorinwasher, line 2794: Where is this being used internally?

I usually argue that we should expose as litle as possible, but in this case, is it a good idea to expose this to the api?

4everTheOne: At the moment it's not, but i'm working on the villager entity i'll need that 😁🤣

In the sleep method, the villager sets the game time in memory.

Thorinwasher: Ah. Then this is alright. But on hindsight, it might not be a good idea to change the gametime like that, but that's for the other pr to decide.

Thorinwasher (approved):

PR #1235, opened 2024-12-15
Implemented setMemory and getMemory in Living Entity

4everTheOne:
# Description
Implemented the set memory and get memory for the living entity.

# Checklist


src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -988,15 +989,13 @@ public Set<UUID> getCollidableExemptions()
 	@Override
 	public <T> T getMemory(@NotNull MemoryKey<T> memoryKey)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return brain.getMemory(memoryKey).orElse(null);
```
Thorinwasher, line 992: Copies the value of T if it's mutable, i.e. a Location, same for setMemory, I guess.

Throws unsupported operation exception if it's not not one of the following values:
null, Location, Long, UUID, Boolean, Integer. See `org.bukkit.craftbukkit.entity.memory.CraftMemoryMapper#fromNms`. Oh, this might actually be impossible to achieve due to generics, not sure.

src/main/java/org/mockbukkit/mockbukkit/entity/ai/BrainMock.java
Thorinwasher, line 1: Style is inconsistent here

src/main/java/org/mockbukkit/mockbukkit/entity/ai/BrainMock.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+@ApiStatus.Internal
+public class BrainMock
+{
+
+	private static final String MEMORY_KEY_CANNOT_BE_NULL = "Memory key cannot be null";
+
+	private final Map<String, Object> memories = Maps.newHashMap();
```
Thorinwasher: Why do you store every key as a string and not memory keys. The constructor of memorykey is private, so there should only be one instance of each memory key with z namespace value

src/test/java/org/mockbukkit/mockbukkit/entity/ai/BrainMockTest.java
```diff
@@ -0,0 +1,198 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class BrainMockTest
+{
+	private final BrainMock brain = new BrainMock();
+
+	@Nested
+	class SetMemory
```
Thorinwasher, line 77: It should also be possible to set a key to null, if the value is null, it looks like it clears the existing value from the memory map.

Thorinwasher (changes requested):
First proper review in a while. This time I compared to paper implementation and found some differences, otherwise looks good.

src/test/java/org/mockbukkit/mockbukkit/entity/ai/BrainMockTest.java
4everTheOne: This was to achieve the same behavior as Minecraft, but attempting to make it a "clean" approach.
At the moment Minecraft "brain" uses an Optional value. So you can have:
- Pair does not exist in the map.
- Pair has an empty optional as value
- Pair has a value

I tried to keep the same behavior but without using the optional:
- Pair does not exist in the map.
- Pair has a null value
- Pair has a value

src/main/java/org/mockbukkit/mockbukkit/entity/ai/BrainMock.java
4everTheOne: In the beginning i was trying to make it agnostic to the type. But looking back it does not make sense. I'll update it, thank you

src/test/java/org/mockbukkit/mockbukkit/entity/ai/BrainMockTest.java
Thorinwasher: Mmm, okay. But inserting a null value for the value parameter in `LivingEntityMock#setMemory` should clear the memory for that key. It's true that the implementation on brainmock can be as is.

Thorinwasher (approved):

PR #1238, opened 2024-12-16
Implemented mocks for Villager

4everTheOne:
# Description
Implemented mocks for [Villager](https://jd.papermc.io/paper/1.21.3/org/bukkit/entity/Villager.html).

# Checklist

# Relates

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
@@ -2,7 +2,6 @@ package org.mockbukkit.mockbukkit.entity;
 
 import org.bukkit.entity.AbstractVillager;
 import org.bukkit.entity.HumanEntity;
-import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.InventoryHolder;
 import org.bukkit.inventory.Merchant;
 import org.bukkit.inventory.MerchantRecipe;
@@ -42,7 +41,7 @@ public abstract class AbstractVillagerMock extends AgeableMock implements Mercha
 	}
 
 	@Override
-	public @NotNull Inventory getInventory()
+	public @NotNull MerchantInventoryMock getInventory()
 	{
 		return this.inventory;
 	}
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
@@ -16,6 +16,11 @@ import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 
+/**
+ * Mock implementation of an {@link AbstractVillager}.
+ *
+ * @see AgeableMock
+ */
 public abstract class AbstractVillagerMock extends AgeableMock implements Merchant, AbstractVillager, InventoryHolder
 {
 	private final MerchantInventoryMock inventory = new MerchantInventoryMock(this, this);
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
@@ -1,0 +1,112 @@
+package org.mockbukkit.mockbukkit.entity;
+
+import org.bukkit.entity.AbstractVillager;
+import org.bukkit.entity.HumanEntity;
+import org.bukkit.inventory.Inventory;
+import org.bukkit.inventory.InventoryHolder;
+import org.bukkit.inventory.Merchant;
+import org.bukkit.inventory.MerchantRecipe;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.mockbukkit.mockbukkit.ServerMock;
+import org.mockbukkit.mockbukkit.inventory.MerchantInventoryMock;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.UUID;
+
+public abstract class AbstractVillagerMock extends AgeableMock implements Merchant, AbstractVillager, InventoryHolder
+{
+	private final MerchantInventoryMock inventory = new MerchantInventoryMock(this, this);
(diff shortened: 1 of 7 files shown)


4everTheOne:
This PR is pending unit tests for gossip methods. I'll work on it and update the PR according.
From the implementation point of view is ready for review, there are no expected changes on the main class.

src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
```diff
@@ -0,0 +1,116 @@
...
+	@Override
+	public @NotNull MerchantInventoryMock getInventory()
+	{
+		return this.inventory;
+	}
+
+	@Override
+	public void resetOffers()
+	{
+		this.recipes.clear();
```
Thorinwasher, line 51: I think this should also trigger updateTrades as well

src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
```diff
@@ -0,0 +1,116 @@
...
+	@Override
+	public int getRecipeCount()
+	{
+		return this.recipes.size();
+	}
+
+	@Override
+	public boolean isTrading()
+	{
+		return this.getTrader() != null;
```
Thorinwasher, line 94: Same behavior as the default implementation. (You don't need this method at all)

src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
```diff
@@ -0,0 +1,116 @@
...
+	@Override
+	public void resetOffers()
+	{
+		this.recipes.clear();
+	}
+
+	@Override
+	public @NotNull List<MerchantRecipe> getRecipes()
+	{
+		return Collections.unmodifiableList(this.recipes);
```
Thorinwasher: Should probably deep clone the list here. I'm having difficulties interpreting the implementation though

src/main/java/org/mockbukkit/mockbukkit/entity/VillagerMock.java
```diff
@@ -0,0 +1,251 @@
...
+	{
+		// TODO:
+		throw new UnimplementedOperationException();
+	}
+
+	@Override
+	public @NotNull Reputation getReputation(@NotNull UUID uniqueId)
+	{
+		Reputation value = this.gossips.get(uniqueId);
+		return (value == null ? new Reputation() : value);
```
Thorinwasher: I think the value is copied in the paper implementation. (Reputation is mutable)

src/main/java/org/mockbukkit/mockbukkit/entity/VillagerMock.java
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	@Override
+	public @NotNull Map<UUID, Reputation> getReputations()
+	{
+		return Map.copyOf(this.gossips);
```
Thorinwasher: You probably should deep copy the `gossips` field

src/main/java/org/mockbukkit/mockbukkit/entity/VillagerMock.java
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	@Override
+	public void setReputation(@NotNull UUID uniqueId, @NotNull Reputation reputation)
+	{
+		Preconditions.checkNotNull(uniqueId, "UniqueId cannot be null");
+		Preconditions.checkNotNull(reputation, "Reputation cannot be null");
+		this.gossips.put(uniqueId, reputation);
```
Thorinwasher: Copy reputation here

Thorinwasher (changes requested):
Some minor comments on absent deep copies and copies of mutable objects.

About `updateTrades` this needs to be done deterministically. I think I have something in mind which would rely on deterministically fussing the input of this. (Really any random thing)

src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
4everTheOne: Where did you find the default implementation? I don't see the default implementation in the `org.bukkit.inventory.Merchant`:
image

src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
4everTheOne: I had the same doubt, but it seems like that paper is only creating a Immutable list of trades.
I'll try to break it down:

1. In this function the NMS Merchant Offers is converted to CraftMerchant, but it is only created once;
```java
public CraftMerchantRecipe asBukkit() {
  return (this.bukkitHandle == null) ? this.bukkitHandle = new CraftMerchantRecipe(this) : this.bukkitHandle;
}
```
2. In this function we are transforming/mapping the merchant offer in Merchant recipe by just calling the function from step 1 and then creating an immutable list.
```java
@Override
default List<MerchantRecipe> getRecipes() {
    return List.copyOf(Lists.transform(this.getMerchant().getOffers(), new Function<>() {
        @Override
        public MerchantRecipe apply(net.minecraft.world.item.trading.MerchantOffer recipe) {
            return recipe.asBukkit();
        }
    }));
}
```

So in the end it seems like that we are simply returning the original reference and not a clone.

src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
Thorinwasher: Whoops. Might be a 1.21.4 thing.( I checked against 1.21.4 because applying patches is a pain)

src/main/java/org/mockbukkit/mockbukkit/entity/AbstractVillagerMock.java
Thorinwasher: Okay, then that explains that. Might be a bug on the paper end if this is the case. As generally all of these operations should effectively deep clone.

Thorinwasher (approved):

PR #1243, opened 2024-12-27
1.21.4

Thorinwasher:
# Description
1.21.4 stuff

# Checklist


src/test/java/org/mockbukkit/mockbukkit/block/data/BlockDataMockTest.java
```diff
@@ -101,12 +102,11 @@ void testHashCode()
 	@Test
 	void testMatchesNotEquals()
```
Thorinwasher, line 101: This test now fails, this is because it was initially invalid from what I know at least.

It should really only work on deserialized data, from the method comment that is. This method seems to rely on that some values are not set when deserializing blockdata in paper. If all the set values matches with another blockdata values then it returns true, otherwise false. We do stuff a little differently here for mockbukkit, where the default values are always set and stored, even when deserializing.

It's more like this issue surfaced to AIR as I changed how the block data was initialized (to something that is more proper)

So my main thought is to either just screw it and have this method with exactly the same implementation as the equals method, or to change how all BlockData is initialized

src/main/resources/translation/en-us.json
4everTheOne, line 1: Was this removal intentional?

Thorinwasher: Yes, there was a duplicate repository `translations`. So I removed the `translation` repository

4everTheOne (approved):
LGTM!

PR #1246, opened 2024-12-29
Implemented Entities for all the Boat variants

4everTheOne:
# Description
Implemented mocks for [Boats](https://jd.papermc.io/paper/1.21.3/org/bukkit/entity/Boat.html) and [Chest Boats](https://jd.papermc.io/paper/1.21.3/org/bukkit/entity/ChestBoat.html).

# Checklist

# Relates

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/boat/PaleOakChestBoatMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/boat/PaleOakChestBoatMock.java
@@ -17,8 +17,8 @@ public class PaleOakChestBoatMock extends ChestBoatMock implements PaleOakChestB
 
     public PaleOakChestBoatMock(@NotNull ServerMock server, @NotNull UUID uuid)
 	{
-        super(server, uuid);
-    }
+		super(server, uuid);
+	}
 
 	@Override
 	public @NotNull EntityType getType()
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/boat/PaleOakChestBoatMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/boat/PaleOakChestBoatMock.java
@@ -1,0 +1,29 @@
+package org.mockbukkit.mockbukkit.entity.boat;
+
+import org.bukkit.entity.EntityType;
+import org.bukkit.entity.boat.PaleOakChestBoat;
+import org.jetbrains.annotations.NotNull;
+import org.mockbukkit.mockbukkit.ServerMock;
+import org.mockbukkit.mockbukkit.entity.ChestBoatMock;
+
+import java.util.UUID;
+
+/**
+ * Mock implementation of a {@link PaleOakChestBoat}.
+ *
+ * @see ChestBoatMock
+ */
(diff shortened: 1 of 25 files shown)


4everTheOne:
The PR https://github.com/MockBukkit/MockBukkit/pull/1245 needs to be merged first, for this build to succeed.

src/main/java/org/mockbukkit/mockbukkit/entity/boat/PaleOakChestBoatMock.java
```diff
@@ -0,0 +1,30 @@
...
+	public PaleOakChestBoatMock(@NotNull ServerMock server, @NotNull UUID uuid)
+	{
+		super(server, uuid);
+	}
+
+	@Override
+	public @NotNull EntityType getType()
+	{
+		return EntityType.PALE_OAK_CHEST_BOAT;
+	}
```
Thorinwasher, line 28: Instead of doing this in every boat entity class, can't you just pass the type over to the parent class in the constructor and define the method there?

Thorinwasher (commented):
Just going to note that this is one of the stupidier changes I have seen from the paper end. How do you now know if something is a boat, do we need to use tags now or something?

Why have separate types for each of these boats, when the implementation is identical for all of them. This is just unecessary with my given knowledge

src/main/java/org/mockbukkit/mockbukkit/entity/boat/PaleOakChestBoatMock.java
4everTheOne: Initially i was thinking in converting the class `BoatMock` to an abstract class and also make the method `getType` abstract. This would require each new implementation to implement this method. (Personally i think this would also be a good think for the other entities, because it easier to miss that. But fortunately it's not a big issue since we have a unit test that automatically checks that when we try to spawn a new entity)

But this is not perfect, since it would break the compatibility with the existing API, and if someone is using the BoatMock directly would be forced to upgrade.

You suggestion is also an alternative option, but i think it would be a different implementation from what we have in the other entities. Today we override the getType in the entity itself. If you think, it's preferred for this use case, i can update it.

4everTheOne:
> Just going to note that this is one of the stupidier changes I have seen from the paper end. How do you now know if something is a boat, do we need to use tags now or something?
> 
> Why have separate types for each of these boats, when the implementation is identical for all of them. This is just unecessary with my given knowledge

Yes, it's a little bit difficult to that with the EntityTypes. With materials you can use tags, but with the entity types you have to work with this type of logic :(
https://github.com/MockBukkit/MockBukkit/blob/d996018a74315f0c46da7474de26ec53f31da2fe/src/main/java/org/mockbukkit/mockbukkit/entity/BoatMock.java#L40-L66

src/main/java/org/mockbukkit/mockbukkit/entity/boat/PaleOakChestBoatMock.java
Thorinwasher: I was more thinking that we would not need to write as much code, but if you really think about it, it's just copy paste so doesn't really matter.

On another note, it seems like there's a tag for boat entity types. We currently only support material tags at the moment though.

Thorinwasher (approved):

Issue #1241, opened 2024-12-22
Make `UnimplementedOperationException` stacktraces more accessible


Thorinwasher:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature

Currently at least gradle hides any stacktrace with the superclass `TestAbortedException`, this is the superclass of `UnimplementedOperationException`. This makes the stacktrace harder to find, and voids some accessibility and clarity that always is good to have.

We of the mockbukkit dev team still want the `UnimplementedOperationException` to retain it's superclass, as it shows that the exception originates from the mockbukkit end. This issue is for a discussion for alternative solutions for this issue.

4everTheOne:
I have an idea for this using the [TestExecutionExceptionHandler](https://junit.org/junit5/docs/5.3.1/api/org/junit/jupiter/api/extension/TestExecutionExceptionHandler.html). 

It would contain a report with the stackstrace, explaining to the user why the test is being skipped and also suggesting to open an issue or to contribute to the project.
![image](https://github.com/user-attachments/assets/ab766ba8-0194-4084-aa14-0efd325fe244)

I have created a POC in this commit https://github.com/MockBukkit/MockBukkit/commit/16847d197722a15482f8ca34950ee26cd0b4b41e.

4everTheOne:
This issue was resolved by https://github.com/MockBukkit/MockBukkit/pull/1256

PR #1256, opened 2025-01-10
Inform the user about the `UnimplementedOperationException`

4everTheOne:
# Description
This PR is a suggestion for issue https://github.com/MockBukkit/MockBukkit/issues/1241.

# Example
image

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
@@ -1,5 +1,11 @@
 package org.mockbukkit.mockbukkit;
 
+import java.lang.reflect.Field;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.logging.Logger;
+
 import org.apache.commons.lang3.reflect.FieldUtils;
 import org.bukkit.Server;
 import org.junit.jupiter.api.extension.AfterAllCallback;
@@ -13,12 +19,6 @@ import org.junit.jupiter.api.extension.TestInstancePostProcessor;
 import org.junit.jupiter.api.extension.TestInstancePreDestroyCallback;
 import org.junit.platform.commons.util.ExceptionUtils;
 import org.mockbukkit.mockbukkit.exception.UnimplementedOperationException;
-
-import java.lang.reflect.Field;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-import java.util.logging.Logger;
 
 /**
  * Extension that mocks the Bukkit singleton before each test and subsequently unmocks it after each test. It will also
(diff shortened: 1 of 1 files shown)


src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
```diff
@@ -149,4 +157,33 @@ public void afterAll(ExtensionContext context) throws Exception
...
+			logger.info("https://github.com/MockBukkit/MockBukkit/issues");
+			logger.info("or consider contributing by submitting a pull request. \uD83D\uDCAA");
+			logger.info("");
+			logger.info("Your support and contributions keep the MockBukkit magic alive.");
+			logger.info("Thank you for your collaboration! \uD83D\uDE4F");
+			logger.info("");
+			logger.info(HORIZONTAL_DIVIDER);
+			logger.info("\t\t\t\t\t\t\uD83D\uDCCB Technical details \uD83D\uDCCB");
+			logger.info(HORIZONTAL_DIVIDER);
+			logger.info(() -> String.format("Paper Version: %s", BuildParameters.PAPER_API_FULL_VERSION));
```
thelooter, line 179: How about also adding the Paper commit? Should be easy to obtain with [ServerBuildInfo.gitCommit()](https://jd.papermc.io/paper/1.21.4/io/papermc/paper/ServerBuildInfo.html#gitCommit())

thelooter (dismissed):

src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
4everTheOne: At the moment with the ServerBuildInfo we are retrieving the MockBukkit commit, not the paper commit:
https://github.com/MockBukkit/MockBukkit/blob/157715f29ce084e17bfc9dc909b2aab6d879e5e5/src/main/java/org/mockbukkit/mockbukkit/services/ServerBuildInfoMock.java#L69-L73

The MockBukkit commit is already present in line 181:
https://github.com/MockBukkit/MockBukkit/blob/838b39a90205461cd931afc14715631a43c227e5/src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java#L179-L183

Thorinwasher: Yep, now that I think about it, it's probably a better idea to put the paper commit there. I think we need to explicitly define the paper commit in for the paper dependency

4everTheOne: I'm not sure in how we could do that in a way that this could be automatically updated. Does renovate automatically update it?

Thorinwasher: Me neither, I don't know how relevant it is anyhow

Thorinwasher:
This definitely needs to be reviewed, I will take a second look into this

Thorinwasher (approved):
Looks good to me, there might be some temporary build issues though, will look into that

PR #1278, opened 2025-01-30
InventoryMock getContents and getItem should return a clone

benwoo1110:
# Description
Aligns with behaviour of bukkit implementation where they call `asCraftMirror` which is essentially cloning the array.

image

# Checklist


src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryMock.java
```diff
@@ -319,7 +319,7 @@ public ItemStack addItem(@NotNull ItemStack item)
 	@Override
 	public ItemStack @NotNull [] getContents()
 	{
-		return items;
+		return items.clone();
```
Thorinwasher: Should this not also return a clone of each element in the array?

benwoo1110: Hm that's a good point, do you happen to know any built in method that does it?

Thorinwasher: I would do this:
```java
Arrays.toStream(items).map(ItemStack::clone).toArray(ItemStack[]::new)
```

Thorinwasher: On second thought this does not work, you need to have some null checks in there

benwoo1110: ok got it

Thorinwasher (approved):

PR #1284, opened 2025-02-09
Implement `getCraftingRecipe`, `craftItem` and `resetRecipes` in ServerMock

4everTheOne:
# Description
Implemented some of the methods related with recipes in `ServerMock` .

We can now use the methods `getCraftingRecipe` and `craftItem` for `Shapeless` recipes.
I tried to develop the code for `Shaped` recipes but this part is so complex that I think it's better to do it in a separated pull request.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
@@ -1108,15 +1108,13 @@ public class ServerMock extends Server.Spigot implements Server
 	@Override
 	public @NotNull ItemStack craftItem(@NotNull ItemStack[] craftingMatrix, @NotNull World world, @NotNull Player player)
 	{
-		Preconditions.checkArgument(player != null, "player cannot be null");
-		return craftItem(craftingMatrix, world);
+		// TODO Auto-generated method stub
+		throw new UnimplementedOperationException();
 	}
 
 	@Override
 	public @NotNull ItemCraftResult craftItemResult(@NotNull ItemStack[] craftingMatrix, @NotNull World world)
 	{
-		Preconditions.checkArgument(world != null, "world must not be null");
-
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
@@ -1124,9 +1122,6 @@ public class ServerMock extends Server.Spigot implements Server
 	@Override
 	public @NotNull ItemCraftResult craftItemResult(@NotNull ItemStack[] craftingMatrix, @NotNull World world, @NotNull Player player)
 	{
-		Preconditions.checkArgument(world != null, "world cannot be null");
-		Preconditions.checkArgument(player != null, "player cannot be null");
-
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
--- a/src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
@@ -1084,25 +1084,39 @@ public class ServerMock extends Server.Spigot implements Server
 		return this.recipeManager.getRecipeByKey(RecipeType.CRAFTING, key);
 	}
 
-	@Nullable
 	@Override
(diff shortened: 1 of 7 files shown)


src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
```diff
@@ -1084,39 +1084,49 @@ public Recipe getRecipe(@NotNull NamespacedKey key)
...
+		{
+			return ItemStack.empty();
+		}
+
+		return recipe.getResult();
 	}
 
 	@Override
-	public @NotNull ItemCraftResult craftItemResult(@NotNull ItemStack[] craftingMatrix, @NotNull World world)
+	public @NotNull ItemStack craftItem(@NotNull ItemStack[] craftingMatrix, @NotNull World world, @NotNull Player player)
```
Thorinwasher, line 1109: Player is unused here. There seems to be some more checks / behavior when a player is a parameter. I would probably just Make this method throw unimplemented operation exception, as there are alternative methods that does not have player related checks

src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
```diff
@@ -1084,39 +1084,49 @@ public Recipe getRecipe(@NotNull NamespacedKey key)
...
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Preconditions.checkArgument(player != null, "player cannot be null");
+		return craftItem(craftingMatrix, world);
 	}
 
 	@Override
-	public @NotNull ItemCraftResult craftItemResult(@NotNull ItemStack[] craftingMatrix, @NotNull World world, @NotNull Player player)
+	public @NotNull ItemCraftResult craftItemResult(@NotNull ItemStack[] craftingMatrix, @NotNull World world)
```
Thorinwasher, line 1116: This is weird, the method body really only need to throw unimplemented operation exception

src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
```diff
@@ -1084,39 +1084,49 @@ public Recipe getRecipe(@NotNull NamespacedKey key)
...
 	{
+		Preconditions.checkArgument(world != null, "world must not be null");
+
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 	}
 
 	@Override
-	public @NotNull ItemStack craftItem(@NotNull ItemStack[] craftingMatrix, @NotNull World world)
+	public @NotNull ItemCraftResult craftItemResult(@NotNull ItemStack[] craftingMatrix, @NotNull World world, @NotNull Player player)
```
Thorinwasher, line 1123: Same here, the method body should just be a stub

src/main/java/org/mockbukkit/mockbukkit/inventory/RecipeManager.java
```diff
@@ -29,12 +34,34 @@ public class RecipeManager
 {
 
 	private static final String RECIPE_TYPE_CANNOT_BE_NULL = "Recipe type cannot be null";
+	private static final String THE_RECIPE_CANNOT_BE_NULL = "The recipe cannot be null";
+	private static final String THE_CRAFTING_MATRIX_CANNOT_BE_NULL = "The craftingMatrix cannot be null";
```
Thorinwasher: Personal opinion here, up for discussion. Is it necessary to not replicate exception messages. Like the content of those messages does not really need to match each other. I know the linter complains about this, but what does really show most clarity in the code?

Thorinwasher (changes requested):

src/main/java/org/mockbukkit/mockbukkit/inventory/RecipeManager.java
4everTheOne: Yes, i agree with you. This is a rule that's a little bit annoying. Would be nice to have a way to disable this rule.

Thorinwasher: I will look into it

Thorinwasher (approved):

PR #1285, opened 2025-02-12
Make url class loader be parrent of all classloaders

Thorinwasher:
# Description

# Checklist


4everTheOne (commented):
Sorry if it's a dumb question, but what's the difference between the 2?

4everTheOne (approved):
I can see that Paper Classloaders also implement it 👍

Thorinwasher:
> Sorry if it's a dumb question, but what's the difference between the 2?

The url classloader loads classes from other files than the one that is running the program. I think that's the main difference. I made it so that no external files are provided, so the implementation is pretty much void

PR #1288, opened 2025-02-19
Adding player-on-player damage and ability to retrieve the killer

svaningelgem:
# Description
I added this test:
```java
    @Test
    void damage_Player_Kills_OtherPlayer()
    {
        var killer = server.addPlayer();
        player.simulateDamage(player.getHealth(), killer);
        assertEquals(0, player.getHealth(), 0);
        assertTrue(player.isDead());
        assertEquals(killer, player.getKiller());
    }
```
And it fails:
Expected :org.mockbukkit.mockbukkit.entity.PlayerMock@d3f6027e
Actual   :null

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/PlayerMock.java
@@ -2362,11 +2362,11 @@ public class PlayerMock extends HumanEntityMock implements Player, SoundReceiver
 
 		DamageSource dmg;
 		String msg;
-		if (getLastDamageCause() != null && getLastDamageCause().getDamageSource().getDirectEntity() instanceof Player killer)
+		if (getLastDamageCause() != null && getLastDamageCause().getDamageSource().getDirectEntity() instanceof Player player)
 		{
 			dmg = DamageSource.builder(DamageType.PLAYER_ATTACK).build();
-			setKiller(killer);
-			msg = getName() + " was slain by " + killer.getName();
+			setKiller(player);
+			msg = getName() + " was slain by " + player.getName();
 		}
 		else
 		{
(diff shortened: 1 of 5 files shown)


src/main/java/org/mockbukkit/mockbukkit/entity/PlayerMock.java
```diff
@@ -2360,28 +2360,49 @@ public void setHealth(double health)
 			return;
 		}
 
+		DamageSource dmg;
+		String msg;
+		if (getLastDamageCause() != null && getLastDamageCause().getDamageSource().getDirectEntity() instanceof Player killer)
+		{
+			dmg = DamageSource.builder(DamageType.PLAYER_ATTACK).build();
+			setKiller(killer);
+			msg = getName() + " was slain by " + killer.getName();
```
Thorinwasher: Up to you, but there's already a translated messages for death events and such. I stole a large translations file from a Minecraft server (legally). The file is in resources/translations, where there's a shit ton of messages you can choose between, just search for `death.attack.*`

Thorinwasher: Example of usecase:
```java
TranslatableComponent translatableComponent = Component.translatable("argument.entity.selector.unknown", Component.text("@h"));
assertEquals("Unknown selector type '@h'", PlainTextComponentSerializer.plainText().serialize(translatableComponent));
```

svaningelgem: How would I do that? what is the proper way to retrieve entries from there?

Thorinwasher: Just find the right translation key, I would say "death.attack.mob" or "death.attack.player" would do the trick. And then you fill the placeholder text with the names of the players or something like that

svaningelgem: I leave this up to you because I think it'll cleaner to combine LivingEntityMock::setHealth & PlayerMock::setHealth as they share sooooo much. At that time you could also add the translatable message in there. Want me to make a ticket for that? (and I want to head to bed ;))

Thorinwasher: I will just merge this, I can make the ticket as well :). Good night!

Thorinwasher (approved):

PR #1302, opened 2025-02-26
Implemented new methods for CreakingMock and RegistryMock

4everTheOne:
# Description
Added logic for the new methods introduced in commits https://github.com/PaperMC/Paper/commit/fc56c728c0790a6d493ac380c313c940e1100615 and https://github.com/PaperMC/Paper/commit/b00875f86d155741d7a185fe107564203f3a9bca.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
@@ -62,7 +62,7 @@ public class CreakingMock extends MonsterMock implements Creaking
 	 */
 	public void setHomeLocation(@Nullable Location homeLocation)
 	{
-		this.homeLocation = homeLocation;
+		this.homeLocation = homeLocation == null ? null : homeLocation.clone();
 	}
 
 	@Override
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
@@ -1,12 +1,17 @@
 package org.mockbukkit.mockbukkit.entity;
 
+import java.util.UUID;
+
+import com.google.common.base.Preconditions;
+import net.kyori.adventure.sound.Sound;
+import org.bukkit.Location;
 import org.bukkit.entity.Creaking;
 import org.bukkit.entity.EntityType;
+import org.bukkit.entity.Player;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
+import org.jspecify.annotations.Nullable;
 import org.mockbukkit.mockbukkit.ServerMock;
-
-import java.util.UUID;
 
 /**
  * Mock implementation of a {@link Creaking}.
@@ -16,6 +21,8 @@ import java.util.UUID;
 @ApiStatus.Experimental
 public class CreakingMock extends MonsterMock implements Creaking
 {
+	private @Nullable Player activatedBy = null;
+	private @Nullable Location homeLocation = null;
 
 	/**
 	 * Constructs a new {@link Creaking} on the provided {@link ServerMock} with a specified {@link UUID}.
@@ -32,6 +39,55 @@ public class CreakingMock extends MonsterMock implements Creaking
 	public @NotNull EntityType getType()
 	{
 		return EntityType.CREAKING;
+	}
+
+	@Override
+	public void activate(@NotNull Player player)
+	{
+		Preconditions.checkArgument(player != null, "player cannot be null");
+		this.activatedBy = player;
+
+		playSound(Sound.sound(
+			org.bukkit.Sound.ENTITY_CREAKING_ACTIVATE,
+			Sound.Source.HOSTILE,
+			1.0f,
+			1.0f
+		));
+	}
+
+	/**
+	 * Gets the home location for this creaking (where its {@link org.bukkit.block.CreakingHeart} could be found).
+	 *
+	 * @param homeLocation the location of the home if available, or null otherwise
+	 */
+	public void setHomeLocation(@Nullable Location homeLocation)
+	{
+		this.homeLocation = homeLocation;
+	}
+
+	@Override
+	public @Nullable Location getHome()
+	{
+		return this.homeLocation != null ? this.homeLocation.clone() : null;
+	}
+
+	@Override
+	public void deactivate()
+	{
+		this.activatedBy = null;
+
+		playSound(Sound.sound(
+				org.bukkit.Sound.ENTITY_CREAKING_DEACTIVATE,
+				Sound.Source.HOSTILE,
+				1.0f,
+				1.0f
+		));
+	}
+
+	@Override
+	public boolean isActive()
+	{
+		return this.activatedBy != null;
 	}
 
 }
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
@@ -56,7 +56,7 @@ public class CreakingMock extends MonsterMock implements Creaking
 	}
 
 	/**
-	 * Gets the home location for this creaking (where its {@link org.bukkit.block.CreakingHeart} could be found).
+	 * Sets the home location for this creaking (where its {@link org.bukkit.block.CreakingHeart} could be found).
 	 *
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
```diff
@@ -34,4 +41,53 @@ public CreakingMock(@NotNull ServerMock server, @NotNull UUID uuid)
 		return EntityType.CREAKING;
 	}
 
+	@Override
+	public void activate(@NotNull Player player)
```
mariokhoury4, line 45: **nit:** Javadoc is always recommended

src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
```diff
@@ -34,4 +41,53 @@ public CreakingMock(@NotNull ServerMock server, @NotNull UUID uuid)
...
+	{
+		Preconditions.checkArgument(player != null, "player cannot be null");
+		this.activatedBy = player;
+
+		playSound(Sound.sound(
+			org.bukkit.Sound.ENTITY_CREAKING_ACTIVATE,
+			Sound.Source.HOSTILE,
+			1.0f,
```
mariokhoury4, line 53: I see that this constant is used in multiple places. What do you think to create a constant for that?

src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
```diff
@@ -34,4 +41,53 @@ public CreakingMock(@NotNull ServerMock server, @NotNull UUID uuid)
...
+			1.0f
+		));
+	}
+
+	/**
+	 * Gets the home location for this creaking (where its {@link org.bukkit.block.CreakingHeart} could be found).
```
mariokhoury4: I think the description in the javadoc is for the getHome and not for the setHome location no?

src/test/java/org/mockbukkit/mockbukkit/entity/CreakingMockTest.java
```diff
@@ -31,4 +39,62 @@ void getType()
...
 	}
 
+	@Nested
+	class Activate
+	{
+		@Test
+		void givenNullPlayer_thenIllegalArgumentException()
+		{
+			IllegalArgumentException e = assertThrows(IllegalArgumentException.class, () -> creaking.activate(null));
+			assertEquals("player cannot be null", e.getMessage());
```
mariokhoury4, line 49: **Suggestion:** assertThat is usually more readable and better for debugging 
Refer to this: https://dzone.com/articles/the-benefit-of-using-assertthat-over-other-assert

Any thought?

src/test/java/org/mockbukkit/mockbukkit/entity/CreakingMockTest.java
```diff
@@ -31,4 +39,62 @@ void getType()
...
+		}
+
+		@Test
+		void givenValidPlayer_thenCreakingIsActivated()
+		{
+			assertFalse(creaking.isActive());
+
+			PlayerMock player = server.addPlayer();
```
mariokhoury4, line 57: **nit:** could be final

src/test/java/org/mockbukkit/mockbukkit/entity/CreakingMockTest.java
```diff
@@ -31,4 +39,62 @@ void getType()
...
+
+			assertTrue(creaking.isActive());
+		}
+
+	}
+
+	@Test
+	void deactivate()
+	{
+		PlayerMock player = server.addPlayer();
```
mariokhoury4, line 68: **ditto**

src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
Thorinwasher: Isn't that already borrowed from the parent method, as it's overriding

src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
Thorinwasher: I think in some cases, just using a value is better or equal in quality. In this case I don't think it matters

src/test/java/org/mockbukkit/mockbukkit/entity/CreakingMockTest.java
Thorinwasher: I don't think it matters, as it's in a test. I never usually write the final modifier to variables or parameters anyhow, as it's often clear from the IDE that a variable is final and I don't want to be too verbose. (But yeah does not really matter)

src/test/java/org/mockbukkit/mockbukkit/entity/CreakingMockTest.java
Thorinwasher: Both are fine from my point of view, but yeah the assertThat method is probably somewhat better

src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
```diff
@@ -34,4 +41,53 @@ public CreakingMock(@NotNull ServerMock server, @NotNull UUID uuid)
...
+	 *
+	 * @param homeLocation the location of the home if available, or null otherwise
+	 */
+	public void setHomeLocation(@Nullable Location homeLocation)
+	{
+		this.homeLocation = homeLocation;
```
Thorinwasher: You might need to clone this field, as Location is mutable, avoids undefined(ish) behavior.

src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
Thorinwasher: Yep, the description is wrong here, definitely needs to be fixed. Thanks for spotting this!

Thorinwasher (changes requested):
Some minor nitpicks.

Just an internal thought here, as paper pretty much has this contract that fields you can set or get which are mutable will be cloned a lot to avoid undefined behavior. Even if we are introducing new methods, I believe we should try to follow that contract to the best of our abilities.

4everTheOne:
Thank you @mariokhoury4 and @Thorinwasher  for the review.
I'm away from the computer at the moment, but I'll update it later today.
(If you prefer to merge this sooner, i'm able to review the changes if any of you has the time to update it, i just can't code on mobile)

Thorinwasher:
> Thank you mariokhoury4 and Thorinwasher for the review. I'm away from the computer at the moment, but I'll update it later today. (If you prefer to merge this sooner, i'm able to review the changes if any of you has the time to update it, i just can't code on mobile)

Sure, I can add some patches for this

src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
4everTheOne: I also agree with Thorinwasher, if we don't add the annotation we inherit it from the superclass.
An example of this can be seen here: https://javadoc.io/doc/org.mockbukkit.mockbukkit/mockbukkit-v1.21/latest/org/mockbukkit/mockbukkit/entity/PlayerMock.html#getLevel()

"Specified by: getLevel in interface Player"

src/main/java/org/mockbukkit/mockbukkit/entity/CreakingMock.java
4everTheOne: Not sure if defining a constant for this would benefit, an alternative would be overloading the method with the default volume and pitch.

Thorinwasher: Do you mean making a utility method for this?

4everTheOne: Exactly, adding a utility method that can be reused for the other entities, so that we could just do `playSound(sound, source)` instead of `playSound(sound, source, volume, pitch)`

Thorinwasher: Definitely a good idea. I can make a separate pr for that, should be doable.

Thorinwasher: There's actually a builder implementation for sound by just using the method Sond.sound().

Thorinwasher (approved):

PR #1311, opened 2025-03-17
Decorated pot block data

Thorinwasher:
# Description
Implement decorated pot block data

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/block/data/DecoratedPotDataMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/block/data/DecoratedPotDataMock.java
@@ -1,0 +1,71 @@
+package org.mockbukkit.mockbukkit.block.data;
+
+import com.google.common.base.Preconditions;
+import org.bukkit.Material;
+import org.bukkit.block.BlockFace;
+import org.bukkit.block.data.type.DecoratedPot;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Set;
+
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockbukkit/mockbukkit/block/data/DecoratedPotDataMock.java
```diff
@@ -0,0 +1,71 @@
...
+	 * Constructs a new {@link DecoratedPotDataMock} for the provided {@link Material}.
+	 *
+	 * @param material The material this data is for.
+	 */
+	public DecoratedPotDataMock()
+	{
+		super(Material.DECORATED_POT);
+		super.set(BlockDataKey.CRACKED, false);
+		super.set(BlockDataKey.FACING, BlockFace.NORTH);
+		super.set(BlockDataKey.WATERLOGGED, false);
```
4everTheOne: Should we use the "super" key word here? Maybe the "this" or none would fit better. Because it may be overwritten later.

4everTheOne (commented):
Looks good to me, just have a small comment.

src/main/java/org/mockbukkit/mockbukkit/block/data/DecoratedPotDataMock.java
Thorinwasher: Good point, I was mostly just trying to be clear where the method comes from, but that is not that relevant in this context anyhow; will change

4everTheOne (approved):

PR #1313, opened 2025-03-19
Fix two player xp functions not throwing

alexcrea:
# Description
small fix to better match paper behavior

Note: Paper use `Preconditions.checkArgument` instead of `new IllegalArgumentException`. But this copy how current setExp do

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/PlayerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/PlayerMock.java
@@ -1959,8 +1959,7 @@ public class PlayerMock extends HumanEntityMock implements Player, SoundReceiver
 	@Override
 	public void setExp(float exp)
 	{
-		if (exp < 0.0 || exp > 1.0)
-			throw new IllegalArgumentException("Experience progress must be between 0.0 and 1.0");
+		Preconditions.checkArgument(exp >= 0.0 && exp <= 1.0, "Experience progress must be between 0.0 and 1.0 (%s)", exp);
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockbukkit/mockbukkit/entity/PlayerMock.java
```diff
@@ -1973,6 +1973,8 @@ public int getLevel()
 	@Override
 	public void setLevel(int level)
 	{
+		if (level < 0)
```
Thorinwasher: > Note: Paper use Preconditions.checkArgument instead of new IllegalArgumentException. But this copy how current setExp do

We use Preconditions.checkArgument as well, so might as well use that here :)

src/main/java/org/mockbukkit/mockbukkit/entity/PlayerMock.java
```diff
@@ -1985,7 +1987,9 @@ public int getTotalExperience()
 	@Override
 	public void setTotalExperience(int exp)
 	{
-		this.expTotal = Math.max(0, exp);
+		if (exp < 0)
```
Thorinwasher: Same here

Thorinwasher (changes requested):
Looks good, I only have one nitpick

alexcrea:
Fixed nitpick
Also edited setExp to use Preconditions because why not

4everTheOne (approved):

Thorinwasher (approved):

PR #1317, opened 2025-03-20
Start migration of ItemEntityMock to ItemMock

4everTheOne:
# Description
This PR is open for discussion.
I saw that the `Item` implementation was not following the same pattern as the other entities (`<NAME>Mock`), this was a little bit confusing for me, because i was thinking it did not exist yet.

In this PR I propose a solution to gradually migrate the users to the `ItemMock` and when we feel that the time is correct, we can remove the deprecated class.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/ItemMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/ItemMock.java
@@ -17,7 +17,7 @@ import java.util.UUID;
  *
  * @see EntityMock
  */
-public class ItemEntityMock extends EntityMock implements Item
+public class ItemMock extends EntityMock implements Item
 {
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/mockbukkit/mockbukkit/entity/ItemMock.java
4everTheOne, line 1: In GithHub is not clear, but the methods in this file are exactly the same from ItemEntityMock

Thorinwasher:
> In this PR I propose a solution to gradually migrate the users to the ItemMock and when we feel that the time is correct, we can remove the deprecated class.

Just noting this down: Any breaking change should be in a major release, so we should probably remove the old class on mockbukkit 5.

Thorinwasher (approved):

Issue #1322, opened 2025-03-24
InventoryMock getContents and getItem should not return a clone


ColinGrime:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Are you using the latest version of MockBukkit?

- [x] I am using the latest version of MockBukkit.

### Minecraft Version

1.21

### Describe the bug

Hi, I noticed that a recent PR (#1278) changed the behavior of `InventoryMock#getItem()` and `#getContents()` to return cloned `ItemStack`'s.  It was done with the intention of matching CraftBukkit behavior since it uses `asCraftMirror`, but that doesn't create a deep clone of the `ItemStack` - it merely wraps the NMS `ItemStack`.

Normally, if you mutate the result of `getItem()` for example, it would be reflected in the inventory.

### Reproducible Test

```java
@Test
void testInventoryDoesNotReflect() {
    Inventory inventory = new InventoryMock(null, InventoryType.PLAYER);
    inventory.setItem(0, new ItemStackMock(Material.DIAMOND, 3));
    inventory.getItem(0).setAmount(15);
    assertEquals(15, inventory.getItem(0).getAmount()); // Does not reflect -- the inventory still shows 3.
}
```

4everTheOne:
Hello @ColinGrime! Thank you for reporting this issue and for the detailed explanation.

I had a look on the issue, and implemented a solution for it in [v4.42.1](https://github.com/MockBukkit/MockBukkit/releases/tag/v4.42.1) (https://github.com/MockBukkit/MockBukkit/pull/1325) could you please check if this version solves this issue? 

Thank you!

ColinGrime:
Hello, thanks for the quick turnaround.  I have verified that the new version fixes the issue.  All my tests pass now.

Thank you!

PR #1325, opened 2025-03-26
Improved InventoryMock to mirror items properly

4everTheOne:
# Description
This PR fixes the issue #1322. Paper implements the delegate pattern in the CraftItemStack to support the changes in the minecraft item. To emulate this implementation and respect that `getItem` and `getContent` are mirror and not same class, it was required to implement a classe named `ItemStackMirror`. This follows the delegate pattern so that we can emulate the "mirror" operation.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/block/BlockMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/block/BlockMock.java
@@ -1,10 +1,5 @@
 package org.mockbukkit.mockbukkit.block;
 
-import org.mockbukkit.mockbukkit.exception.UnimplementedOperationException;
-import org.mockbukkit.mockbukkit.block.data.BlockDataMock;
-import org.mockbukkit.mockbukkit.block.state.BlockStateMock;
-import org.mockbukkit.mockbukkit.metadata.MetadataTable;
-import org.mockbukkit.mockbukkit.tags.internal.InternalTag;
 import com.destroystokyo.paper.block.BlockSoundGroup;
 import com.google.common.base.Preconditions;
 import org.bukkit.Chunk;
@@ -30,6 +25,11 @@ import org.bukkit.util.Vector;
 import org.bukkit.util.VoxelShape;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.mockbukkit.mockbukkit.block.data.BlockDataMock;
+import org.mockbukkit.mockbukkit.block.state.BlockStateMock;
+import org.mockbukkit.mockbukkit.exception.UnimplementedOperationException;
+import org.mockbukkit.mockbukkit.metadata.MetadataTable;
+import org.mockbukkit.mockbukkit.tags.internal.InternalTag;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -311,9 +311,7 @@ public class BlockMock implements Block
 	@Override
 	public @NotNull BlockState getState()
 	{
-		// This will always return a snapshot of the BlockState, not the actual state.
-		// This is optional with Paper but for Spigot it simply works like that.
-		return state.getSnapshot();
+		return this.state;
 	}
 
 	@Override
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMirror.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMirror.java
@@ -1,0 +1,610 @@
+package org.mockbukkit.mockbukkit.inventory;
+
+import com.google.common.base.Preconditions;
+import io.papermc.paper.datacomponent.DataComponentBuilder;
+import io.papermc.paper.datacomponent.DataComponentType;
+import io.papermc.paper.inventory.ItemRarity;
+import io.papermc.paper.inventory.tooltip.TooltipContext;
+import io.papermc.paper.persistence.PersistentDataContainerView;
+import io.papermc.paper.registry.set.RegistryKeySet;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.event.HoverEvent;
+import org.bukkit.Material;
+import org.bukkit.UndefinedNullability;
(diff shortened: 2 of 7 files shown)


src/main/java/org/mockbukkit/mockbukkit/block/BlockMock.java
```diff
@@ -311,9 +311,7 @@ public BlockFace getFace(@NotNull Block block)
 	@Override
 	public @NotNull BlockState getState()
 	{
-		// This will always return a snapshot of the BlockState, not the actual state.
-		// This is optional with Paper but for Spigot it simply works like that.
-		return state.getSnapshot();
```
4everTheOne, line 316: This comment is kinda true. In paper, the Block state is always a new instance, but the inventory data is stored in the minecraft block which is always the same. So to emulate the inventory we need the inventory to be stored in the state, otherwise there's no other ways to persist the block inventory, since we don't work directly with minecraft entities.

src/test/java/org/mockbukkit/mockbukkit/inventory/ItemStackMirrorTest.java
```diff
@@ -0,0 +1,60 @@
...
+		@Test
+		void givenAnyChangeExecutedInTheMirrorItem_ShouldBeAppliedInTheOriginal()
+		{
+			ItemStack originalItem = new ItemStackMock(Material.DIAMOND, 1);
+			ItemStack mirror = ItemStackMirror.create(originalItem);
+
+			mirror.setAmount(5);
+
+			assertEquals(5, mirror.getAmount());
+			assertEquals(originalItem, mirror);
```
Thorinwasher, line 56: Would this also pass the test?
```java
assertEquals(mirror, originalItem);
```
I looked into assert equals code, and it seems like it only checks if `obj1.equals(obj2)` not `obj2.equals(obj1)`

In this case, it might cause some issues, ItemStackMirror#equals needs to be tested as well

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMirror.java
```diff
@@ -0,0 +1,610 @@
...
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.function.UnaryOperator;
+
+/**
+ * Helper class that emulates the Bukkit mirror behaviour at CraftItemStack#asCraftMirror(net.minecraft.world.item.ItemStack original).
+ *
+ * @see ItemStack
+ */
+public final class ItemStackMirror extends ItemStack
```
Thorinwasher, line 44:
```suggestion
@ApiStatus.Internal
public final class ItemStackMirror extends ItemStack
```

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMirror.java
Thorinwasher, line 1: I'm not well versed in mockito, but would it be possible to create a delagete class using mockito?

Thorinwasher (changes requested):

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMirror.java
Thorinwasher: For say, you could keep this as an utility class, and use the mockito Spy functionality on what now is the delegate. Or am I not thinking this through enough... Like the instance should not be the same, but all values should reflect, even with mutations.

I'm mostly thinking about this, as when a new method gets added to item stack, its child classes needs to also implement those methods, otherwise there's going to be a stack overflow when using that method.

Thorinwasher: Or is it possible to do with bytebuddy 🤔. Anyhow the gest of it, is to make the delegating class be auto generated

src/test/java/org/mockbukkit/mockbukkit/inventory/ItemStackMirrorTest.java
4everTheOne: I'll will add unit tests for that to validate, thanks

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMirror.java
4everTheOne: Probably we could use the Mockito spy, but at the moment this is a dependency that is not included in the project. If you think it is ok to include Mockito as a project dependency i can try to use it.

Thorinwasher: Yeah, it's ok

src/test/java/org/mockbukkit/mockbukkit/inventory/ItemStackMirrorTest.java
4everTheOne: Check, this is ok.

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMirror.java
4everTheOne: I have tried to do that using this code:
```java
@ApiStatus.Internal
public final class ItemStackMirror
{

	@NotNull
	public static ItemStack create(@NonNull ItemStack item)
	{
		Preconditions.checkArgument(item != null, "item cannot be null");

		if (Mockito.mockingDetails(item).isSpy())
		{
			return item;
		}
		return Mockito.spy(item);
	}

	private ItemStackMirror()
	{
		// Hide the public constructor
	}

}
```

But the tests fail because the equals method is not being delegated to the original object:
imagem

It seems like an issue on Mockito end https://github.com/mockito/mockito/issues/2331

Thorinwasher: Alright, then we abandon ship (I will just approve this pr)

Thorinwasher (approved):

PR #1329, opened 2025-03-27
Remove equals and hash code override in inventory mock + other fixes

Thorinwasher:
# Description
I think a lot of tests are failing here, and I believe that's because the inventory should not be cloned when copying a state. But damn inventory stuff is so messy. There's no general principle to follow so messy....

With the current implementation, two separate inventories would be considered equal if they had the same contents, this is inherently wrong. I enountered this issue from doing the following:
```java
Set.of(Bukkit.createInventory(owner1, size, title), Bukkit.createInventory(owner2, size, title));
```
These two inventories should be different, but as both have no item contents, they are considered equal each other
# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryMock.java
@@ -708,13 +708,13 @@ public class InventoryMock implements Inventory
 	@ApiStatus.Internal
 	public Inventory getSnapshot()
 	{
-		try
+		if (InventoryMock.class.equals(getClass()))
 		{
-			return this.getClass().getConstructor(this.getClass()).newInstance(this);
+			return new InventoryMock(this);
 		}
-		catch (NoSuchMethodException | InstantiationException | IllegalAccessException | InvocationTargetException e)
+		else
 		{
-			throw new IllegalStateException(e);
+			throw new UnimplementedOperationException(String.format("%s does not implement method getSnapshot", this.getClass().getSimpleName()));
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


4everTheOne:
I agree with you, the inventory is so messy... Probably i'm missing something, but if we include the "owner" in the equals operation, would that fix it?

Thorinwasher:
> I agree with you, the inventory is so messy... Probably i'm missing something, but if we include the "owner" in the equals operation, would that fix it?

That could possibly fix it, but it feels like it would not handle inventory snapshots well. So I'm just implementing inventory snapshots to something that looks reasonable at the moment, but who knows

Thorinwasher:
Just going to note that snapshot inventories seem to be copied whenever the state is copied, so that's that. 

Probably something wrong with this implementation as well, but hell. I bet it's closer to the paper one at least this time

src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryMock.java
```diff
@@ -703,9 +705,17 @@ public boolean isEmpty()
...
 	{
-		return new InventoryMock(this);
+		try
+		{
+			return this.getClass().getConstructor(this.getClass()).newInstance(this);
+		}
+		catch (NoSuchMethodException | InstantiationException | IllegalAccessException | InvocationTargetException e)
+		{
+			throw new IllegalStateException(e);
+		}
```
4everTheOne: Alternative suggestion for this to avoid using the reflection:
```java
		if (InventoryMock.class.equals(getClass()))
		{
			return new InventoryMock(this);
		}
		else
		{
			throw new UnimplementedOperationException(String.format("%s does not implement method getSnapshot", this.getClass().getSimpleName()));
		}
```
At the moment all the others classes implement this, but if an new one is added, this would notify us about that.

4everTheOne (commented):
Overall looks good to me, just left a suggestion to avoid the reflection.
I'm just worried about the impact that this can have on the consumers tests.

4everTheOne (approved):

PR #1331, opened 2025-03-27
Implement more methods in HumanEntity and Projectile

4everTheOne:
# Description
This PR is a mix in cleanup and implement new methods for the HumanEntity and Projectile.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/HumanEntityMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/HumanEntityMock.java
@@ -68,6 +68,9 @@ public abstract class HumanEntityMock extends LivingEntityMock implements HumanE
 	private int foodLevel = 20;
 	protected boolean blocking;
 	private int sleepTicks = 0;
+	private int saturatedRegenRate = 10;
+	private int unsaturatedRegenRate = 80;
+	private int starvationRate = 80;
 
 	/**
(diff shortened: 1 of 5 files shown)


src/main/java/org/mockbukkit/mockbukkit/entity/HumanEntityMock.java
```diff
@@ -519,8 +530,16 @@ public boolean discoverRecipe(@NotNull NamespacedKey recipe)
...
+		Preconditions.checkArgument(recipes != null, "Recipes cannot be null");
+		AtomicInteger count = new AtomicInteger();
+		recipes.forEach(recipe ->
+		{
+			if (this.discoveredRecipes.add(recipe))
+			{
+				count.incrementAndGet();
+			}
+		});
+		return count.get();
```
Thorinwasher: When I looked at the paper implementation, there seems to be some kind of validation here; the namespaced key needs to be from a registered recipe to be added to the count and recipe collection.

Thorinwasher (changes requested):

src/main/java/org/mockbukkit/mockbukkit/entity/HumanEntityMock.java
4everTheOne: Thank you for the feedback, I completely missed that. I've implemented the behavior.

Thorinwasher (approved):

Issue #1088, opened 2024-08-25
Implement blocks


4everTheOne:
### Is there an existing issue for this?

- [X] I have searched the existing issues

#### Block Data
The list of block data:
- [x] Ageable
- [X] AmethystCluster
- [x] AnaloguePowerable
- [X] Bamboo
- [x] Barrel
- [X] Bed
- [x] Beehive
- [ ] Bell
- [x] BigDripleaf
- [x] Bisected
- [X] BlockData
- [x] BrewingStand
- [x] Brushable
- [x] BubbleColumn
- [x] Cake
- [x] CalibratedSculkSensor
- [X] Campfire
- [x] Candle
- [x] CaveVines
- [x] CaveVinesPlant
- [ ] Chain
- [x] Chest
- [x] ChiseledBookshelf
- [x] Cocoa
- [x] CommandBlock
- [x] Comparator
- [x] CopperBulb
- [x] CoralWallFan
- [x] Crafter
- [x] CreakingHeart
- [x] DaylightDetector
- [X] DecoratedPot
- [x] Directional
- [x] Dispenser
- [x] Door
- [x] Dripleaf
- [x] EnderChest
- [x] EndPortalFrame
- [x] Farmland
- [X] Fence
- [x] Fire
- [x] Furnace
- [x] Gate
- [x] GlassPane
- [ ] GlowLichen
- [x] Grindstone
- [x] HangingMoss
- [x] HangingSign
- [x] Hatchable
- [x] Hopper
- [ ] Jigsaw
- [ ] Jukebox
- [x] Ladder
- [ ] Lantern
- [ ] Leaves
- [x] Lectern
- [x] Levelled
- [x] Light
- [x] Lightable
- [ ] LightningRod
- [x] MangrovePropagule
- [ ] MossyCarpet
- [x] MultipleFacing
- [x] NoteBlock
- [x] Observer
- [x] Orientable
- [ ] PinkPetals
- [x] Piston
- [ ] PistonHead
- [ ] PitcherCrop
- [ ] PointedDripstone
- [x] Powerable
- [x] Rail
- [x] RedstoneRail
- [x] RedstoneWallTorch
- [x] RedstoneWire
- [x] Repeater
- [ ] ResinClump
- [x] RespawnAnchor
- [x] Rotatable
- [x] Sapling
- [ ] Scaffolding
- [ ] SculkCatalyst
- [x] SculkSensor
- [ ] SculkShrieker
- [ ] SculkVein
[...]

PR #1333, opened 2025-04-02
Implement some of block datas

4everTheOne:
# Description
This is a PR related with issue #1088. This contains the base code to improve the BlockData and also implements the following:
- Candle
- Orientable
- Levelled
- Lightable
- Rails

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
@@ -258,7 +258,8 @@ public class BlockDataMock implements BlockData
 	 * @param <B> The expected class type.
 	 */
 	@SuppressWarnings("unchecked")
-	protected <B> Set<B> getValues(B... inputValues) {
+	protected <B> Set<B> getValues(B... inputValues)
+	{
 		ImmutableSet.Builder<B> values = ImmutableSet.builder();
 		for (B value : inputValues)
 		{
--- a/src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
@@ -17,6 +17,7 @@ import org.bukkit.block.BlockSupport;
 import org.bukkit.block.BlockType;
 import org.bukkit.block.PistonMoveReaction;
 import org.bukkit.block.data.BlockData;
+import org.bukkit.block.data.Levelled;
 import org.bukkit.block.data.Orientable;
 import org.bukkit.block.data.type.AmethystCluster;
 import org.bukkit.block.data.type.Bamboo;
@@ -502,7 +503,8 @@ public class BlockDataMock implements BlockData
 			Switch.class, SwitchDataMock::new,
 			Bamboo.class, m -> new BambooDataMock(),
 			DecoratedPot.class, m -> new DecoratedPotDataMock(),
-			Orientable.class, OrientableMock::new
+			Orientable.class, OrientableMock::new,
+			Levelled.class, LevelledDataMock::new
 		);
 
 		for (var entry : factories.entrySet())
--- a/src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
@@ -18,6 +18,7 @@ import org.bukkit.block.BlockType;
 import org.bukkit.block.PistonMoveReaction;
 import org.bukkit.block.data.BlockData;
 import org.bukkit.block.data.Levelled;
+import org.bukkit.block.data.Lightable;
 import org.bukkit.block.data.Orientable;
 import org.bukkit.block.data.type.AmethystCluster;
 import org.bukkit.block.data.type.Bamboo;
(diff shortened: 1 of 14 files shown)


src/main/java/org/mockbukkit/mockbukkit/block/data/LevelledDataMock.java
```diff
@@ -0,0 +1,44 @@
...
+	@Override
+	public int getLevel()
+	{
+		return this.get(BlockDataKey.LEVEL);
+	}
+
+	@Override
+	public void setLevel(int level)
+	{
+		this.set(BlockDataKey.LEVEL, level);
```
Thorinwasher, line 31: You need to validate the input here, should be within the required levels, as far as I know at least

(Should really be part of the BlockDataKey validation, but meh)

src/main/java/org/mockbukkit/mockbukkit/block/data/OrientableMock.java
```diff
@@ -0,0 +1,41 @@
...
+	@Override
+	public @NotNull Axis getAxis()
+	{
+		return this.get(BlockDataKey.AXIS);
+	}
+
+	@Override
+	public void setAxis(@NotNull Axis axis)
+	{
+		this.set(BlockDataKey.AXIS, axis);
```
Thorinwasher, line 36: This should have a check to only allow the axes specified in `#getAxes`

Thorinwasher (changes requested):

src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -217,6 +248,26 @@ protected <T> void set(@NotNull BlockDataKey key, @NotNull T value)
...
+	 */
+	@SuppressWarnings("unchecked")
+	protected <B> Set<B> getValues(B... inputValues)
+	{
+		ImmutableSet.Builder<B> values = ImmutableSet.builder();
+		for (B value : inputValues)
+		{
+			values.add(value);
+		}
+		return values.build();
```
Thorinwasher:
```suggestion
		return Arrays.stream(inputValues).collect(Collectors.toSet());
```
Just going to point out that this method also kind of loses it's value, as the alternative to this is something very short anyways

src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -45,6 +55,27 @@
 public class BlockDataMock implements BlockData
 {
 
+	/**
+	 * This factory tries to create the block data from a material {@link Tag}.
+	 */
+	private static final Map<Tag<Material>, Function<Material, BlockDataMock>> FACTORIES_BY_TAGS = Map.of(
+			Tag.CANDLES, CandleDataMock::new,
+			Tag.RAILS, RailDataMock::new
+	);
```
Thorinwasher: I was a bit confused here initially; I thought you would replace all the functionallity of attemptMockByTag, but that's not the case. I'm fine with both implementations here, really. But I think we have to choose which one to have, to reduce maintanence and confusion

src/main/java/org/mockbukkit/mockbukkit/block/data/OrientableMock.java
4everTheOne: I've updated the code to match this behaviour. But i'll not be able to add unit test for this, since by default the Axis return all values. To validate this i would have to extend the class and override the method, but i don't think this makes sense, since for some reason Paper does use "extension" for Data.

src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
4everTheOne: Nice catch, thanks. I'll update it

src/main/java/org/mockbukkit/mockbukkit/block/data/OrientableMock.java
Thorinwasher: That's fine, not everything has to be tested (a slippery slope though)

src/main/java/org/mockbukkit/mockbukkit/block/data/CandleDataMock.java
```diff
@@ -0,0 +1,69 @@
...
+	@Override
+	public int getCandles()
+	{
+		return this.get(BlockDataKey.CANDLES);
+	}
+
+	@Override
+	public void setCandles(int candles)
+	{
+		this.set(BlockDataKey.CANDLES, candles);
```
Thorinwasher, line 32: This value should also be truncated between the maximum and minimum allowed candles

src/main/java/org/mockbukkit/mockbukkit/block/data/RailDataMock.java
```diff
@@ -0,0 +1,54 @@
...
+	@Override
+	public @NotNull Shape getShape()
+	{
+		return this.get(BlockDataKey.RAIL_SHAPE);
+	}
+
+	@Override
+	public void setShape(@NotNull Shape shape)
+	{
+		this.set(BlockDataKey.RAIL_SHAPE, shape);
```
Thorinwasher, line 35: Seems like these shapes are limited within the bounds of getShapes()

src/main/java/org/mockbukkit/mockbukkit/block/data/RailDataMock.java
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	@Override
+	public @NotNull Set<Shape> getShapes()
+	{
+        return Arrays.stream(Shape.values()).collect(Collectors.toSet());
```
Thorinwasher, line 41: I believe the shapes might be material dependent, for say powered rail does not allow turns.

Thorinwasher (changes requested):
Sorry for adding another review here, but I found some more things that I missed before.

src/main/java/org/mockbukkit/mockbukkit/block/data/RailDataMock.java
4everTheOne: That's correct, but those are considered special types of rails and are covered by `RedstoneRail` (This includes Powered Rails and Detector Rails). If you agree, i'll include those in a separate request.

Thorinwasher (approved):

PR #1335, opened 2025-04-04
Implement block data (Part 2)

4everTheOne:
# Description
This is a PR related with issue #1088 and implements the following block data:
- Ageable
- AnaloguePowerable
- Barrel
- Bisected
- Sapling
- Snowable

Additionally to this a baseline of unit tests was included in file `BlockTypeMockTest`. The use cases that are not implemented yet have a `@Disabled` annotation, that can be just removed in the next developments. This will make it easier to validate future PR's and avoid conflicts.

# Checklist


src/main/java/org/mockbukkit/mockbukkit/block/data/AgeableDataMock.java
```diff
@@ -0,0 +1,41 @@
...
+	{
+		Preconditions.checkArgument(age >= 0, "Age cannot be negative.");
+		this.set(BlockDataKey.AGE_KEY, age);
+	}
+
+	@Override
+	public int getMaximumAge()
+	{
+		throw new UnimplementedOperationException("Get maximum age is not implemented!");
+	}
```
4everTheOne: I did not implement this, because we have so many configurations:
imagem
So I left it without an upper limit

Thorinwasher: You could add something to metaminer for this, might be possible

src/main/java/org/mockbukkit/mockbukkit/block/data/AnaloguePowerableBlockDataMock.java
```diff
@@ -0,0 +1,40 @@
...
+	public void setPower(int power)
+	{
+		Preconditions.checkArgument(power >= 0 && power <= this.getMaximumPower(), "Power must be between 0 and %s", this.getMaximumPower());
+		this.set(BlockDataKey.POWER, power);
+	}
+
+	@Override
+	public int getMaximumPower()
+	{
+		return 15;
```
Thorinwasher: Just double checking, can this value change depending on material?

src/main/java/org/mockbukkit/mockbukkit/block/data/BarrelDataMock.java
```diff
@@ -0,0 +1,59 @@
...
+		Preconditions.checkArgument(this.getFaces().contains(facing), "BlockFace %s is not a valid BlockFace.", facing);
+		this.set(BlockDataKey.FACING, facing);
+	}
+
+	@Override
+	public @NotNull Set<BlockFace> getFaces()
+	{
+		return Stream.of(BlockFace.values())
+				.filter(BlockFace::isCartesian)
+				.collect(Collectors.toSet());
```
Thorinwasher: Is up/down allowed?

4everTheOne:
Thank you for your feedback @Thorinwasher. I've taken you comments in consideration and updated the MetaMiner to extract this values automatically. This values are now stored in the material.json. To make sure the custom values are not printed in getAsString i had to also include a new property named `printableStates` to only include the fields that are printable in bukkit.

Thorinwasher:
> Thank you for your feedback @Thorinwasher. I've taken you comments in consideration and updated the MetaMiner to extract this values automatically. This values are now stored in the material.json. To make sure the custom values are not printed in getAsString i had to also include a new property named `printableStates` to only include the fields that are printable in bukkit.

I'm on mobile now, but I think I partly understand your implementation. Up to discussion, but wouldn't this be better to have this separate to states? I don't think it's really states, it's static properties that depends on the material.

4everTheOne:
> > Thank you for your feedback @Thorinwasher. I've taken you comments in consideration and updated the MetaMiner to extract this values automatically. This values are now stored in the material.json. To make sure the custom values are not printed in getAsString i had to also include a new property named `printableStates` to only include the fields that are printable in bukkit.
> 
> I'm on mobile now, but I think I partly understand your implementation. Up to discussion, but wouldn't this be better to have this separate to states? I don't think it's really states, it's static properties that depends on the material.

Do you have any ideia in how we could do this? The only option i'm seeing at moment does not seem to be a good way to do it. This would require the creation a new file like `block_data.json` that would only include the information that is not printable.
This does not seem to be a good option, because we would have 2 separated files that have similar responsibility, one for printable and another for non printable.

Thorinwasher:
> Do you have any ideia in how we could do this? The only option i'm seeing at moment does not seem to be a good way to do it. This would require the creation a new file like block_data.json that would only include the information that is not printable.
This does not seem to be a good option, because we would have 2 separated files that have similar responsibility, one for printable and another for non printable.

You could just have it in the materials file, like that's fine. So maybe something like this:
```json
[
  {
    "material" : "material_name",
    "state_property": true,
    ...
    "constants" : {
      "max_age" : 3
      ...
    }
  }
]
```
And then you can exclude the field "constants" in whatever is used for that data. Then how you use that in memory, probably just add another map like the one available. I think it's just a good idea to separate these properties though

Thorinwasher:
@4everTheOne Got any ideas on how to continue this, or should I step in?

4everTheOne:
Hello! Sorry i was away for a few days.
That seems good to me. I'm going to take a little bit of time to get back to it, but if you have the opportunity feel free to update it 🙂

4everTheOne:
@Thorinwasher i tried to do some experiments with the suggested changed, and the json looks good. We are something like this:
```json
	"minecraft:zombie_wall_head": {
		"constantValues": {
			"faces": [
				"EAST",
				"NORTH",
				"SOUTH",
				"WEST"
			]
		},
		"facing": "north",
		"powered": false
	}
```
I'm having issues to access this values in the code, since at the moment the only place we have this values is inside `BlockData.data`.

Thorinwasher:
> @Thorinwasher i tried to do some experiments with the suggested changed, and the json looks good. We are something like this:
> 
> ```json
> 	"minecraft:zombie_wall_head": {
> 		"constantValues": {
> 			"faces": [
> 				"EAST",
> 				"NORTH",
> 				"SOUTH",
> 				"WEST"
> 			]
> 		},
> 		"facing": "north",
> 		"powered": false
> 	}
> ```
> 
> I'm having issues to access this values in the code, since at the moment the only place we have this values is inside `BlockData.data`.

Ah, I see.... Hmm alternative is to simply just have it in a separate JSON then. Would that simplify things? :thinking: 

```java
Type type = new TypeToken<Map<Material, Map<String, Object>>>() {}.getType();
blockData = gson.fromJson(jsonObject, type);
```
I guess this needs to be changed then. Direct object serialization is alien for me though

Thorinwasher:
> Hello! Sorry i was away for a few days. That seems good to me. I'm going to take a little bit of time to get back to it, but if you have the opportunity feel free to update it 🙂

Alright, I might do some PR's to this pr then

Thorinwasher:
See pr https://github.com/4everTheOne/MockBukkit/pull/1

Thorinwasher (dismissed):

Thorinwasher (approved):
@4everTheOne I pushed some commits. If you want to, you can take a look into them, other than that.

Loooks good to me othewise

4everTheOne:
Seems good to me too. Should we merge it @Thorinwasher ?

Thorinwasher:
Sure

PR #1355, opened 2025-04-29
Fix a multiple of dependency issues

Thorinwasher:
# Description
- Remove all adventure-api transient dependencies from `net.kyori:adventure-platform-bungeecord:4.3.4"`
- Make some dependencies api scope, as they are used in the mockbukkit api

# Checklist


4everTheOne (approved):
I can't measure the impact this has for the consumers, but seems good me :)

PR #1387, opened 2025-06-24
Add support for Hanging Sign State

4everTheOne:
# Description
This is a PR related with issue #1088 and implements the Hanging Sign block state:

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/block/state/SignStateMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/block/state/SignStateMock.java
@@ -39,7 +39,7 @@ public class SignStateMock extends TileStateMock implements Sign
 	public SignStateMock(@NotNull Material material)
 	{
 		super(material);
-		checkType(material, Tag.SIGNS);
+		checkType(material, Tag.ALL_SIGNS);
 		this.front = new SignSideMock();
 		this.back = new SignSideMock();
 	}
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockbukkit/mockbukkit/block/state/SignStateMock.java
```diff
@@ -39,7 +39,7 @@ public class SignStateMock extends TileStateMock implements Sign
 	public SignStateMock(@NotNull Material material)
 	{
 		super(material);
-		checkType(material, Tag.SIGNS);
+		checkType(material, Tag.ALL_SIGNS);
```
Thorinwasher, line 42: Wouldn't this mean that the normal sign state mock can now be the hanging sign type?

4everTheOne: Yes that's correct. If we look into the implementation an "hanging sign" extends sign.
That was the reason i changed that so when we call super no exception is thown.

Thorinwasher: I'm saying you can do this:
```java
new SignStateMock(Material.ACACIA_HANGING_SIGN)
```
That would not be possible from the previous version. Doing this for a `HangingSignStateMock` should of course be possible.

I kind of don't like the idea of having all of these constructors for block states public, like this should already be achievable using the paper API. Maybe that's a separate PR though

4everTheOne: I understood your point, and it's valid. But in reality this snippet seems to make sense to me. A hanging sign is a sign, in reality the follow the same properties, the only difference is that they are hanging 😁

Thorinwasher (approved):

PR #1388, opened 2025-06-26
Refactored logic to create Block States into a factory

4everTheOne:
# Description

This PR tries to improve the existing logic to create a block state by migrating the creation into a `BlockStateMockFactory`. With this refactor, the code can be centralized in one place and can be easily tested.

By running the MetaMiner, a resource is generate that contains the list of states for each material. With this we can be sure that the code is correct in each update.

Some of the states are not implemented yet and because of that the current csv was updated accordingly here:
https://github.com/MockBukkit/MockBukkit/commit/35b93629eebd73b1dda161f73441530a7ad611b3
- ChiseledBookshelf
- Crafter
- CreakingHeart
- DecoratedPot
- BrushableBlock
- TrialSpawner
- Vault

# Checklist


Thorinwasher (commented):
Looks good to me, I think I only have one question:

Could you add the deprecated for removal annotation to all constructors of block states? I think we should simply just hide the constructors for all of these, i.e. probably make them at least package-private in the next major release. Otherwise, I can do that.

4everTheOne:
> Looks good to me, I think I only have one question:
> 
> Could you add the deprecated for removal annotation to all constructors of block states? I think we should simply just hide the constructors for all of these, i.e. probably make them at least package-private in the next major release. Otherwise, I can do that.

Yes I agree with you, can we include that in a separate PR? My concern is that the PR size will get huge :)

Thorinwasher (approved):

PR #1396, opened 2025-07-01
Add support for Shaped Recipes

4everTheOne:
# Description

> (...) many recipes must have their ingredients placed in the correct relative positions on the crafting grid. These are commonly known as shaped recipes. Ingredients in shaped recipes can be ‘moved’ up, down, left, or right. They can also be flipped side-ways. For example, a 3×1 recipe, such as [bread](https://minecraft.fandom.com/wiki/Bread), can be made using the top, middle, or bottom row of the 3×3 grid, and a [bow](https://minecraft.fandom.com/wiki/Bow) may be made with the strings placed on the left instead of on the right.

Source: https://minecraft.fandom.com/wiki/Crafting

This PR emulates shaped recipe behavior so that players can use the other 674 recipes available in the crafting grid, making it a total of 969 recipes available.

Examples:

## Move behavior:
Pos1
Pos2

## Flipping behavior:
Normal
Flipped

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/RecipeManager.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/RecipeManager.java
@@ -347,6 +347,22 @@ public class RecipeManager
 		return false;
 	}
 
+	/**
+	 * Retrieves the character from the specified position in a shaped recipe grid.
+	 * <p>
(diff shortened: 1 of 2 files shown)


4everTheOne:
Replaced by https://github.com/MockBukkit/MockBukkit/pull/1404

PR #1400, opened 2025-07-04
feat: Add (partially implemented) ticking behaviour for entities and worlds

### LINKED ISSUE #1399: PotionEffect duration not decreasing with scheduler ticks

### Minecraft Version

1.21

## Bug Description
When using MockBukkit's scheduler with `performTicks()`, PotionEffect durations remain constant and don't decrease over time. While effects do expire naturally after their full duration, this causes incorrect behavior when temporarily removing and re-adding effects - they reset to full duration instead of maintaining their remaining time.

## Expected Behavior
PotionEffect durations should decrease by 1 tick per game tick when `server.getScheduler().performTicks()` is called. When a PotionEffect object is removed and re-applied, it should maintain its remaining duration, matching real Bukkit behavior.

## Actual Behavior
PotionEffect durations remain constant at their initial value. When removed and re-applied, effects reset to their original duration instead of preserving remaining time.

## Impact
This makes it impossible to properly test plugins that:
- Store PotionEffect objects and expect them to maintain remaining duration
- Temporarily remove and re-apply effects (e.g., for conditional effect management)
- Implement mechanics that depend on preserving effect timing across removal/re-application cycles

## Additional Notes
Effects do expire naturally after their full duration, and scheduled tasks execute correctly. The core issue is that PotionEffect objects don't track elapsed time, causing duration resets when effects are re-applied. In real Minecraft, the server's tick system updates effect durations continuously, preserving remaining time even when effects are temporarily removed and re-added.

### Reproducible Test

```java
	@Test
	void testEffectDecreasesOnTick() {
		PlayerMock player = server.addPlayer();
		var SPEED_FOR_3S = new PotionEffect(PotionEffectType.SPEED, 3, 0);
		player.addPotionEffect(SPEED_FOR_3S);
		assertEquals(3, ((PotionEffect)(player.getActivePotionEffects().toArray()[0])).getDuration());

		// 1 tick.
		server.getScheduler().performTicks(1);

		// verify it's OK
		assertEquals(3, SPEED_FOR_3S.getDuration()); // The original shouldn't be modified
		var effect = player.getPotionEffect(PotionEffectType.SPEED);
		assertEquals(2, ((PotionEffect)(player.getActivePotionEffects().toArray()[0])).getDuration());
		assertEquals(2, effect.getDuration());
		player.removePotionEffect(PotionEffectType.SPEED);

		assertFalse(player.hasPotionEffect(PotionEffectType.SPEED));

		// 2 ticks..
		server.getScheduler().performTicks(1);

		assertEquals(2, effect.getDuration(), "Effect should have stayed the same");
		player.addPotionEffect(effect);
		assertTrue(player.hasPotionEffect(PotionEffectType.SPEED));

		// 3 ticks (but skipped 1)
		server.getScheduler().performTicks(1);
		assertEquals(1, ((PotionEffect)(player.getActivePotionEffects().toArray()[0])).getDuration());

		// expired on tick 4
		server.getScheduler().performTicks(1);
		assertFalse(player.hasPotionEffect(PotionEffectType.SPEED));
	}
```

### Anything else?

_No response_

svaningelgem:
See #1399 
Closes #1399

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -37,10 +37,10 @@ dependencies {
 
 	implementation("net.bytebuddy:byte-buddy:1.17.6")
 
-	compileOnly("org.projectlombok:lombok:1.18.34")
-	annotationProcessor("org.projectlombok:lombok:1.18.34")
-	testCompileOnly("org.projectlombok:lombok:1.18.34")
-	testAnnotationProcessor("org.projectlombok:lombok:1.18.34")
+	compileOnly("org.projectlombok:lombok:1.18.38")
+	annotationProcessor("org.projectlombok:lombok:1.18.38")
+	testCompileOnly("org.projectlombok:lombok:1.18.38")
+	testAnnotationProcessor("org.projectlombok:lombok:1.18.38")
 
 	// LibraryLoader dependencies
 	implementation("org.apache.maven:maven-resolver-provider:3.8.5")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -37,6 +37,11 @@ dependencies {
 
 	implementation("net.bytebuddy:byte-buddy:1.17.6")
 
+	compileOnly("org.projectlombok:lombok:1.18.34")
+	annotationProcessor("org.projectlombok:lombok:1.18.34")
+	testCompileOnly("org.projectlombok:lombok:1.18.34")
+	testAnnotationProcessor("org.projectlombok:lombok:1.18.34")
+
 	// LibraryLoader dependencies
 	implementation("org.apache.maven:maven-resolver-provider:3.8.5")
 	implementation("org.apache.maven.resolver:maven-resolver-connector-basic:1.9.18")
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
@@ -1584,12 +1584,10 @@ public abstract class EntityMock extends Entity.Spigot implements Entity, Messag
 
 	public void tick()
 	{
-		if (isDead())
+		if (!isDead())
 		{
-			return;
+			++ticksLived;
 		}
-
-		++ticksLived;
 	}
 
 }
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
@@ -4,6 +4,8 @@ import com.destroystokyo.paper.block.TargetBlockInfo;
 import com.destroystokyo.paper.entity.TargetEntityInfo;
 import com.google.common.base.Preconditions;
 import io.papermc.paper.world.damagesource.CombatTracker;
+import lombok.Getter;
+import lombok.Setter;
 import net.kyori.adventure.util.TriState;
 import org.bukkit.Bukkit;
 import org.bukkit.FluidCollisionMode;
@@ -84,13 +86,15 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 	/**
 	 * NoDamage ticks
 	 */
+	@Getter @Setter
 	private int noDamageTicks = 0;
-	private int maxNoDamageTicks = 20;
+	@Getter @Setter
+	private int maximumNoDamageTicks = 20;
 	/**
 	 * Whether the entity is alive.
 	 */
 	protected boolean alive = true;
-	private boolean gliding = false;
+	@Getter @Setter private boolean gliding = false;
 	private boolean jumping = false;
 	private boolean riptiding = false;
 
@@ -103,11 +107,20 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 	private boolean collidable = true;
 	private boolean ai = true;
 	private boolean swimming;
+	/**
+	 * Set whether this entity is slumbering.
+	 */
+	@Getter @Setter
 	private boolean sleeping;
+	/**
+	 * Set whether this entity is climbing.
+	 */
+	@Getter @Setter
 	private boolean climbing;
 	private double absorptionAmount;
 	private int arrowCooldown;
 	private int arrowsInBody;
+	@Getter @Setter
 	private @Nullable Player killer;
 
 	private final Set<ActivePotionEffect> activeEffects = new HashSet<>();
@@ -564,18 +577,6 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 	}
 
 	@Override
-	public int getMaximumNoDamageTicks()
-	{
-		return this.maxNoDamageTicks;
-	}
-
-	@Override
-	public void setMaximumNoDamageTicks(int ticks)
-	{
-		this.maxNoDamageTicks = ticks;
-	}
-
-	@Override
 	public double getLastDamage()
 	{
 		// TODO Auto-generated method stub
@@ -587,18 +588,6 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 	{
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
-	}
-
-	@Override
-	public int getNoDamageTicks()
-	{
-		return this.noDamageTicks;
-	}
-
-	@Override
-	public void setNoDamageTicks(int ticks)
-	{
-		this.noDamageTicks = ticks;
 	}
 
 	@Override
@@ -615,18 +604,6 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 		// TODO Auto-generated method stub
 		throw new UnimplementedOperationException();
 
-	}
-
-	@Override
-	public @Nullable Player getKiller()
-	{
-		return this.killer;
-	}
-
-	@Override
-	public void setKiller(@Nullable Player killer)
-	{
-		this.killer = killer;
 	}
 
 	@Override
@@ -803,11 +780,7 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 	@Override
 	public boolean isLeashed()
 	{
-		if (!(this.leashHolder instanceof Mob))
-		{
-			return false;
-		}
-		return this.leashHolder != null;
+		return this.leashHolder instanceof Mob;
 	}
 
 	@Override
@@ -835,18 +808,6 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 		}
 		this.leashHolder = holder;
 		return true;
-	}
-
-	@Override
-	public boolean isGliding()
-	{
-		return this.gliding;
-	}
-
-	@Override
-	public void setGliding(boolean gliding)
-	{
-		this.gliding = gliding;
 	}
 
 	@Override
@@ -882,38 +843,6 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 		this.riptiding = isRiptiding;
 	}
 
-	/**
-	 * Set whether this entity is slumbering.
-	 *
-	 * @param sleeping If this entity is slumbering
-	 */
-	public void setSleeping(boolean sleeping)
-	{
-		this.sleeping = sleeping;
-	}
-
-	@Override
-	public boolean isSleeping()
-	{
-		return this.sleeping;
-	}
-
-	/**
-	 * Set whether this entity is climbing.
-	 *
-	 * @param climbing If this entity is climbing
-	 */
-	public void setClimbing(boolean climbing)
-	{
-		this.climbing = climbing;
-	}
-
-	@Override
-	public boolean isClimbing()
-	{
-		return this.climbing;
-	}
-
 	@Override
 	public void setAI(boolean ai)
 	{
@@ -934,15 +863,8 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 	{
 		Preconditions.checkNotNull(target, "Target cannot be null");
 
-		if (this instanceof Player)
-		{
-			((Player) this).attack(target);
-		}
-		else
-		{
-			// TODO Auto-generated method stub
-			throw new UnimplementedOperationException();
-		}
+		// TODO Auto-generated method stub
+		throw new UnimplementedOperationException();
 	}
 
 	@Override
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
@@ -59,12 +59,12 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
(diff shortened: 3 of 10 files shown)


svaningelgem:
@Thorinwasher : codecov says that coverage went down (in LivingEntityMock). But I don't see it? Patch % is 100%, but change is -0.65% ?? So, the coverage was 100.65% before 🤣 ?

I added a few more tests now in order (in the hope of) to increase the coverage.
I also introduced lombok to remove some boilerplate code. It won't have any repercussions downstream as it's an annotation preprocessor.

build.gradle.kts
```diff
@@ -37,6 +37,11 @@ dependencies {
 
 	implementation("net.bytebuddy:byte-buddy:1.17.6")
 
+	compileOnly("org.projectlombok:lombok:1.18.34")
+	annotationProcessor("org.projectlombok:lombok:1.18.34")
+	testCompileOnly("org.projectlombok:lombok:1.18.34")
+	testAnnotationProcessor("org.projectlombok:lombok:1.18.34")
```
4everTheOne: @MockBukkit/maintainers what are your opinion about this?
I don't see any problem in using lombok, but i think it should be a community decision :)

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -165,12 +169,23 @@ public synchronized long getCurrentTick()
...
-	 */
-	public synchronized void performOneTick()
+	private void processEntities()
 	{
-		currentTick++;
+		for (World world : Bukkit.getWorlds())
+		{
+			for (Entity entity : world.getEntities())
+			{
+				if (entity instanceof LivingEntityMock living)
```
4everTheOne: In the `EntityMock` class there's a tick method, maybe we could re-use that instead?

src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -103,11 +107,20 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
...
+	 */
+	@Getter @Setter
 	private boolean climbing;
 	private double absorptionAmount;
 	private int arrowCooldown;
 	private int arrowsInBody;
+	@Getter @Setter
 	private @Nullable Player killer;
 
 	private final Set<ActivePotionEffect> activeEffects = new HashSet<>();
```
4everTheOne: Should we replace this with a `Map<String, ActivePotionEffect>` to improve the code readability and performance?

4everTheOne (commented):
Thank you for your contribution @svaningelgem!

I left some comments that are open for discussion ;)

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
svaningelgem: Thanks, I didn't know this. Adjusted now to do just that.
Additional benefit: ANY entity is automatically "ticking" now when doing `getScheduler().performTicks()`

src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
svaningelgem: Thanks, I adjusted this now to use a map. I think it will improve the performance, but code-readability-wise I feel the benefits are negligible.

svaningelgem:
@4everTheOne : I added now also some ticking for the world, because I was having issues with my plugin code that relied on these to calculate the offsets vs the active potion effects.

I will still see if there is any coverage that needs improvement, but I'm guessing this is the final one.

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
```diff
@@ -1584,12 +1584,10 @@ public void lookAt(double v, double v1, double v2, @NotNull LookAnchor lookAncho
...
 	{
-		if (isDead())
+		if (!isDead())
 		{
-			return;
+			++ticksLived;
 		}
-
-		++ticksLived;
 	}
```
Thorinwasher, line 1591: This is a no-op change. Is this really necessary to add to this pr? Like this is more of a weak preference, doesn't really matter does it.

build.gradle.kts
```diff
@@ -37,6 +37,11 @@ dependencies {
 
 	implementation("net.bytebuddy:byte-buddy:1.17.6")
 
+	compileOnly("org.projectlombok:lombok:1.18.38")
+	annotationProcessor("org.projectlombok:lombok:1.18.38")
+	testCompileOnly("org.projectlombok:lombok:1.18.38")
+	testAnnotationProcessor("org.projectlombok:lombok:1.18.38")
```
Thorinwasher, line 43: I don't like lombok, to be real here. What's your opinion @4everTheOne ?

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -2928,4 +2928,9 @@ public void playSound(@NotNull Location location, @NotNull Sound sound, @NotNull
 		throw new UnimplementedOperationException();
 	}
 
+	public void tick()
+	{
+		fullTime++;
```
Thorinwasher, line 2933: Not sure, but shouldn't this change the gametime as well?

src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -822,11 +785,7 @@ public boolean teleport(@NotNull Location location, PlayerTeleportEvent.@NotNull
 	@Override
 	public boolean isLeashed()
 	{
-		if (!(this.leashHolder instanceof Mob))
-		{
-			return false;
-		}
-		return this.leashHolder != null;
+		return this.leashHolder instanceof Mob;
```
Thorinwasher, line 788: I think this is another no-op change. 

This is the code in paper:
```
return this.getHandle().getLeashHolder() != null;
```
See PaperLeashable

Both implementations are wrong, but a fix for this is probably best in another PR.

src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -953,15 +868,8 @@ public void attack(@NotNull Entity target)
...
-		{
-			((Player) this).attack(target);
-		}
-		else
-		{
-			// TODO Auto-generated method stub
-			throw new UnimplementedOperationException();
-		}
+		// TODO Auto-generated method stub
+		throw new UnimplementedOperationException();
```
Thorinwasher, line 872: Was gonna comment on that this removes behavior, but then I realized the previous behavior is stackoverflow, so 👍

Thorinwasher (commented):
I think you're doing more things than you need in one PR. Except for that, it looks good. I just think it's better to leave things untouched, especially if they have nothing to do with the current PR

build.gradle.kts
Thorinwasher: Oh, you wrote the same question, I'm personally against it, but it does not matter that much. From my point of view getters and setters should be avoided anyhow, but that might not be possible as we are faking a whole API, which happily uses those types of methods

Thorinwasher:
> @Thorinwasher : codecov says that coverage went down (in LivingEntityMock). But I don't see it? Patch % is 100%, but change is -0.65% ?? So, the coverage was 100.65% before 🤣 ?
> 
> I added a few more tests now in order (in the hope of) to increase the coverage. I also introduced lombok to remove some boilerplate code. It won't have any repercussions downstream as it's an annotation preprocessor.

I'm not sure how it works, it could be some floating point precision or something. not sure how that coverage calculation is done. I also have this nagging feeling that some tests show different coverage randomly, as some have regrettably system time related behaviors (which you never should have in unit tests)

svaningelgem:
> I think you're doing more things than you need in one PR. Except for that, it looks good. I just think it's better to leave things untouched, especially if they have nothing to do with the current PR

I do agree, I tend to get carried away when debugging and writing tests. However, I think this is relatively limited?

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
svaningelgem: Yeah, I first had another implementation. But I think it makes more sense this way: if you're not dead: increase your time alive.

build.gradle.kts
svaningelgem: I like it because it can remove a LOT of boilerplate code. I'm interested to learn more why you don't like it though?

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
svaningelgem: True, however, I'm wondering what the difference between these 2 is. Because now I'm just doing ++ for both?

src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
svaningelgem: I changed this indeed because that does the same (and looks much cleaner). Shall I revert?

src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
svaningelgem: Yeah, I got this during a test.

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
Thorinwasher: I really think it does not matters, generally the previous behavior is used to un-nest a series of nested if statements, but this does not matter for such small amount of code. I think every programmer will understand this anyhow

svaningelgem:
Adjustments made, please suggest what I still need to change to have this merged?

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
svaningelgem: Shall I revert?

build.gradle.kts
Thorinwasher: Yep, that's why I'm not fully against it. Generally the boilerplate is a sign that your code is not ideally made, you can usually get around with normal java using records and such. For this case though, we can not avoid that, but there's also going to be less thought put into the method as it's just an annotation. The copy operation is very common to use in getters and setters in the api implementations, so might not do that much, and might be missleading.

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
Thorinwasher: Not sure, I think one might be world dependant, whilst the other one is server dependant. Generally both are the same, but sometimes the world can be younger than the server

Thorinwasher:
> I do agree, I tend to get carried away when debugging and writing tests. However, I think this is relatively limited?

It's fine, and I can also get into the pr scope creep sometimes as well 😅.

Thorinwasher:
> Adjustments made, please suggest what I still need to change to have this merged?

Mostly lombok that needs some thought into it

src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
Thorinwasher: Nah, it's fine. I will make an issue on this one though that probably wont be patched by me at least

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
Thorinwasher: Nope, it does not matter, it just does not do anything

svaningelgem:
> Mostly lombok that needs some thought into it

It's your library, so your choice ;-). Though I have seen a few other places that could use getter/setters too. I like Python's properties (or attributes or dataclasses). It's much cleaner in my opinion... Never really like the getter/setter in Java as it feels like an indirection.
However, being what it is (language/API), I feel even less happy with leaving stuff that's basically boilerplate code...

My 3ct :-) (inflation and all)

build.gradle.kts
4everTheOne: I get your point. The biggest downside i see with using it, is that since we are replicating behavior of paper in unit testing, some of the setters require preconditions. If the preconditions are not in lombok, like an int range or it requires a custom message for example. These examples would require us to create the preconditions manually which means for that method we cannot use lombok. If this is the most common use case i feel like it may be unnecessary..

svaningelgem: A cursory look through mockbukkit shows a lot of places where this could be used (ie: where it's simply a get/set operation on the variable).

Thorinwasher: > A cursory look through mockbukkit shows a lot of places where this could be used (ie: where it's simply a get/set operation on the variable).

From my look into this, most have setters have a precondition, and a lot of getters does a copy operation or other operations in them

I guess I could try this with a regex search in the whole project

Thorinwasher: Note: this is mostly me having a fun time with regex and might not give an accurate view of things.

### Getters
There seems to be 1383 return statements that returns a value:
- 3 returns null all the time.
- 734 statements directly returns an internal value

This means for getters, about 50% can be replaced with lombok.
(I used this statement `get[A-Za-z0-9]+\(\)\s+\{\s+return\s+[a-zA-Z.]+;` and varied the end of it a bit)
### Setters
There seems to be about 1009 setters in total that changes an internal value
- 371 purely sets the value based on the parameter and does no checks.

That means for setters, about 37% of setters will have the trivial implementation.
(I used this statement `set[A-Za-z0-9]+\([\@a-zA-Z ]+\)\s+\{\s+this.[a-zA-Z0-9]+ = [a-zA-Z0-9]+;` and varied the end a bit to find different identifiers)

svaningelgem: neat trick. Maybe skip comments too? Because sometimes I've seen comments in there as well?
But there you go.
This is also not accounting for the `@RequiredArgsConstructor` that sometimes can be used (gut feeling: + 30%)

Thorinwasher: I did not consider that, I don't think it's going to be that prevalent, and I also think most cases will not match, as it's a multi line regex, which does not match with the `/` or `*` symbol. Mostly me having fun though 😛

Thorinwasher: @4everTheOne Your opinion? You rule over this matter now with the given data.

4everTheOne: Sorry for the delay, i didn't see the notification about the reply.

> @4everTheOne Your opinion? You rule over this matter now with the given data.

Well, if you give me that power.. i'll say we can include :)
I can see some additional use cases that we can use this for (i already felt the need for it in the past to be honest) some of theme are the Builder annotation and the equal and hashcode.

Given this i think we should allow that :)

4everTheOne (approved):

PR #1403, opened 2025-07-05
chore: lots of cleanups in the code

svaningelgem:
I noticed a few areas that could use improvements.
(well, actually IntelliJ noticed those, and I just started to apply them 🤣 )

Doing that I also fixed a bug with `containsValue` (it was only checking the first command, not all!)


src/test/java/org/mockbukkit/mockbukkit/command/brigadier/bukkit/BukkitBrigadierForwardingMapMockTest.java
```diff
@@ -85,7 +85,7 @@ void put_brigadier()
 		assertEquals(initial + 4, map.size());
 		assertFalse(map.isEmpty());
 		assertTrue(map.containsKey("other_command"));
-		assertTrue(map.keySet().contains("other_command"));
+		assertTrue(map.containsKey("other_command"));
```
4everTheOne: This is now a duplicate of the previous line. This line can be probably removed

4everTheOne (commented):
Curiosity from my side: Did you use any tool for this or you did by hand?

svaningelgem:
> Curiosity from my side: Did you use any tool for this or you did by hand?

IntelliJ :-)

svaningelgem:
Hey @Thorinwasher & @4everTheOne , I'm going to leave this as-is for you to review.
Cleaning up the whole project is a work that will take a much longer time. And I think this PR is already too large as it is.

I wrapped it up by improving the code coverage of anything touched (which is a lot :/). Coverage of the whole project improved by 0.53% (yay!).

Work mostly done by IntelliJ, but here and there I had to get my hands dirty. Most notable is the BrigadierMap class, I shrunk it a LOT. Seems it was just implementing an AbstractMap?? Without changes, all tests keep passing. So I guess that's fine.

svaningelgem:
Depending on PR #1400

src/main/java/org/mockbukkit/mockbukkit/entity/EntityTypesMock.java
```diff
@@ -434,6 +434,7 @@ public <E extends Entity, M extends EntityMock> Builder register(@NotNull Class<
 			Preconditions.checkArgument(bukkitClazz.isAssignableFrom(mockClazz), "The class %s is not a subclass of %s", mockClazz, bukkitClazz);
 			Preconditions.checkArgument(mockFactory != null, "Cannot register a null mock factory");
 			Preconditions.checkArgument(!mapping.containsKey(bukkitClazz), "Cannot register type %s because it's already registered.", bukkitClazz);
+			Preconditions.checkArgument(mockClazz != PlayerMock.class, "Not allowed to register %s.", mockClazz);
```
4everTheOne: Should we replace this with `mockClazz.isAssignableFrom(Player.class)`?

svaningelgem: That is indeed nicer

svaningelgem:
About this sonar issue: that was there before, but because I touched the code it comes up again.
Please advice if I need to do something about it.

4everTheOne (approved):
Thank you for the contribution. This looks good to me, but i will need a second review for this, specially for the `BukkitBrigadierForwardingMapMock`. From my point of view it should not have any impact, but i don't have enough context :)

src/test/java/org/mockbukkit/mockbukkit/command/brigadier/bukkit/BukkitBrigadierForwardingMapMockTest.java
```diff
@@ -56,8 +59,8 @@ void putGetAndRemove()
 		map.put("other_command", command);
 		assertEquals(command, map.get("other_command"));
 		assertTrue(map.containsKey("other_command"));
-		assertTrue(map.values().contains(command));
-		assertTrue(map.keySet().contains("other_command"));
+		assertTrue(map.containsValue(command));
+		assertTrue(map.containsKey("other_command"));
```
Thorinwasher: These are meant to test if the collection views contains the correct elements, BrigadierForwardingMap is still implementing its own map. Even though your method change would make sense normally, it's still a different operation, and should be treated as such when testing.

src/test/java/org/mockbukkit/mockbukkit/block/data/SwitchDataMockTest.java
```diff
@@ -151,7 +151,7 @@ private static Stream<Material> getPossibleMaterials()
 	{
 		Set<Material> possibleMaterials = new HashSet<>(Tag.BUTTONS.getValues());
 		possibleMaterials.add(Material.LEVER);
-		return Stream.of(possibleMaterials.toArray(new Material[possibleMaterials.size()]));
+		return Stream.of(possibleMaterials.toArray(new Material[0]));
```
Thorinwasher:
```suggestion
		return Stream.of(possibleMaterials.toArray(Material[]::new));
```

src/test/java/org/mockbukkit/mockbukkit/block/state/DecoratedPotStateMockTest.java
```diff
@@ -116,13 +116,13 @@ void getSnapshot()
 	@Test
 	void notEqualNull()
 	{
-		assertNotEquals(pot, null);
+		assertNotEquals(null, pot);
```
Thorinwasher: Order of operations is important here.
```suggestion
		// Checks if the pot is not equal to null according to the pot equals method, don't swap
		assertNotEquals(pot, null);
```
Might not be necessary for null, anyhow, JUnit uses the equals method of the first object supplied in the assertEquals

src/test/java/org/mockbukkit/mockbukkit/block/state/DecoratedPotStateMockTest.java
```diff
@@ -116,13 +116,13 @@ void getSnapshot()
...
 	}
 
 	@Test
 	void notEqualObject()
 	{
-		assertNotEquals(pot, new Object());
+		assertNotEquals(new Object(), pot);
```
Thorinwasher:
```suggestion
		// Checks if the pot is not equal to the object using the pot equals method, don't swap
		assertNotEquals(pot, new Object());
```
Might be a reduntant check though

src/test/java/org/mockbukkit/mockbukkit/damage/DamageSourceMockTest.java
```diff
@@ -208,16 +244,56 @@ void equals_and_hashCode_WhenSameInstance()
...
+		assertNotEquals(damageSourceMock.hashCode(), copy.hashCode());
+		assertNotEquals(copy, damageSourceMock);
+		assertNotEquals(copy.hashCode(), damageSourceMock.hashCode());
+	}
+
 	@Test
 	void equals_and_hashCode_WhenNotHavingSameType()
 	{
-		assertNotEquals(damageSourceMock, new Object());
+		assertNotEquals(new Object(), damageSourceMock);
```
Thorinwasher: Again, order of operations matter here, warning is being dumb.
```suggestion
		assertNotEquals(damageSourceMock, new Object());
```

src/test/java/org/mockbukkit/mockbukkit/entity/EntityMockTest.java
```diff
@@ -377,7 +377,7 @@ void equals_DifferentUUID_Different()
 	@Test
 	void equals_DifferentObject_Different()
 	{
-		assertNotEquals(entity, new Object());
+		assertNotEquals(new Object(), entity);
```
Thorinwasher: Order of operations

src/test/java/org/mockbukkit/mockbukkit/entity/ShulkerBulletMockTest.java
```diff
@@ -16,10 +16,7 @@
 
 import java.util.UUID;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotSame;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
Thorinwasher: Wildcard import should be avoided. Just a weak policy we have. You have to manually change this in your IDE, and run a reformatting on all the code.

src/test/java/org/mockbukkit/mockbukkit/inventory/meta/EnchantmentStorageMetaMockTest.java
```diff
@@ -134,6 +134,8 @@ void testEquals()
 	{
 		EnchantmentStorageMetaMock meta = new EnchantmentStorageMetaMock();
 		assertEquals(meta, meta);
+		// TODO: Strange behaviour --> Accurate?
+		assertEquals(new ItemMetaMock(), meta);
```
Thorinwasher, line 138: It's using the equals method in item meta mock, seems like that one does not check if the class of the other object is explicitly the same.

src/test/java/org/mockbukkit/mockbukkit/matcher/inventory/meta/ItemMetaLoreMatcherTest.java
```diff
@@ -61,4 +68,170 @@ protected Matcher<?> createMatcher()
...
+		itemMeta.setLore(null);
+		assertMatches(doesNotHaveLore(Component.text("Hello")), itemMeta);
+	}
+
+	@Nested
+	class DescriptionTests
+	{
+
+		// Test implementation of Description to capture what gets written
+		private static class TestDescription implements Description
```
Thorinwasher, line 105: ~~Mockito~~

src/test/java/org/mockbukkit/mockbukkit/potion/PotionEffectTypeMockTest.java
```diff
@@ -118,7 +118,7 @@ void testEquals_DifferentId()
 	void testEquals_DifferentType()
 	{
 		PotionEffectTypeMock effect = new PotionEffectTypeMock(NamespacedKey.minecraft("speed"), 1, "Speed", false, Color.fromRGB(8171462));
-		assertNotEquals(effect, new Object());
+		assertNotEquals(new Object(), effect);
```
Thorinwasher: Order of operations

src/test/java/org/mockbukkit/mockbukkit/profile/PlayerTexturesMockTest.java
```diff
@@ -144,7 +144,7 @@ void testEquals() throws MalformedURLException
 		Assertions.assertEquals(playerTexturesMock, playerTexturesMock);
 
 		// Assert that other objects are not equal
-		Assertions.assertNotEquals(playerTexturesMock, new Object());
+		Assertions.assertNotEquals(new Object(), playerTexturesMock);
```
Thorinwasher: Order of operations

src/main/java/org/mockbukkit/mockbukkit/entity/EntityTypesMock.java
```diff
@@ -434,6 +434,7 @@ public <E extends Entity, M extends EntityMock> Builder register(@NotNull Class<
 			Preconditions.checkArgument(bukkitClazz.isAssignableFrom(mockClazz), "The class %s is not a subclass of %s", mockClazz, bukkitClazz);
 			Preconditions.checkArgument(mockFactory != null, "Cannot register a null mock factory");
 			Preconditions.checkArgument(!mapping.containsKey(bukkitClazz), "Cannot register type %s because it's already registered.", bukkitClazz);
+			Preconditions.checkArgument(!mockClazz.isAssignableFrom(Player.class), "Not allowed to register %s.", mockClazz);
```
Thorinwasher, line 437: This is a operation change, it might very well be correct, but you have sneaked it in with 154 other file changes.

src/main/java/org/mockbukkit/mockbukkit/inventory/RecipeManager.java
```diff
@@ -207,13 +207,9 @@ public static List<Recipe> loadDefaultRecipes(RecipeType recipeType)
 	{
 		return switch (recipeType)
 		{
-			case RecipeType.BLASTING -> Collections.emptyList();
-			case RecipeType.CAMPFIRE_COOKING -> Collections.emptyList();
+			case RecipeType.BLASTING, RecipeType.STONECUTTING, RecipeType.SMOKING, RecipeType.SMITHING,
```
Thorinwasher: Operation change

Thorinwasher (changes requested):
You have loads of no-op changes, and then you are sneaking in other things that actually change the behavior of the library. Please don't ever do that again, as it's never going to be properly reviewed. I'm including adding new tests and such.

On a normal day, I would have just not viewed through it, but in this case I actually went through it

Thorinwasher:
> Doing that I also fixed a bug with containsValue (it was only checking the first command, not all!)

Then I can make another PR to papermc 🎉, because that implementation is directly copied from there. OR do you want to?

src/main/java/org/mockbukkit/mockbukkit/entity/EntityTypesMock.java
svaningelgem: In this particular case I was trying to improve the coverage for this spawn method...

src/test/java/org/mockbukkit/mockbukkit/entity/ShulkerBulletMockTest.java
svaningelgem: Yeah, I think the threshold is like 3 or 5 and after that intellij starts to group them under a star

src/test/java/org/mockbukkit/mockbukkit/command/brigadier/bukkit/BukkitBrigadierForwardingMapMockTest.java
svaningelgem: strange, and sub-optimal :)

src/test/java/org/mockbukkit/mockbukkit/entity/ShulkerBulletMockTest.java
svaningelgem: image

```
Go to Settings → Editor → Code Style → Java → Imports tab:

Set Class count to use import with '*' to a high number (e.g., 999)
Set Names count to use static import with '*' to a high number (e.g., 999)
```

src/test/java/org/mockbukkit/mockbukkit/inventory/meta/EnchantmentStorageMetaMockTest.java
svaningelgem: I know, shall I add that check? Because now it's inaccurate I think.

src/main/java/org/mockbukkit/mockbukkit/entity/EntityTypesMock.java
svaningelgem: Actually looking into it again: this is NOT an operational change.
No player should ever be added here. Because it has repercussions down the line (most notably in the spawn method). So it's basically adding a safeguard (and yes, we can argue semantics here ;))

src/main/java/org/mockbukkit/mockbukkit/inventory/RecipeManager.java
svaningelgem: Actually no operation change here? It's just grouping all these empty lists into 1 branch?

src/test/java/org/mockbukkit/mockbukkit/matcher/inventory/meta/ItemMetaLoreMatcherTest.java
svaningelgem: What do you mean by this? I know mockito isn't in mockbukkit? But I searched in the whole codebase and it's not used?

svaningelgem:
> > Doing that I also fixed a bug with containsValue (it was only checking the first command, not all!)
> 
> Then I can make another PR to papermc 🎉, because that implementation is directly copied from there. OR do you want to?

Feel free :-).
I don't remember exactly where it was, but it was somewhere there was a loop over commands, and it was just checking the first one. Nothing else. And it probably worked because the first was a correct command anyway.

src/main/java/org/mockbukkit/mockbukkit/entity/EntityTypesMock.java
Thorinwasher: A safe guard, even though it's a good idea, is a operational change.

src/main/java/org/mockbukkit/mockbukkit/inventory/RecipeManager.java
Thorinwasher: My bad. 😅

Thorinwasher:
> Feel free :-).
I don't remember exactly where it was, but it was somewhere there was a loop over commands, and it was just checking the first one. Nothing else. And it probably worked because the first was a correct command anyway.

I made a pr myself on the isEmpty method for this class. They had an inverted output for the isEmpty method 😱

src/test/java/org/mockbukkit/mockbukkit/command/brigadier/bukkit/BukkitBrigadierForwardingMapMockTest.java
Thorinwasher: ~~Yep, it would be greater if it checked both equals method. But that's just the way it is~~ whops wrong comment

src/test/java/org/mockbukkit/mockbukkit/matcher/inventory/meta/ItemMetaLoreMatcherTest.java
Thorinwasher: Haha, I think we need to add mockito to mockbukkit, might be useful for situations like this, when we want to have more in depth information on a mock class. Probably another pr

svaningelgem: Other PR, definitely ;-). And not now ;) at least not for me.

Thorinwasher (approved):

PR #1404, opened 2025-07-05
feat: Adding shaped recipes.

svaningelgem:
# Description

# Checklist

### DIFF
--- a/src/test/java/org/mockbukkit/mockbukkit/inventory/RecipeManagerTest.java
+++ b/src/test/java/org/mockbukkit/mockbukkit/inventory/RecipeManagerTest.java
@@ -125,9 +125,10 @@ class RecipeManagerTest
 		@Test
 		void testPreconditions()
 		{
+			ItemStack itemStack = ItemStack.of(Material.AIR);
 			assertThrows(
 					IllegalArgumentException.class,
-					() -> manager.getRecipesFor(null, ItemStack.of(Material.AIR))
+					() -> manager.getRecipesFor(null, itemStack)
 			);
 
 			assertThrows(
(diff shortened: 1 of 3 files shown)


src/test/java/org/mockbukkit/mockbukkit/inventory/RecipeManagerTest.java
```diff
@@ -308,12 +308,6 @@ class ShapedRecipe
 					"CRIMSON_PLANKS, CRIMSON_DOOR",
 					"WARPED_PLANKS, WARPED_DOOR",
 					"COPPER_INGOT, COPPER_DOOR",
-					"EXPOSED_COPPER, EXPOSED_COPPER_DOOR",
```
svaningelgem, line 311: These don't work in minecraft anyhow

4everTheOne:
Hello! Thanks for your contribution! 
Looks like we have worked the same thing ☺️

https://github.com/MockBukkit/MockBukkit/pull/1396

I'll review it and see what we can retrieve from both solutions

svaningelgem:
Ooops. Mine seems simpler. But yours have more added tests & docs (and mine has full coverage 🎉). I'll stay off it to not duplicate our efforts any more :)

4everTheOne:
Hello again @svaningelgem! So, i agree with you, your solution is easier to understand than mine :)
I noticed that some use cases were missing, like when a shape has "null" ingredients like boats, and when the recipes is mirrored was not implemented.
I opened this PR at https://github.com/svaningelgem/MockBukkit/pull/1 to propose a solution for that. Let me know your thoughts :)

svaningelgem:
> Hello again @svaningelgem! So, i agree with you, your solution is easier to understand than mine :) I noticed that some use cases were missing, like when a shape has "null" ingredients like boats, and when the recipes is mirrored was not implemented. I opened this PR at [svaningelgem#1](https://github.com/svaningelgem/MockBukkit/pull/1) to propose a solution for that. Let me know your thoughts :)

I merged everything into this branch. I didn't really have any comments as it seemed straightforward.
I cleaned it up a little bit more though.
Also added full coverage (the mirror was sometimes not needed, and there always is at least 1 row).

4everTheOne (approved):

PR #1405, opened 2025-07-06
feat: easily load plugin via its class

svaningelgem:
# Description

I know there is already a `loadPlugin(<class>, <params>)` method, but that one needs a descriptor file.
This method adds the benefit of transparency: if the file is not there, a simple dummy descriptor is created and used.
So people don't need to really bother if they don't care about names and versions and so on in their tests, but just about the functionality.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/plugin/PluginManagerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/plugin/PluginManagerMock.java
@@ -408,6 +408,18 @@ public class PluginManagerMock extends PermissionManagerMock implements PluginMa
 	}
 
 	/**
+	 * Load a plugin from a class. The description will be guessed as best as possible.
+	 *
+	 * @param class1      The plugin to load.
+	 * @return The loaded plugin.
+	 */
+	public @NotNull JavaPlugin loadPlugin(@NotNull Class<? extends JavaPlugin> class1)
+	{
+		PluginDescriptionFile description = new PluginDescriptionFile(class1.getName(), "0.0.0", class1.getName());
+		return loadPlugin(class1, description, new Object[0]);
+	}
+
+	/**
 	 * Load a plugin from a class. It will use the system resource {@code plugin.yml} as the resource file.
 	 *
 	 * @param description The {@link PluginDescriptionFile} that contains information about the plugin.
--- a/src/main/java/org/mockbukkit/mockbukkit/plugin/PluginManagerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/plugin/PluginManagerMock.java
@@ -421,7 +421,7 @@ public class PluginManagerMock extends PermissionManagerMock implements PluginMa
 		}
 		catch (PluginLoadException ignored)
 		{
-			PluginDescriptionFile description = new PluginDescriptionFile(class1.getName(), "0.0.0", class1.getName());
+			PluginDescriptionFile description = new PluginDescriptionFile(class1.getSimpleName(), "0.0.0", class1.getSimpleName());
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockbukkit/mockbukkit/plugin/PluginManagerMock.java
```diff
@@ -407,6 +407,25 @@ public void registerLoadedPlugin(@NotNull Plugin plugin)
...
+	 */
+	public @NotNull JavaPlugin loadPlugin(@NotNull Class<? extends JavaPlugin> class1)
+	{
+		try
+		{
+			return loadPlugin(class1, new Object[0]);
+		}
+		catch (PluginLoadException ignored)
+		{
+			PluginDescriptionFile description = new PluginDescriptionFile(class1.getSimpleName(), "0.0.0", class1.getSimpleName());
```
4everTheOne: According to the Javadocs it has to be the full location: https://jd.papermc.io/paper/1.21.7/org/bukkit/plugin/PluginDescriptionFile.html#%3Cinit%3E(java.lang.String,java.lang.String,java.lang.String)
```suggestion
			PluginDescriptionFile description = new PluginDescriptionFile(class1.getSimpleName(), "0.0.0", class1.getName());
```

src/test/java/org/mockbukkit/mockbukkit/plugin/PluginManagerMockTest.java
```diff
@@ -390,6 +392,22 @@ void loadPlugin_InvalidName_DoesntLoad(String name) throws ReflectiveOperationEx
...
+		JavaPlugin loadedPlugin = pluginManager.loadPlugin(TestPlugin.class);
+		assertInstanceOf(JavaPlugin.class, loadedPlugin);
+		assertEquals("MockBukkitTestPlugin", loadedPlugin.getName());
+		assertEquals("0.1.0", loadedPlugin.getDescription().getVersion());
+
+		// FileNotFound flow
+		loadedPlugin = pluginManager.loadPlugin(JavaPlugin.class); // Internal plugin, can't find plugin.yml file
+		assertInstanceOf(JavaPlugin.class, loadedPlugin);
+		assertEquals("JavaPlugin", loadedPlugin.getName());
+		assertEquals("0.0.0", loadedPlugin.getDescription().getVersion());
```
4everTheOne: Can we separate this into 2 unit tests?
1. Loading plugin with descriptor file
2. Loading plugin without descriptor file

4everTheOne (commented):
Thank you for the contribution [svaningelgem](https://github.com/svaningelgem). This is a simple improvement but for sure a really nice to have!

Just have some small suggestions for improvement.

src/test/java/org/mockbukkit/mockbukkit/plugin/PluginManagerMockTest.java
svaningelgem: Done

svaningelgem:
@Thorinwasher , @4everTheOne : could you re-run the failed flows again? They should work as-is (retried it locally and it works without a problem)

4everTheOne:
> @Thorinwasher , @4everTheOne : could you re-run the failed flows again? They should work as-is (retried it locally and it works without a problem)

Looks like it persisted. Looks like there's a new commit in the paper: https://github.com/PaperMC/Paper/commits/main/
That may have created changes that need to be worked on. I'll check that in 30 minutes if not solved until there

4everTheOne:
Issue has been resolved in https://github.com/MockBukkit/MockBukkit/pull/1406

4everTheOne (approved):

4everTheOne:

PR #1408, opened 2025-07-09
feat: Adding particle spawning

svaningelgem:
# Description

Very simple implementation for particle spawning. As this is a purely visual feature, I just keep track inside the world which particles have been spawned so it can be checked by the developer.

# Checklist

Depends on the ticking introduced in #1400

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
@@ -2931,6 +2931,7 @@ public class WorldMock implements World
 	public void tick()
 	{
 		fullTime++;
+		gameTime++;
 	}
 
 }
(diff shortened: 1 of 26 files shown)


svaningelgem:
Hi @4everTheOne , this PR won't work until #1400 is merged as it depends on the "ticks" introduced in that one.

We're still waiting for your decision on lombok there ;-)

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -1887,103 +1890,104 @@ public boolean isGameRule(@Nullable String rule)
...
+	public List<SpawnedParticle> getSpawnedParticles()
+	{
+		return Collections.unmodifiableList(spawnedParticles);
+	}
+
+	public void clearSpawnedParticles()
+	{
+		spawnedParticles.clear();
+	}
+
```
4everTheOne: Since this methods do not override any method from Bukkit/Paper, can you please include some Javadocs for it please?

4everTheOne (commented):
Looks good to me, just have a small request :)

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
svaningelgem: done

4everTheOne (approved):

PR #1410, opened 2025-07-11
Centralize cloning logic in constructors

4everTheOne:
# Description

This is a small PR to cleanup the cloning mess. At the moment we have logic to clone objects in the clone method and in the constructor and this should not happen.

# Checklist


src/main/java/org/mockbukkit/mockbukkit/block/data/AgeableDataMock.java
```diff
@@ -37,4 +47,11 @@ public int getMaximumAge()
 		return this.getLimitationValue(BlockDataLimitation.Type.MAX_AGE);
 	}
 
+	@Override
+	@SuppressWarnings({"java:S2975", "java:S1182"})
```
4everTheOne, line 51: >@SuppressWarnings({"java:S2975", "java:S1182"})

This suppresses 2 sonar errors:

- **[java:S2975](https://sonarcloud.io/organizations/mockbukkit/rules?open=java%3AS2975&rule_key=java%3AS2975)**:
> "clone" should not be overridden

We are using the copy constructor as suggested, but Bukkit requires the use of this method to clone.

- **[java:S1182](https://sonarcloud.io/organizations/mockbukkit/rules?open=java%3AS1182&rule_key=java:S1182)**:
> Classes that override "clone" should be "Cloneable" and call "super.clone()"

We will not call super.clone, since we delegate that to constructor.

src/main/java/org/mockbukkit/mockbukkit/block/data/AgeableDataMock.java
```diff
@@ -37,4 +47,11 @@ public int getMaximumAge()
...
+	public @NotNull AgeableDataMock clone()
+	{
+		return new AgeableDataMock(this);
```
svaningelgem, line 54: this looks super repetitive. And easy to forget when adding a new subclass.
Couldn't you do something in the parent like "this.getClass()::new(this)"?

src/test/java/org/mockbukkit/mockbukkit/block/data/AgeableDataMockTest.java
```diff
@@ -100,4 +101,15 @@ void givenAge15Material()
 
 	}
 
+	@Test
+	void validateClone()
+	{
+		ageable.setAge(3);
+
+		@NotNull AgeableDataMock cloned = ageable.clone();
```
svaningelgem, line 108: This test doesn't really test anything?
I would do:
```
		AgeableDataMock cloned = ageable.clone();
		assertEquals(ageable, cloned);
		assertEquals(ageable.getAge(), cloned.getAge());
		ageable.setAge(3);
		assertNotEquals(ageable, cloned);
		assertNotEquals(ageable.getAge(), cloned.getAge());
```

src/main/java/org/mockbukkit/mockbukkit/block/data/AgeableDataMock.java
svaningelgem:
```java
@SneakyThrows
@SuppressWarnings("unchecked")
public T clone() {
   Constructor<? extends T> constructor = (Constructor<? extends T>) this.getClass()
       .getDeclaredConstructor(this.getClass());
   return constructor.newInstance(this);
}
```

svaningelgem: The benefit of this is that you don't have to implement it everywhere, and if the copy constructor is not implemented, it will fail for that class.

src/test/java/org/mockbukkit/mockbukkit/block/data/AgeableDataMockTest.java
svaningelgem: Generic testing could be something like this:
```java
@ParameterizedTest
@EnumSource(Material.class)
void testCloneForAllMaterials(Material material) {
    Object original = BlockDataMockFactory.mock(material);
    Object cloned = invokeClone(original);
    
    assertEquals(original, cloned);
    assertNotSame(original, cloned);
    
    // Find a setter method and test mutation
    Method setter = findFirstSetter(original.getClass());
    if (setter != null) {
        Object newValue = generateValueForSetter(setter);
        invokeSetter(setter, original, newValue);
        
        assertNotEquals(original, cloned);
    }
}

@SneakyThrows
private Object invokeClone(Object obj) {
    Method cloneMethod = obj.getClass().getMethod("clone");
    return cloneMethod.invoke(obj);
}

private Method findFirstSetter(Class<?> clazz) {
    return Arrays.stream(clazz.getMethods())
        .filter(method -> method.getName().startsWith("set"))
        .filter(method -> method.getParameterCount() == 1)
        .findFirst()
        .orElse(null);
}

@SneakyThrows
private void invokeSetter(Method setter, Object obj, Object value) {
    setter.invoke(obj, value);
}

private Object generateValueForSetter(Method setter) {
    Class<?> paramType = setter.getParameterTypes()[0];
    if (paramType == int.class) return 42;
    if (paramType == boolean.class) return true;
    if (paramType == String.class) return "test";
    // Add more types as needed
    return null;
}
```

Same principle would apply to the other tests.
That way the code will start to fail once someone forget to add the copy-constructor.

src/main/java/org/mockbukkit/mockbukkit/block/data/AgeableDataMock.java
4everTheOne: Thank you for the suggestion @svaningelgem . I'm not a big fan of reflection overall and i tend to avoid it. (For more details check this https://discord.com/channels/792754410576019477/809866105275023381/1317226104020402277 )

> Couldn't you do something in the parent like "this.getClass()::new(this)"?

Unfortunately i'm not seeing a way to do so, since the parent class is not required to have that constructor. With interface we can set a method as required but not the constructor itself.

src/test/java/org/mockbukkit/mockbukkit/block/data/AgeableDataMockTest.java
4everTheOne: Thank you for the feedback, i'll review the tests and include a test to remind to implement the clone :)

src/main/java/org/mockbukkit/mockbukkit/block/data/AgeableDataMock.java
svaningelgem: Yeah, but then the user needs to implement that interface. There needs to be some kind of check somewhere. Because I'm sure sooner rather than later someone is going to forget it...
I tend to agree with you on reflection, but I didn't know how else to enforce it. Precondition somewhere? Or looping the files (if there is no global registry available) and checking in a test they do implement that copy constructor?

4everTheOne: I think the best approach is like you suggested, iterate each matearial that is a block (isBlock() method) and execute the clone operation

svaningelgem: Some things are in registries, so you could lean on that. And say that it needs to implement the interface. Compile-time protection. But others you cannot do that sadly enough. Or at least not that I can see.

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/BundleMetaMock.java
```diff
@@ -43,7 +43,7 @@ public BundleMetaMock(@NotNull ItemMeta meta)
 
 		if (meta instanceof BundleMeta bundleMeta)
 		{
-			this.items = new ArrayList<>(bundleMeta.getItems());
+			this.items = new ArrayList<>(bundleMeta.getItems().stream().map(ItemStack::clone).toList());
```
svaningelgem, line 46: Can't comment on the right line, but shouldn't this be :
```
	public BundleMetaMock(@NotNull ItemMeta meta)
	public BundleMetaMock(@NotNull BundleMetaMock meta)
```
To be a correct copy-constructor?

I've seen this in a few places and am wondering why they are sometimes not a pure copy-constructor!

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ItemMetaMock.java
```diff
@@ -373,41 +372,10 @@ && isFireResistant() == meta.isFireResistant()
 	}
 
 	@Override
+	@SuppressWarnings({"java:S2975", "java:S1182"})
 	public ItemMetaMock clone()
 	{
-		try
```
svaningelgem, line 378: That's what I'm talking about ! 😎

src/main/java/org/mockbukkit/mockbukkit/profile/PlayerProfileMock.java
```diff
@@ -275,7 +275,7 @@ public boolean equals(Object obj)
 	}
 
 	@Override
-	@SuppressWarnings("MethodDoesntCallSuperMethod")
+	@SuppressWarnings({"java:S2975", "MethodDoesntCallSuperMethod"})
 	public org.bukkit.profile.@NotNull PlayerProfile clone()
```
svaningelgem, line 279: Shouldn't this not null come from jetbrains? Just a strange direct reference here.

src/test/java/org/mockbukkit/mockbukkit/block/data/BlockDataMockTest.java
```diff
@@ -262,6 +268,53 @@ void deserialize_missingFields()
...
+	{
+
+		/**
+		 * Unit test to validate that extend {@link BlockDataMock} implement the clone method.
+		 *
+		 * @param material The material to be used in the block state
+		 */
+		@ParameterizedTest
+		@MethodSource("getPossibleBlockData")
+		void givenPossibleBlockData(Material material)
```
svaningelgem, line 283: This method goes through the defined BlockDataMockFactory entries. Which is a good start.
However, it doesn't loop the inventory/meta entries.
For this I didn't see any factory aggregation. So I would still strongly suggest to use reflection to get all classes in that certain package to test the clone.

Under a test, I don't have any objections for using reflections. Especially not when it generalizes code for things that can be forgotten easily.

You actually demonstrated it already ;-) ⚠

There is no test for clone in `EnchantmentStorageMetaMockTest`. And yes, it's easy to verify. However, right now we're focused on it. But in next PRs we will not be, and it's easy to skip.

So I would still STRONGLY suggest you add reflection to test them all in an automated way. So you just point that method to a all the packages you want to have tested, and let it run each time - during tests.
*Bam* Noone can fuggetabotit!

src/main/java/org/mockbukkit/mockbukkit/profile/PlayerProfileMock.java
Thorinwasher: There's so many not null markers, and paper API uses multiple of them. I think this one might just be the same marker as the on used by paper api. We should really choose one annotation to use for this, and replace all with that

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/BundleMetaMock.java
4everTheOne: That's a great question, i noticed that too. Not sure if that's the main reason, but i saw that that is also happening in PaperMC:
imagem
Maybe it was a replication of this?

src/main/java/org/mockbukkit/mockbukkit/profile/PlayerProfileMock.java
4everTheOne: The way this anotation is set is so misleading! In reality it's a NotNull annotation from JetBrains the "org.bukkit.profile" is from the "PlayerProfile" here is clean version:
```java
public @NotNull org.bukkit.profile.PlayerProfile clone()
```

src/test/java/org/mockbukkit/mockbukkit/block/data/BlockDataMockTest.java
4everTheOne: I'll try to find a solution for that, let me check :)

4everTheOne: I've included a solution for that in: https://github.com/MockBukkit/MockBukkit/pull/1410/commits/867a36f64a8efa08460852a412c8ff2065015151#diff-5c61ae26def64da0c51688a66069e927c95b3f4754bbf31520c6b8f2aca24a69R1721-R1753
Please let me know if this solves it :)

svaningelgem: I've added some thoughts there

Thorinwasher (approved):
Looks good to me, we can merge after the merge conflicts have been fixed

PR #1411, opened 2025-07-13
fix: having a mixin with an injection didn't work.

svaningelgem:
# Description

```
@ExtendWith(MockBukkitExtension.class)
public class CommonMixinTest {
    @MockBukkitInject
    protected ServerMock server;
}

public class MyTest extends CommonMixinTest {
// here it fails to locate "server" to do the proper injection.
}
```

# Checklist


svaningelgem:
Already fixed with #1412 
(verified)

PR #1412, opened 2025-07-13
feat: allow for more injections

svaningelgem:
# Description

right now you can only inject the server, but what about player, world or plugins? Quite often used as well.
I extended the code to add this behaviour.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
@@ -25,7 +25,6 @@ import org.mockbukkit.mockbukkit.world.WorldMock;
 import java.lang.reflect.Field;
 import java.util.List;
 import java.util.Optional;
-import java.util.Set;
 import java.util.logging.Logger;
 
 /**
@@ -147,10 +146,6 @@ public class MockBukkitExtension implements TestInstancePostProcessor, TestInsta
 	private static final String HORIZONTAL_DIVIDER = "------------------------------------------------------------------------------------";
 
 	private final Logger logger = Logger.getLogger("MockBukkitExtension");
-	private final Set<Class<?>> serverSupportedTypes = Set.of(Server.class, ServerMock.class);
-	private final Set<Class<?>> playerSupportedTypes = Set.of(Player.class, PlayerMock.class);
-	private final Set<Class<?>> worldSupportedTypes = Set.of(World.class, WorldMock.class);
-	private final Set<Class<?>> pluginSupportedTypes = Set.of(Plugin.class, PluginMock.class);
 
 	private int playerCounter = 0;
 	private int worldCounter = 0;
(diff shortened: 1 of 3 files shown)


src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
```diff
@@ -93,35 +148,75 @@ public class MockBukkitExtension implements TestInstancePostProcessor, TestInsta
...
+			return MockBukkit.getOrCreateMock();
+		}
+		else if (playerSupportedTypes.contains(type))
+		{
+			final String playerName = "Player" + playerCounter++;
+			return MockBukkit.getOrCreateMock().addPlayer(playerName);
+		}
+		else if (worldSupportedTypes.contains(type))
+		{
+			final String worldName = "World" + worldCounter++;
```
Thorinwasher: Would it otherwise be possible to require a name parameter string in the annotation, which you could use here?
You could have it be optional in the annotation, and then require it only for certain fields, throwing some kind of exception when the test starts. That could possibly clarify where this name comes from, or what do you think?

src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
```diff
@@ -93,35 +148,75 @@ public class MockBukkitExtension implements TestInstancePostProcessor, TestInsta
...
+			if (mockObject != null)
+			{
+				FieldUtils.writeField(field, testInstance, mockObject, true);
+			}
 		}
 	}
 
+	private @Nullable Object createMockForType(Class<?> type)
+	{
+		if (serverSupportedTypes.contains(type))
```
Thorinwasher:
```suggestion
		if (type.isAssignableFrom(ServerMock.class))
```

src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
```diff
@@ -93,35 +148,75 @@ public class MockBukkitExtension implements TestInstancePostProcessor, TestInsta
...
+		{
+			return MockBukkit.getOrCreateMock();
+		}
+		else if (playerSupportedTypes.contains(type))
```
Thorinwasher:
```suggestion
		else if (type.isAssignableFrom(PlayerMock.class)
```

src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
```diff
@@ -93,35 +148,75 @@ public class MockBukkitExtension implements TestInstancePostProcessor, TestInsta
...
+		{
+			final String playerName = "Player" + playerCounter++;
+			return MockBukkit.getOrCreateMock().addPlayer(playerName);
+		}
+		else if (worldSupportedTypes.contains(type))
```
Thorinwasher:
```suggestion
		else if (type.isAssignableFrom(WorldMock.class))
```

src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
```diff
@@ -93,35 +148,75 @@ public class MockBukkitExtension implements TestInstancePostProcessor, TestInsta
...
+		{
+			final String worldName = "World" + worldCounter++;
+			return MockBukkit.getOrCreateMock().addSimpleWorld(worldName);
+		}
+		else if (pluginSupportedTypes.contains(type))
```
Thorinwasher:
```suggestion
		else if (type.isAssignableFrom(PluginMock.class))
```

src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
```diff
@@ -130,19 +225,26 @@ public void preDestroyTestInstance(ExtensionContext context)
...
-		final boolean paramIsCorrectType = parameterContext.getParameter().getType() == ServerMock.class;
+		final Class<?> paramType = parameterContext.getParameter().getType();
 		final boolean paramHasCorrectAnnotation = parameterContext.isAnnotated(MockBukkitInject.class);
-		return paramIsCorrectType && paramHasCorrectAnnotation;
+
+		return paramHasCorrectAnnotation && (
+				serverSupportedTypes.contains(paramType) ||
+						playerSupportedTypes.contains(paramType) ||
+						worldSupportedTypes.contains(paramType) ||
+						pluginSupportedTypes.contains(paramType)
```
Thorinwasher: See the isAssignableFrom method

Thorinwasher (commented):
You could probably use the isAssignableFrom method in Class. Other than that, I left more of a question.

src/main/java/org/mockbukkit/mockbukkit/MockBukkitExtension.java
svaningelgem: Added, pure optional behaviour. In most of my tests, I don't care what a player is called, I just care I have a player object ;). But if it's necessary, you can name them however you want.

Thorinwasher (approved):

PR #1413, opened 2025-07-13
fix: Support Multiple Potion Effects of the Same Type

svaningelgem:
# Support Multiple Potion Effects of the Same Type

## What's Changed
We've updated how Minecraft handles potion effects to match modern gameplay, where players can have multiple effects of the same type active at once (like having both a weak and strong regeneration effect).

## The Problem
Previously, adding a new potion effect would completely replace any existing effect of the same type. This didn't match how Minecraft actually works - if you drink a weak healing potion and then a strong one, both effects should be active, with the stronger one taking priority.

## The Solution
Now potion effects work like a priority system:
- **Strongest effects take precedence**: Higher amplifier wins, then longer duration
- **Effects stack in the background**: Weaker effects wait their turn instead of disappearing
- **Smooth transitions**: When a strong effect wears off, the next strongest automatically becomes active
- **Proper notifications**: Plugins get notified whenever the active effect changes

## Real-World Example
1. Player drinks Regeneration I (weak, 60 seconds)
2. Player drinks Regeneration III (strong, 10 seconds) 
3. Regeneration III is active (stronger wins)
4. After 10 seconds, Regeneration III expires
5. Regeneration I automatically becomes active for the remaining 50 seconds
6. Plugins are notified about both the expiration and the effect change

## Why This Matters
This change makes MockBukkit's potion system behave exactly like real Minecraft, ensuring plugins that manage potion effects work correctly in testing environments. It's especially important for plugins that track effect changes, manage custom potions, or implement complex buff systems.

The update maintains full backward compatibility - existing code continues to work, but now with more realistic and flexible potion effect behavior.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
@@ -43,6 +43,7 @@ import org.bukkit.potion.PotionEffect;
 import org.bukkit.potion.PotionEffectType;
 import org.bukkit.util.RayTraceResult;
 import org.bukkit.util.Vector;
+import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.Range;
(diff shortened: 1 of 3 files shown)


4everTheOne (approved):

4everTheOne:

PR #1417, opened 2025-07-18
Upgrade to paper 1.21.8

4everTheOne:
# Description

Upgrade to paper 1.21.8

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/BeeMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/BeeMock.java
@@ -5,6 +5,7 @@ import net.kyori.adventure.util.TriState;
 import org.bukkit.Location;
 import org.bukkit.entity.Bee;
 import org.bukkit.entity.EntityType;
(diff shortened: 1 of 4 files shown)


x9nico (approved):

4everTheOne:
Implemented until build number 6: https://papermc.io/downloads/paper

src/main/java/org/mockbukkit/mockbukkit/entity/BeeMock.java
```diff
@@ -160,6 +162,19 @@ public int getTicksSincePollination()
...
+	{
+		Preconditions.checkArgument(time >= 0, "Time since sting cannot be negative");
+		this.timeSinceSting = time;
+	}
+
+	@Override
+	public int getTimeSinceSting()
+	{
+		return this.timeSinceSting;
+	}
```
Thorinwasher, line 176: We have lombok now, so you could have potentially used that one instead

Thorinwasher (approved):

PR #1418, opened 2025-07-18
Fixing infinite effects

svaningelgem:
# Description

When you add a lower effect, it doesn't get added. It triggers a "CHANGE" event, but doesn't get added.
I also cleaned up the setup of this test class a little bit now.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
@@ -652,7 +652,15 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 
 		if (!event.isCancelled())
 		{
-			activeEffects.computeIfAbsent(type, k -> new PriorityQueue<>()).add(new ActivePotionEffect(effect));
+			var pq = activeEffects.computeIfAbsent(type, k -> new PriorityQueue<>());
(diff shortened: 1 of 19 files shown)


svaningelgem:
I will first get #1420 merged before proceeding with this one. As this combines 2 separate things.

src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -652,7 +652,15 @@ public EntityPotionEffectEvent addPotionEffect(@NotNull PotionEffect effect, Ent
 
 		if (!event.isCancelled())
 		{
-			activeEffects.computeIfAbsent(type, k -> new PriorityQueue<>()).add(new ActivePotionEffect(effect));
+			var pq = activeEffects.computeIfAbsent(type, k -> new PriorityQueue<>());
```
Thorinwasher, line 655: Var 😭 . Haha, just me being a stubborn man refusing to use that keyword

Thorinwasher (approved):

src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
svaningelgem: It makes coding sooooo much easier and cleaner!

PR #1361, opened 2025-05-05
Implement data components

Thorinwasher:
# Description
Doesn't handle default values at the moment, so I did not implement the getter part of this
# Checklist
The following items should be checked before the pull request can be merged.
- [ ] Code follows existing style.
- [ ] Unit tests added (if applicable).

PR #1426, opened 2025-07-23
Refactor item serialization to match bukkit behaviour

4everTheOne:
# Description
Refactor the serialization system to match the bukkit behaviour.

# Checklist


src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ColorableArmorMetaMock.java
```diff
@@ -34,20 +38,20 @@ public ColorableArmorMetaMock(ItemMeta meta)
...
-			this.color = colorableArmorMeta.isDyed() ? colorableArmorMeta.getColor().asRGB() : null;
+			this.color = colorableArmorMeta.isDyed() ? colorableArmorMeta.getColor() : null;
 		}
 	}
 
 	@Override
 	public @NotNull Color getColor()
 	{
-		return this.color == null ? DEFAULT_LEATHER_COLOR : Color.fromRGB(this.color & 0xFFFFFF);
+		return this.color == null ? DEFAULT_LEATHER_COLOR : this.color;
```
Thorinwasher: Bukkit color is mutable. Returning the color instance will lead to unwanted behavior

src/main/java/org/mockbukkit/mockbukkit/inventory/SerializableMeta.java
```diff
@@ -0,0 +1,230 @@
...
+{
+
+	static final String TYPE_FIELD = "meta-type";
+
+	static final ImmutableMap<Class<? extends ItemMetaMock>, String> classMap;
+	static final ImmutableMap<String, Function<Map<String, Object>, ? extends ItemMetaMock>> factoryMap;
+
+	static
+	{
+		classMap = ImmutableMap.<Class<? extends ItemMetaMock>, String>builder()
```
Thorinwasher: Mostly code style here, but you could just assign the classMap static constant directly with this builder chain

src/main/java/org/mockbukkit/mockbukkit/inventory/SerializableMeta.java
```diff
@@ -0,0 +1,230 @@
...
+				.put(SuspiciousStewMetaMock.class, "SUSPICIOUS_STEW")
+				// .put(EntityTagMetaMock.class, "ENTITY_TAG")
+				.put(CompassMetaMock.class, "COMPASS")
+				.put(BundleMetaMock.class, "BUNDLE")
+				// .put(MusicInstrumentMetaMock.class, "MUSIC_INSTRUMENT")
+				.put(OminousBottleMetaMock.class, "OMINOUS_BOTTLE")
+				.put(ItemMetaMock.class, "UNSPECIFIC")
+				.build();
+
+		final ImmutableMap.Builder<String, Function<Map<String, Object>, ? extends ItemMetaMock>> classConstructorBuilder = ImmutableMap.builder();
```
Thorinwasher: Also code style, but instead of having this large static block, I prefer just making a private method that returns the wanted object. That way you can clearly assign the factory map directly in it's definition, like this:
```java
static final ImmutableMap<String, Function<Map<String, Object>, ? extends ItemMetaMock>> factoryMap = compileFactoryMap(classMap);
```

src/main/java/org/mockbukkit/mockbukkit/inventory/SerializableMeta.java
```diff
@@ -0,0 +1,230 @@
...
+import java.util.function.Function;
+
+@SerializableAs("ItemMeta")
+public class SerializableMeta implements ConfigurationSerializable
+{
+
+	static final String TYPE_FIELD = "meta-type";
+
+	static final ImmutableMap<Class<? extends ItemMetaMock>, String> classMap;
+	static final ImmutableMap<String, Function<Map<String, Object>, ? extends ItemMetaMock>> factoryMap;
```
Thorinwasher: Should these be package private, or just private?

Thorinwasher (commented):
One thought that just crossed my mind, is to generate test files containing the serialized items on a paper server. Say you do this process:
- Serialize variations of items in a paper server
- Save the items into a file, or multiple files
- Deserialize each item data into an item
- Serialize the item data
- Compare the paper serialization to the MockBukkit serialization, check if something is missing.

I don't know, is that a good or bad idea?

Just going to note that the currently draft PR #1361 will require a major rewrite for this item serialization behavior.

src/main/java/org/mockbukkit/mockbukkit/inventory/SerializableMeta.java
4everTheOne: Thank you for the advice! It's a more elegant way for sure :)

4everTheOne:
> One thought that just crossed my mind, is to generate test files containing the serialized items on a paper server. Say you do this process:
> 
>     * Serialize variations of items in a paper server
> 
>     * Save the items into a file, or multiple files
> 
>     * Deserialize each item data into an item
> 
>     * Serialize the item data
> 
>     * Compare the paper serialization to the MockBukkit serialization, check if something is missing.
> 
> 
> I don't know, is that a good or bad idea?
> 
> Just going to note that the currently draft PR #1361 will require a major rewrite for this item serialization behavior.

That would be a little bit hard to achieve, since this would require "mutation" on each type of data. For example, for the chest it would require validating at least 3 cases: "Is water logged?", "What is the chest type?", "What is the direction he is looking at?"
I think the code would look ugly tbh :)

4everTheOne:
But if you see more use cases that i can include as test, please let me know

Thorinwasher:
> That would be a little bit hard to achieve, since this would require "mutation" on each type of data. For example, for the chest it would require validating at least 3 cases: "Is water logged?", "What is the chest type?", "What is the direction he is looking at?"
I think the code would look ugly tbh :)

Alright, then let's not, if you don't think it's viable

Thorinwasher:
Also, we are experiencing publishing issues with maven central, something about them changing their validation, so we have to delay merging this

Thorinwasher (approved):

PR #1428, opened 2025-07-26
chore: Adding `@ExtendWith` everywhere

svaningelgem:
# Description

Adding the `@ExtendWith(MockBukkitExtension.class)` and `@MockBukkitInject` annotations everywhere possible.
(major code cleanup: net ~2k lines of code removed)

I reformatted all the code just to make certain no checkStyle issues would appear.

A few tests had to be changed (mostly) in order to unload the created world. Because the injection mechanism places entities already inside a world (as it should be), but there were like 10 tests or so where creating a pure mock was tested (ie: one not placed in the world yet).

# Checklist


Thorinwasher (approved):

4everTheOne:
Should we release this as a patch instead? There are improvements in the main code

svaningelgem:
Yeah, I think that's best. I added (normally non breaking) changes to the injection annotations

Thorinwasher:
Whops I missed that, I only saw the metaminer patches, which does not change mockbukkit behavior

svaningelgem:
I added the ability to use the injection if all you have is a default constructor. No matter what type it is. That reduced a lot of defaults in the code too.

PR #1453, opened 2025-09-09
Improve BlockDataKey encoding and decoding

### LINKED ISSUE #1433: BlockData#getAsString() returns incorrect value

### Minecraft Version

1.21

### Describe the bug

The string returned by `BlockData.getAsString(false)` is incorrect. 

### Reproducible Test

For example, try
```
Material.BAMBOO_DOOR.createBlockData().getAsString(false)
```

This should return `minecraft:bamboo_door[facing=north,half=lower,hinge=left,open=false,powered=false]` but instead returns `minecraft:bamboo_door[facing=north,half=bottom,hinge=left,open=false,powered=false]`. Other similar blocks with two parts have the same issue.

half=bottom -> half=lower
half=top -> half=upper

### Anything else?

List of blocks this issue occurs for
```
StateType Mapping Errors:
State type mismatch for material BAMBOO_DOOR, type=air, value=bamboo_door
State type mismatch for material DARK_OAK_DOOR, type=air, value=dark_oak_door
State type mismatch for material WAXED_EXPOSED_COPPER_DOOR, type=air, value=waxed_exposed_copper_door
State type mismatch for material ACACIA_DOOR, type=air, value=acacia_door
State type mismatch for material OAK_DOOR, type=air, value=oak_door
State type mismatch for material IRON_DOOR, type=air, value=iron_door
State type mismatch for material WEATHERED_COPPER_DOOR, type=air, value=weathered_copper_door
State type mismatch for material LILAC, type=air, value=lilac
State type mismatch for material PITCHER_PLANT, type=air, value=pitcher_plant
State type mismatch for material CRIMSON_DOOR, type=air, value=crimson_door
State type mismatch for material LARGE_FERN, type=air, value=large_fern
State type mismatch for material PALE_OAK_DOOR, type=air, value=pale_oak_door
State type mismatch for material COPPER_DOOR, type=air, value=copper_door
State type mismatch for material JUNGLE_DOOR, type=air, value=jungle_door
State type mismatch for material SPRUCE_DOOR, type=air, value=spruce_door
State type mismatch for material WAXED_COPPER_DOOR, type=air, value=waxed_copper_door
State type mismatch for material SUNFLOWER, type=air, value=sunflower
State type mismatch for material BIRCH_DOOR, type=air, value=birch_door
State type mismatch for material PEONY, type=air, value=peony
State type mismatch for material MANGROVE_DOOR, type=air, value=mangrove_door
State type mismatch for material WAXED_WEATHERED_COPPER_DOOR, type=air, value=waxed_weathered_copper_door
State type mismatch for material ROSE_BUSH, type=air, value=rose_bush
State type mismatch for material WAXED_OXIDIZED_COPPER_DOOR, type=air, value=waxed_oxidized_copper_door
State type mismatch for material CHERRY_DOOR, type=air, value=cherry_door
State type mismatch for material TALL_SEAGRASS, type=air, value=tall_seagrass
State type mismatch for material WARPED_DOOR, type=air, value=warped_door
State type mismatch for material OXIDIZED_COPPER_DOOR, type=air, value=oxidized_copper_door
State type mismatch for material EXPOSED_COPPER_DOOR, type=air, value=exposed_copper_door
State type mismatch for material TALL_GRASS, type=air, value=tall_grass
```

4everTheOne:
# Description

Refactors the BlockDataKey to use the "Encoder" and "Decoder" concept, and implements a custom encoder for `HALF` to fix issue https://github.com/MockBukkit/MockBukkit/issues/1433.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataKey.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataKey.java
@@ -49,12 +49,15 @@ import org.bukkit.block.data.type.Wall;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.mockbukkit.mockbukkit.block.data.decoder.BooleanPropertyDecode;
+import org.mockbukkit.mockbukkit.block.data.decoder.EnumPropertyDecode;
+import org.mockbukkit.mockbukkit.block.data.decoder.IntegerPropertyDecode;
+import org.mockbukkit.mockbukkit.block.data.decoder.PropertyDecoder;
+import org.mockbukkit.mockbukkit.block.data.encoder.PropertyEncoder;
 
 import java.util.Arrays;
-import java.util.Locale;
 import java.util.Objects;
 import java.util.Set;
-import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
@@ -67,12 +70,12 @@ public enum BlockDataKey
 	/**
 	 * Stores the {@link BlockFace} a {@link Directional} block is facing towards.
 	 */
-	FACING("facing", string -> BlockFace.valueOf(string.toUpperCase(Locale.ROOT)), Directional.class::isInstance),
+	FACING("facing", EnumPropertyDecode.of(BlockFace.class), Directional.class::isInstance),
 
 	/**
 	 * Stores whether a {@link Campfire} is a signal fire (hay block underneath).
 	 */
-	SIGNAL_FIRE("signal_fire", Boolean::parseBoolean, Campfire.class::isInstance),
+	SIGNAL_FIRE("signal_fire", BooleanPropertyDecode.INSTANCE, Campfire.class::isInstance),
 
 	/**
 	 * Stores what {@link Bisected.Half} a {@link Bisected} block is placed in.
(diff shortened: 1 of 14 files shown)


src/main/java/org/mockbukkit/mockbukkit/block/data/decoder/BooleanPropertyDecode.java
```diff
@@ -0,0 +1,27 @@
...
+
+	@Override
+	public @Nullable Boolean decode(@Nullable Object value)
+	{
+		return switch (value)
+		{
+			case null -> null;
+			case Boolean bool -> bool;
+			case String string -> Boolean.parseBoolean(string);
+			default -> throw new UnsupportedOperationException("Unable to decode boolean from value: " + value);
```
Thorinwasher, line 22: I say illegal argument exception, but hey that's just a preference

src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataKey.java
```diff
@@ -65,132 +70,142 @@ public enum BlockDataKey
...
 	 * Stores whether a {@link Campfire} is a signal fire (hay block underneath).
 	 */
-	SIGNAL_FIRE("signal_fire", Boolean::parseBoolean, Campfire.class::isInstance),
+	SIGNAL_FIRE("signal_fire", BooleanPropertyDecode.INSTANCE, Campfire.class::isInstance),
 
 	/**
 	 * Stores what {@link Bisected.Half} a {@link Bisected} block is placed in.
 	 */
-	HALF("half", string -> Bisected.Half.valueOf(string.toUpperCase(Locale.ROOT)), Bisected.class::isInstance),
+	HALF("half",  BisectedDataMock.HalfDecoder.INSTANCE, BisectedDataMock.HalfEncoder.INSTANCE, Bisected.class::isInstance),
```
Thorinwasher, line 83: Apparently there are some edge cases for this specific property. It seems that single block bisected structures, such as stairs and trapdoors have a name related to the enum name of the bisected half, whilst multi block structures uses this naming convention. As per the bisected documentation (I also confirmed this with tests):

> 'half' denotes which half of a two block tall material this block is.
> In game it may be referred to as either (top, bottom) or (upper, lower).

I think you should be able to split the `HALF` enum element into `MULTI_BLOCK_HALF` and `SINGLE_BLOCK_HALF`, from my memory this should be achievable by just adding one more element and being more stringent with your class instance checks, but you might need to add more changes elsewhere.

src/main/java/org/mockbukkit/mockbukkit/block/data/decoder/NumberPropertyDecode.java
```diff
@@ -0,0 +1,31 @@
...
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+@NoArgsConstructor(access = AccessLevel.PROTECTED)
+public abstract class NumberPropertyDecode<A extends Number> implements PropertyDecoder<A>
```
Thorinwasher, line 11: This one might be redundant, same with the encoder; You only seem to have an integer representation of this, and I think that's really all you going to get with the block data format

src/main/java/org/mockbukkit/mockbukkit/block/data/encoder/PropertyEncoder.java
```diff
@@ -0,0 +1,24 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import javax.annotation.Nullable;
+
+@ApiStatus.Internal
+public interface PropertyEncoder<T>
+{
+
+	@Nullable
+	Object encode(@Nullable T value);
```
Thorinwasher, line 13: Why not this?
```suggestion
	String encode(@Nullable T value);
```
I don't understand why a non string value would be needed here

src/main/java/org/mockbukkit/mockbukkit/block/data/encoder/PropertyEncoder.java
```diff
@@ -0,0 +1,24 @@
+package org.mockbukkit.mockbukkit.block.data.encoder;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+import javax.annotation.Nullable;
+
+@ApiStatus.Internal
+public interface PropertyEncoder<T>
```
Thorinwasher, line 9: PropertySerializer?

From what I know, encode is when you go from object -> bytes, whilst serialize is when you go from object -> text. Haven't looked it up though.

On another note, it might actually be best to merge PropertyDecoder and PropertyEncoder into one class

src/main/java/org/mockbukkit/mockbukkit/block/data/encoder/PropertyEncoder.java
```diff
@@ -0,0 +1,24 @@
...
+{
+
+	@Nullable
+	Object encode(@Nullable T value);
+
+	default Object encodeIfPossible(@Nullable Object value)
+	{
+		Class<T> valueType = getValueType();
+		return valueType.isInstance(value) ? encode(valueType.cast(value)) : value;
+	}
```
Thorinwasher, line 19: From my point of view, this should return a defined object, like for example a string or null

src/main/java/org/mockbukkit/mockbukkit/block/data/decoder/PropertyDecoder.java
```diff
@@ -0,0 +1,14 @@
...
+import org.jetbrains.annotations.ApiStatus;
+
+import javax.annotation.Nullable;
+
+@ApiStatus.Internal
+public interface PropertyDecoder<T>
+{
+
+	@Nullable
+	T decode(@Nullable Object value);
```
Thorinwasher, line 12:
```suggestion
	T decode(@Nullable String value);
```
Easier to deal with if it's done like this

src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -227,12 +228,18 @@ protected <T> void set(@NotNull BlockDataKey key, @NotNull T value)
...
+		T value;
+		if (valueStored == null)
 		{
 			Object temp = BlockDataMockRegistry.getInstance().getDefault(getMaterial(), key.key());
 			value = castObject(key, temp);
 		}
+		else
+		{
+			value = (T) key.decode(valueStored);
+		}
```
Thorinwasher, line 242: I'm probably repeating myself, but why would you need to decode the value again if it's done in line 132?

src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
```diff
@@ -210,7 +210,8 @@ protected <T> void set(@NotNull BlockDataKey key, @NotNull T value)
 
 		checkProperty(key);
 
-		this.data.put(key.key(), value);
+		var valueToSave = key.encode(value);
+		this.data.put(key.key(), valueToSave);
```
Thorinwasher, line 214: Why encode the value here? Why not do it on block data serialization?

Thorinwasher (changes requested):
From what I understand, these change would sometimes serialize values to the data field in BlockData, but also sometimes not. Is there any way to avoid this?

I might be too entrenched in types for Java though, who knows. Some kind of discussion on this would be great though.

src/main/java/org/mockbukkit/mockbukkit/block/data/encoder/PropertyEncoder.java
4everTheOne: I changed from String to object, because of this:
imagem
There are some objects that are not string, like the case of boolean

src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
4everTheOne: You mean in the `getAsString` method? I don't see a way to map the key name to the corresponding block data key, because there are some cases that have the same name:
https://github.com/MockBukkit/MockBukkit/blob/584b2933e6e04590cda9b8d137a59f6737e58b98/src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataKey.java#L110-L114

4everTheOne (commented):
I replied to some of the comments. I will get back to the drawing board and try to find a better solution

src/main/java/org/mockbukkit/mockbukkit/block/data/BlockDataMock.java
Thorinwasher: Okay, but you should be able to fetch the correct key using `BlockDataKey#fromKey(String, BlockData)` later on in that method.

4everTheOne:
A improved solution was proposed at https://github.com/MockBukkit/MockBukkit/pull/1456

PR #1461, opened 2025-09-29
Implement PlayerInventory.setStorageContents method

benwoo1110:
# Description
Implement PlayerInventory.setStorageContents. This is for the first 36 inventory slot of player that excludes armor etc.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/PlayerInventoryMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/PlayerInventoryMock.java
@@ -9,8 +9,8 @@ import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.PlayerInventory;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jspecify.annotations.NonNull;
 import org.mockbukkit.mockbukkit.entity.HumanEntityMock;
-import org.mockbukkit.mockbukkit.exception.UnimplementedOperationException;
 
 import java.util.Arrays;
 
@@ -77,8 +77,25 @@ public class PlayerInventoryMock extends InventoryMock implements PlayerInventor
 	@Override
 	public void setStorageContents(ItemStack[] items) throws IllegalArgumentException
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException("setStorageContests has not been implemented for Player Inventories");
+		if (items == null)
+		{
+			throw new NullPointerException("ItemStack was null");
+		}
+		else if (items.length > 36)
+		{
+			throw new IllegalArgumentException("ItemStack array too large (max: 36, was: " + items.length + ")");
+		}
+		for (int i = 0; i < getSize(); i++)
+		{
+			if (i < items.length && items[i] != null)
+			{
+				setItem(i, items[i]);
+			}
+			else
+			{
+				setItem(i, null);
+			}
+		}
 	}
 
 	@Override
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/PlayerInventoryMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/PlayerInventoryMock.java
@@ -81,10 +81,7 @@ public class PlayerInventoryMock extends InventoryMock implements PlayerInventor
 		{
 			throw new NullPointerException("ItemStack was null");
 		}
-		else if (items.length > 36)
-		{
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockbukkit/mockbukkit/inventory/PlayerInventoryMock.java
```diff
@@ -77,8 +77,25 @@ public HumanEntity getHolder()
...
+		{
+			if (i < items.length && items[i] != null)
+			{
+				setItem(i, items[i]);
+			}
+			else
+			{
+				setItem(i, null);
+			}
+		}
```
4everTheOne: I noticed that a similar logic already exists in InventoryMock.
https://github.com/MockBukkit/MockBukkit/blob/4bd5c3396731cf0bc112a23523b40f054f336e53/src/main/java/org/mockbukkit/mockbukkit/inventory/InventoryMock.java#L328-L342
If we remove this method definition is it enough to fix it? If so, we can avoid duplicating the logic :)

benwoo1110: No, bcu the `getSize()` method in the parent class code is different, its for the entire inventory. So it will override the armor content as well, which it shouldn't for `setStorageContent`.

What we can do is create a protected method that is like `setSlots(ItemStack[] items, int size)` to extract out the common logic.

Thorinwasher: From my point of view, replicating a bit of code is fine to do here and there. 

This implementation looks like it clears all armor contents, as it iterates through the size of the whole inventory, is that intended behavior? (I don't have the time right now to compare with paper code or test this)

src/test/java/org/mockbukkit/mockbukkit/inventory/PlayerInventoryMockTest.java
Thorinwasher, line 1: Could you add a test here, if the armor contents get cleared or not if you use the set storage content method

Thorinwasher (changes requested):

src/test/java/org/mockbukkit/mockbukkit/inventory/PlayerInventoryMockTest.java
benwoo1110: yea sure, the armor contents should NOT get cleared

src/main/java/org/mockbukkit/mockbukkit/inventory/PlayerInventoryMock.java
benwoo1110: `setStorageContent` should not clear armour content, only the "storage area" which is the 3 rows and hotbar (i.e. 9 * 4 = 36), which is different from `setContent`

src/main/java/org/mockbukkit/mockbukkit/inventory/PlayerInventoryMock.java
```diff
@@ -77,8 +77,25 @@ public HumanEntity getHolder()
...
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException("setStorageContests has not been implemented for Player Inventories");
+		if (items == null)
+		{
+			throw new NullPointerException("ItemStack was null");
+		}
+		else if (items.length > 36)
+		{
+			throw new IllegalArgumentException("ItemStack array too large (max: 36, was: " + items.length + ")");
+		}
```
Thorinwasher:
```suggestion
		Preconditions.checkArgument(items.length <= 36, "ItemStack array too large (max: 36, was: " + items.length + ")");
```

src/main/java/org/mockbukkit/mockbukkit/inventory/PlayerInventoryMock.java
```diff
@@ -77,8 +77,25 @@ public HumanEntity getHolder()
 	@Override
 	public void setStorageContents(ItemStack[] items) throws IllegalArgumentException
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException("setStorageContests has not been implemented for Player Inventories");
+		if (items == null)
+		{
+			throw new NullPointerException("ItemStack was null");
+		}
```
Thorinwasher:
```suggestion
		Preconditions.checkNotNull(items, "ItemStack was null");
```

Thorinwasher (commented):
You can also use the preconditions util class.

4everTheOne (approved):

Thorinwasher (approved):
Looks good to me

Issue #853, opened 2023-09-19
WorldMock#isDayTime() always returning false


thelooter:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Are you using the latest version of MockBukkit?

- [X] I am using the latest version of MockBukkit.

### Minecraft Version

1.2

### Describe the bug

`WorldMock#isDayTime()` always returns false. This is most likely due to it being missed while Stubbing and therefore just receiving the default Implementation by the IDE

### Reproducible Test

```java
@Test
void testIsDayTime()
{
	WorldMock world = new WorldMock(Material.DIRT, 3);
	world.setTime(1000);
	assertTrue(world.isDayTime());
}
```

Nirbose:
True
https://github.com/MockBukkit/MockBukkit/blob/5bc30ee7864b9679eb0c4c9d047234233ff37ead/src/main/java/be/seeseemelk/mockbukkit/WorldMock.java#L1432-L1435

Thorinwasher:
One easy fix is to just throw UnimplementedOperationException for this. Otherwise we could link this to the scheduler or something? I really don't know exactly how this works though, is the daytime linked to a time in the world or one in the server/scheduler?

JustNOTRO:
I really don't know if this issue is still relevant, but a possible solution would be to create a global variable 'dayTime' in WorldMock and set it to true when invoking WorldMock#setTime() if value 1000 (or other day values) occur . Then return this variable on the isDayTime method.

You will obviously want to do the same when trying set time to night (updating it to false).

4everTheOne:
Fixed by https://github.com/MockBukkit/MockBukkit/pull/1465

PR #1465, opened 2025-09-30
Implement WorldMock.isDayTime() method

Mertzenich:
# Description

This pull request aims to resolve issue #853 by implementing `WorldMock.isDayTime()`. Unintuitively, Bukkit's `World.isDayTime()` method corresponds to the players ability to sleep, not the actual game day/night times. For example, calling `isDayTime()` during a thunderstorm will return `false` even if the time is noon (6000t). My implementation logic makes use of the sleep tick-times found on the Minecraft Wiki (https://minecraft.wiki/w/Bed#Sleeping). I manually verified the accuracy of this solution by writing a small plugin which asserts every tick that my mock implementation matches the result of Bukkit's `World.isDayTime()` method.

# Checklist


4everTheOne (approved):

PR #1474, opened 2025-10-10
Fix BlockData -> Material mappings for CAVE_AIR, VOID_AIR, and SOUL_FIRE

Axionize:
# Checklist


4everTheOne:
Thank you for your contribution! ~~Can you please include unit tests for this items? Thanks in advance~~
After looking into the code it's already covered by the tests of block data.

4everTheOne:
Could you elaborate a little bit more on the issue you are trying to solve?

Axionize:
Material.CAVE_AIR.getBlockData().getMaterial() returns the wrong material. Likewise for VOID_AIR and SOUL_FIRE.

It constructs an instance of BlockData that returns Material.AIR for each of them. Instead of the correct respective materials.

I found this out when my tests to verify PacketEvent's block mappings were falling once again. I'm sure your remember that issue.

https://github.com/MockBukkit/MockBukkit/issues/1433

The same test that caught that issue also caught this one.

4everTheOne:
Thank you for the clarification and for pointing this out. Yes i remember that issue, and the problem seems to be related with Metaminer generator. It is generating a incorrect output. I have opened this PR https://github.com/Axionize/MockBukkit/pull/1 into this branch to fix the unit tests, let me know if you agree with it.

Axionize:
Oh so this is a result of code that was generated? Interesting.

Well LGTM, anything else we need to do to fix this permanntly?

Thorinwasher:
> Oh so this is a result of code that was generated? Interesting.

Yeah, we need to do that to be able to keep up with all the paper api changes. Just JSON files for what I know. We could also start generating java classes with something like java poet, but nah.

Thorinwasher (approved):
LGTM 👍

Thorinwasher (approved):
LGTM 👍

PR #1476, opened 2025-10-10
feat(blockdata): Add GlassPaneDataMock with unit tests

thelooter:
- Implemented `GlassPaneDataMock` for `GlassPane` block data.
- Created corresponding unit tests in `GlassPaneDataMockTest`.
- Enabled previously disabled test for `GlassPane`.
- Updated `BlockDataMockFactory` to support `GlassPane`.

# Description

# Checklist


src/main/java/org/mockbukkit/mockbukkit/block/data/GlassPaneDataMock.java
```diff
@@ -0,0 +1,93 @@
...
+	{
+		try
+		{
+			return Optional.of(BlockDataKey.valueOf(blockFace.name()));
+		}
+		catch (IllegalArgumentException e)
+		{
+			return Optional.empty();
+		}
+	}
```
4everTheOne, line 92: [Open for discussion] Maybe we could replace this with a switch? I think that it would improve the code readability.

4everTheOne (commented):
Looks good to me! Just have a small suggestion that is open to discussion

4everTheOne:
(Relates with https://github.com/MockBukkit/MockBukkit/issues/1088)

src/main/java/org/mockbukkit/mockbukkit/block/data/GlassPaneDataMock.java
thelooter: This was actually just copied from the Fencedata mock. Might be worth abstracting away into a utility class.

4everTheOne (approved):

PR #1486, opened 2025-10-20
Enforce maxPlayers limit when adding players (PlayerListMock#addPlayer)

walker84837:
# Description

This pull request enforces the `maxPlayers` limit in `PlayerListMock`, as it wasn't being enforced, allowing more players to be added than the specified maximum.

The changes include:
- Adding a check to the `addPlayer` method to stop more players being added once the maximum has been reached.
- Introducing a new unit test, to verify the `maxPlayera` limit is being enforced correctly.
- Consequently, removing outdated comments (and respective Javadoc line) indicating that the `maxPlayers` limit was not enforced.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/PlayerListMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/PlayerListMock.java
@@ -47,13 +47,11 @@ public class PlayerListMock
 
 	/**
 	 * Sets the maximum number of online players.
-	 * <b>This is not currently enforced.</b>
 	 *
 	 * @param maxPlayers The maximum amount of players.
 	 */
 	public void setMaxPlayers(int maxPlayers)
 	{
-		// TODO: The maxPlayers setting is currently not enforced.
 		this.maxPlayers = maxPlayers;
 	}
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockbukkit/mockbukkit/PlayerListMock.java
```diff
@@ -91,6 +89,10 @@ public ProfileBanListMock getProfileBans()
 	@ApiStatus.Internal
 	public synchronized void addPlayer(@NotNull PlayerMock player)
 	{
+		if (this.onlinePlayers.size() >= this.maxPlayers)
+		{
+			return;
```
4everTheOne: Can we update this method to return a Boolean so that when true, it means that the player was added successfully and when false it was not?
I think it may be useful to identify if a player was added or not

4everTheOne (commented):
Looks good to me, just a suggestion that is open for discussion

src/main/java/org/mockbukkit/mockbukkit/PlayerListMock.java
walker84837: Sure, that sounds like a good idea. I'll update it so that we can easily tell if a player has been added successfully.

4everTheOne (approved):

4everTheOne:
Thanks for your contribution!

PR #1487, opened 2025-10-28
Small improvements to block mock

4everTheOne:
# Description

This PR adds a small cleanup to the `BlockMock`

# Checklist


src/main/java/org/mockbukkit/mockbukkit/block/BlockMock.java
```diff
@@ -407,37 +394,25 @@ public boolean isEmpty()
...
 	public boolean isSolid()
 	{
-		if (InternalTag.SOLID_BLOCKS.isTagged(this.getType()))
-		{
-			return true;
-		}
-		else if (InternalTag.NON_SOLID_BLOCKS.isTagged(this.getType()))
-		{
-			return false;
-		}
```
Thorinwasher, line 424: This has always been bugging me for a while. I don't think we need the solid/non solid tag anymore, unsure if we ever did

Thorinwasher (commented):
Whops, I forgot to send this message. Looks good to me, I think we need to remove the solid non solid tags, pretty much anything to do with internal tags. Should that be in a separate PR?

4everTheOne:
I can include it here, i'll update it later

Thorinwasher (approved):

PR #1490, opened 2025-10-29
feat(blockdata): Implement CaveVines Blockdata

thelooter:
# Description

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/block/data/CaveVineDataMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/block/data/CaveVineDataMock.java
@@ -26,12 +26,6 @@ public class CaveVineDataMock extends AgeableDataMock implements CaveVines
 	}
 
 	@Override
-	public boolean isBerries()
-	{
-		return super.get(BERRIES);
-	}
-
-	@Override
 	public void setBerries(boolean berries)
 	{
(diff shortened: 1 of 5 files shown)


src/main/java/org/mockbukkit/mockbukkit/block/data/CaveVineDataMock.java
```diff
@@ -0,0 +1,46 @@
...
+	public boolean hasBerries()
+	{
+		return super.get(BERRIES);
+	}
+
+	@Override
+	public boolean isBerries()
+	{
+		return super.get(BERRIES);
+	}
```
4everTheOne: We don't need to implement this, since the `isBerries()` is deprecated and has a default implementation:
```java
    /**
     * Gets the value of the 'berries' property.
     *
     * @return the 'berries' value
     * @deprecated bad name, use {@link #hasBerries()}
     */
    @Deprecated
    default boolean isBerries() {
        return this.hasBerries();
    }
```

4everTheOne:
```suggestion
```

4everTheOne (commented):
Looks good to me, i just have a nitpick comment

4everTheOne (approved):

PR #1493, opened 2025-10-31
Add support for ItemStack components

4everTheOne:
# Description

Add basic support for ItemStack components

# Checklist


Thorinwasher:
My try on that is this one https://github.com/MockBukkit/MockBukkit/pull/1361

Might be a better idea to just go with the simple solution though, which is to just store a map with values like is done here

Thorinwasher (approved):

Thorinwasher:
Just going step by step is definitely a better idea, we can just apply #1361 step by step

PR #1497, opened 2025-11-06
Add Paper SHA-512 to tag

### LINKED ISSUE #1470: Paper hash in release?

### Describe the feature

Quick idea: what if we include the SHA-512 hash of the Paper API in our release text? It could help people verify exactly which Paper version we built against, especially since it's a snapshot and not always obvious.

Here's a Kotlin function that grabs the SHA-512 from a resolved Gradle dependency:

```kotlin
fun Project.getDependencySha512(
    group: String,
    name: String,
    version: String,
    configurationName: String = "compileClasspath"
): String? {
    val artifacts = configurations.getByName(configurationName)
        .resolvedConfiguration.resolvedArtifacts

    val match = artifacts.firstOrNull {
        it.moduleVersion.id.group == group &&
            it.moduleVersion.id.name == name &&
            it.moduleVersion.id.version == version
    } ?: return null

    val sha512 = MessageDigest.getInstance("SHA-512")
        .digest(match.file.readBytes())
        .joinToString("") { "%02x".format(it) }

    return sha512
}
```

We could use it like this:

```kotlin
val hash = getDependencySha512(
    group = "io.papermc.paper",
    name = "paper-api",
    version = "1.21.8-R0.1-SNAPSHOT"
)
```

Then include the hash in the release tag like:

```
## MockBukkit v1.2.3

- Paper API: 1.21.8-R0.1-SNAPSHOT
- Paper API SHA-512: dd7d535ed7e2f7504c8683d7cfd0ca89e88bb923553bed0172e8214d70946136e74a147505dffec8f593d741fb40875e41a2a8a4ef98479c0ef266cae652cf81
```

Would make it easier to track and verify builds. What do you think?

4everTheOne:
# Description
This PR adds the SHA-512 for paper-api in the tag (https://github.com/MockBukkit/MockBukkit/issues/1470).

Sample:
---

### v4.XX.0

Triggered by: PR 1496
Commit: 2de32ce5

Paper version: 1.21
Paper full version: 1.21.10-R0.1-SNAPSHOT
Paper resolved version: io.papermc.paper:paper-api:1.21.10-R0.1-SNAPSHOT:20251104.194106-38
Paper SHA-512: 6615ff81f9e2217699120faefe3b0665c771763a3a2eebff587991e8bb8596f63ec1f718c0790617ed30e951ac286856dd237491806d7a8618d28f6765269563

---

# Checklist


4everTheOne:
With the `Paper resolved version` we can uniquely identify the paper version, so let me know if you think that the hash is needed.

Thorinwasher (approved):

PR #1509, opened 2025-12-23
Avoid possible deadlocks in scheduler

frengor:
# Description

The current implementation performs the entirety of the server ticking logic while holding a `synchronized` lock (on the scheduler instance). This can lead to deadlocks when scheduling and running tasks with async threads.

An example of possible deadlock is the following (assume there exists a lock named `lock`):

**Main thread**
```java
runTask(myPlugin, () -> {
    lock.lock();
    // ...
    lock.unlock();
});
performOneTick();
```

**Async thread**
```java
lock.lock();
runTask(myPlugin, () -> {});
lock.unlock();
```

A possible execution which results in a deadlock is:
- Main thread registers the task
- Async thread takes the `lock` lock
- Main thread calls `performOneTick()`, taking the lock on the scheduler instance
- Main thread runs the registered task and tries to take the `lock` lock, which is held by the async thread
- Async thread calls `runTask` and tries to take the lock on the scheduler instance, which is held by the main thread
- Deadlock!

To prevent deadlocks, this PR reduces the scope of the synchronized blocks, making sure no user-provided task is run while holding a lock.

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
@@ -238,9 +238,17 @@ public class BukkitSchedulerMock implements BukkitScheduler
 	/**
 	 * Perform one tick on the server.
 	 */
-	public synchronized void performOneTick()
+	public void performOneTick()
 	{
-		currentTick++;
+		// Using a synchronized block here is necessary to avoid tasks being skipped by processTasks().
+		// runTask*(..) methods will synchronize with this synchronized block and keep the lock from the moment they read
+		// currentTick to when they register their task into scheduledTasks. This ensures that the main thread will wait
+		// for the async threads to register their tasks before incrementing currentTick. Thus, since the delay of every
+		// task is always >= 1, tasks scheduled during a tick can never be skipped
+		synchronized (this)
+		{
+			currentTick++;
+		}
 
 		processWorlds();
 		processEntities();
@@ -286,17 +294,12 @@ public class BukkitSchedulerMock implements BukkitScheduler
 	public void waitAsyncTasksFinished()
 	{
 		// Cancel repeating tasks so they don't run forever.
-		synchronized (scheduledTasks.tasks)
-		{
-			scheduledTasks.tasks.entrySet().stream()
-					.filter(entry -> entry.getValue() instanceof RepeatingTask)
-					.forEach(entry -> scheduledTasks.cancelTask(entry.getKey()));
+		scheduledTasks.cancelRepeatingTasks();
 
-			// Make sure all tasks get to execute. (except for repeating asynchronous tasks, they only will fire once)
-			while (scheduledTasks.getScheduledTaskCount() > 0)
-			{
-				performOneTick();
-			}
+		// Make sure all tasks get to execute. (except for repeating asynchronous tasks, they only will fire once)
+		while (scheduledTasks.getScheduledTaskCount() > 0)
+		{
+			performOneTick();
 		}
 
 		// Wait for all tasks to finish executing.
@@ -372,45 +375,51 @@ public class BukkitSchedulerMock implements BukkitScheduler
 	}
 
 	@Override
-	public synchronized @NotNull BukkitTask runTask(@NotNull Plugin plugin, @NotNull Runnable task)
+	public @NotNull BukkitTask runTask(@NotNull Plugin plugin, @NotNull Runnable task)
 	{
 		return runTaskLater(plugin, task, 1L);
 	}
 
 	@Override
 	@Deprecated(since = "1.7.10")
-	public synchronized @NotNull BukkitTask runTask(@NotNull Plugin plugin, @NotNull BukkitRunnable task)
+	public @NotNull BukkitTask runTask(@NotNull Plugin plugin, @NotNull BukkitRunnable task)
 	{
 		return runTask(plugin, (Runnable) task);
 	}
 
 	@Override
-	public synchronized @NotNull BukkitTask runTaskLater(@NotNull Plugin plugin, @NotNull Runnable task, long delay)
+	public @NotNull BukkitTask runTaskLater(@NotNull Plugin plugin, @NotNull Runnable task, long delay)
 	{
 		delay = Math.max(delay, 1);
-		ScheduledTask scheduledTask = new ScheduledTask(id.getAndIncrement(), plugin, true, currentTick + delay, task);
-		scheduledTasks.addTask(scheduledTask);
-		return scheduledTask;
+		synchronized (this) // See comment at the start of performOneTick()
+		{
+			ScheduledTask scheduledTask = new ScheduledTask(id.getAndIncrement(), plugin, true, currentTick + delay, task);
+			scheduledTasks.addTask(scheduledTask);
+			return scheduledTask;
+		}
 	}
 
 	@Override
-	public synchronized @NotNull BukkitTask runTaskTimer(@NotNull Plugin plugin, @NotNull Runnable task, long delay, long period)
+	public @NotNull BukkitTask runTaskTimer(@NotNull Plugin plugin, @NotNull Runnable task, long delay, long period)
 	{
 		delay = Math.max(delay, 1);
-		RepeatingTask repeatingTask = new RepeatingTask(id.getAndIncrement(), plugin, true, currentTick + delay, period, task);
-		scheduledTasks.addTask(repeatingTask);
-		return repeatingTask;
+		synchronized (this) // See comment at the start of performOneTick()
+		{
+			RepeatingTask repeatingTask = new RepeatingTask(id.getAndIncrement(), plugin, true, currentTick + delay, period, task);
+			scheduledTasks.addTask(repeatingTask);
+			return repeatingTask;
+		}
 	}
 
 	@Override
 	@Deprecated(since = "1.7.10")
-	public synchronized @NotNull BukkitTask runTaskTimer(@NotNull Plugin plugin, @NotNull BukkitRunnable task, long delay, long period)
+	public @NotNull BukkitTask runTaskTimer(@NotNull Plugin plugin, @NotNull BukkitRunnable task, long delay, long period)
 	{
 		return runTaskTimer(plugin, (Runnable) task, delay, period);
 	}
 
 	@Override
-	public synchronized int scheduleSyncDelayedTask(@NotNull Plugin plugin, @NotNull Runnable task, long delay)
+	public int scheduleSyncDelayedTask(@NotNull Plugin plugin, @NotNull Runnable task, long delay)
 	{
 		Logger.getLogger(LOGGER_NAME).warning("Consider using runTaskLater instead of scheduleSyncDelayTask");
 		return runTaskLater(plugin, task, delay).getTaskId();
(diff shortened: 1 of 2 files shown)


Thorinwasher:
I don't think I understand this PR exactly, I just wrote a test like this:
```java
	@Test
	void aTest(){
		MockBukkit.mock();
		BukkitSchedulerMock schedulerMock = (BukkitSchedulerMock) Bukkit.getScheduler();
		Lock lock = new ReentrantLock();
		JavaPlugin myPlugin = MockBukkit.createMockPlugin();
		schedulerMock.runTaskAsynchronously(myPlugin, () -> {
			lock.lock(); // Locks the lock
			schedulerMock.runTask(myPlugin, () -> {});
		});
		schedulerMock.runTask(myPlugin, () -> {
			// Will be called by performOneTick while holding the lock on the scheduler instance
			lock.lock(); // Tries to lock the lock (already held by the async thread)
		});
		schedulerMock.performOneTick();
	}
```
This causes a deadlock. The problem is that this seems to cause a deadlock with your changes as well, it kind of makes sense that this would be an issue regardless of the scheduler implementation, as you need to actively release a lock if you have locked it.

Could you just provide a test where these changes would be applicable? If it actually fixes something, I will give it a go, and merge it.

frengor:
My example was only meant to show an execution that deadlocks, rather than a full executable example, so for the sake of brevity I didn't include the unlocking stuff. I edited the PR description with a better example.

A test that deadlocks without this PR is the following:
```java
	@Test
	void deadlockTest() {
		MockBukkit.mock();
		BukkitSchedulerMock schedulerMock = (BukkitSchedulerMock) Bukkit.getScheduler();
		Lock lock = new ReentrantLock();
		CyclicBarrier barrier = new CyclicBarrier(2);
		JavaPlugin myPlugin = MockBukkit.createMockPlugin();
		schedulerMock.runTaskAsynchronously(myPlugin, () -> {
			lock.lock();
			try {
				barrier.await();
			} catch (Exception e) {
			}
			schedulerMock.runTask(myPlugin, () -> {});
			lock.unlock(); // Never reached with old implementation
		});
		schedulerMock.runTask(myPlugin, () -> {
			try {
				barrier.await();
			} catch (Exception e) {
			}
			lock.lock();
			lock.unlock(); // Never reached with old implementation
		});
		schedulerMock.performOneTick();
	}
```

Thorinwasher:
Thanks for providing an example, I think I understand a bit now on how your example would deadlock!

I will just write down my thoughts on what's happening there.

- "Async" thread locks the lock
- "Async" thread waits until the "main" thread is running all tasks through the method performOneTick
- After that, it tries to schedule a task, but as that method and the performOneTick is marked as syncronized as well, it will wait indefinitely, because the "main" thread tries to lock the lock. Both threads are waiting for each other (one on the lock, and another one on the syncronized lock), therefore achieving a deadlock.

So by not having all methods in the scheduler perform a lock on the scheduler instance, most of these issues would be avoided. I think I am ready to do some code review on this. Thanks again for showing me something new :smile:

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
Thorinwasher, line 259: I think it might still be a good idea to keep this specific method syncronized. Process worlds, process entities, process tasks should be done only at one thread at a time.

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -372,91 +375,97 @@ public void waitAsyncEventsFinished()
...
 		delay = Math.max(delay, 1);
-		ScheduledTask scheduledTask = new ScheduledTask(id.getAndIncrement(), plugin, true, currentTick + delay, task);
-		scheduledTasks.addTask(scheduledTask);
-		return scheduledTask;
+		synchronized (this) // See comment at the start of performOneTick()
+		{
+			ScheduledTask scheduledTask = new ScheduledTask(id.getAndIncrement(), plugin, true, currentTick + delay, task);
+			scheduledTasks.addTask(scheduledTask);
+			return scheduledTask;
+		}
```
Thorinwasher, line 402: Couldn't this perform a lock on scheduledTasks instead?

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -372,91 +375,97 @@ public void waitAsyncEventsFinished()
...
 		delay = Math.max(delay, 1);
-		RepeatingTask repeatingTask = new RepeatingTask(id.getAndIncrement(), plugin, true, currentTick + delay, period, task);
-		scheduledTasks.addTask(repeatingTask);
-		return repeatingTask;
+		synchronized (this) // See comment at the start of performOneTick()
+		{
+			RepeatingTask repeatingTask = new RepeatingTask(id.getAndIncrement(), plugin, true, currentTick + delay, period, task);
+			scheduledTasks.addTask(repeatingTask);
+			return repeatingTask;
+		}
```
Thorinwasher, line 414: Same here, might be better to lock on scheduledTasks.

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -544,95 +553,119 @@ public boolean isQueued(int taskId)
...
-				new AsyncRunnable(task));
-		scheduledTasks.addTask(scheduledTask);
-		return scheduledTask;
+		synchronized (this) // See comment at the start of performOneTick()
+		{
+			ScheduledTask scheduledTask = new ScheduledTask(id.getAndIncrement(), plugin, false, currentTick + delay,
+					new AsyncRunnable(task));
+			scheduledTasks.addTask(scheduledTask);
+			return scheduledTask;
+		}
```
Thorinwasher, line 594: Same here, might be better to lock on scheduledTasks.

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -544,95 +553,119 @@ public boolean isQueued(int taskId)
...
-				new AsyncRunnable(task));
-		scheduledTasks.addTask(scheduledTask);
-		return scheduledTask;
+		synchronized (this) // See comment at the start of performOneTick()
+		{
+			RepeatingTask scheduledTask = new RepeatingTask(id.getAndIncrement(), plugin, false, currentTick + delay, period,
+					new AsyncRunnable(task));
+			scheduledTasks.addTask(scheduledTask);
+			return scheduledTask;
+		}
```
Thorinwasher, line 613: Same here, might be better to lock on scheduledTasks.

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -544,95 +553,119 @@ public boolean isQueued(int taskId)
...
+	public void runTaskLater(@NotNull Plugin plugin, @NotNull Consumer<? super BukkitTask> task, long delay)
 	{
 		delay = Math.max(delay, 1);
-		ScheduledTask scheduledTask = new ScheduledTask(id.getAndIncrement(), plugin, true, currentTick + delay, task);
-		scheduledTasks.addTask(scheduledTask);
+		synchronized (this) // See comment at the start of performOneTick()
+		{
+			ScheduledTask scheduledTask = new ScheduledTask(id.getAndIncrement(), plugin, true, currentTick + delay, task);
+			scheduledTasks.addTask(scheduledTask);
+		}
```
Thorinwasher, line 649: Same here, might be better to lock on scheduledTasks.

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
```diff
@@ -544,95 +553,119 @@ public boolean isQueued(int taskId)
...
+	public void runTaskTimer(@NotNull Plugin plugin, @NotNull Consumer<? super BukkitTask> task, long delay, long period)
 	{
 		delay = Math.max(delay, 1);
-		RepeatingTask repeatingTask = new RepeatingTask(id.getAndIncrement(), plugin, true, currentTick + delay, period, task);
-		scheduledTasks.addTask(repeatingTask);
+		synchronized (this) // See comment at the start of performOneTick()
+		{
+			RepeatingTask repeatingTask = new RepeatingTask(id.getAndIncrement(), plugin, true, currentTick + delay, period, task);
+			scheduledTasks.addTask(repeatingTask);
+		}
```
Thorinwasher, line 667: Same here, might be better to lock on scheduledTasks.

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
Thorinwasher, line 258: With the above mentioned changes, plus this. I think you also get thread safety on performOneTick.

Inside process tasks:
```java
syncronized (scheduledTasks) {
		List<ScheduledTask> oldTasks = scheduledTasks.getCurrentTaskList();
}
// Use tasks outside syncronized, that way it will not deadlock.
```

Thorinwasher (commented):
I'm honestly not that well used to dealing thread locks. I think your current changes on specifically performOneTick might break something, as that method technically isn't thread safe. I think it might be better to keep that method syncronized (meaning it will lock with the scheduler instance) and lock all task scheduling things to the scheduledTasks object.

This is more of a hypothetical, as performOneTick should already not be used in multiple threads. I would also not be surprised if I messed up with this, as I said, I'm not used to this kind of thinking.

frengor:
> I think your current changes on specifically performOneTick might break something, as that method technically isn't thread safe.
> [...]
> This is more of a hypothetical, as performOneTick should already not be used in multiple threads.

I would say that calling `performOneTick` in any other thread than the main one is already incorrect, as most of MockBukkit implementation isn't thread-safe. Thus, I don't think not having the entire method synchronized is an issue. I think it would even make sense to have `AsyncCatcher.catchOp("server tick")` as the first line of `performOneTick` to catch incorrect usages.

> I think it might be better to keep that method syncronized (meaning it will lock with the scheduler instance) and lock all task scheduling things to the scheduledTasks object.

Keeping the entire `performOneTick` synchronized was the reason of the deadlock, so I am a bit scared of keeping the synchronized lock there. To be 100% safe, the tasks should never be run while holding a lock, although using a separate lock for task scheduling stuff would probably be enough to avoid deadlocks (I might be wrong here though).

Aside from that, if you'd prefer to have a separate object to synchronize on for the task scheduling stuff instead of the scheduler instance I can do that.

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
frengor: Yes, but only if the [synchronized block](https://github.com/frengor/MockBukkit/blob/df666d77bff2c24d715fa13eb62b67744d0ec471/src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java#L248) in `performOneTick` also uses `scheduledTasks`, as we need to prevent `performOneTick` from incrementing `currentTick` in between the creation of the new task and its addition to scheduledTasks.

src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java
frengor: I think this synchronized block is not needed, since `scheduledTasks` (or better the `TaskList` class) is already thread-safe (i.e. that synchronization already happens inside the `TaskList` class).

Thorinwasher:
> I think it would even make sense to have AsyncCatcher.catchOp("server tick") as the first line of performOneTick to catch incorrect usages.

Honestly my first thought, any other thread should throw an error. I think we should just go with that, there might be some code that breaks from that though :monocle_face: .

> Aside from that, if you'd prefer to have a separate object to synchronize on for the task scheduling stuff instead of the scheduler instance I can do that.

Feel free to do what you want. I think throwing an error might be the better idea.

Thorinwasher:
> Keeping the entire performOneTick synchronized was the reason of the deadlock, so I am a bit scared of keeping the synchronized lock there.

Could alternatively lock on another object.

frengor:
> Feel free to do what you want. I think throwing an error might be the better idea.

Yeah I think I'll go the throwing an error route, I think it's safer.

> there might be some code that breaks from that though 🧐 .

IMO, if someone is calling performOneTick from different threads they've already messed up badly. I think an error is better, at least they'll be aware.

Thorinwasher:
Perfect. Just commit that, and I will merge this! :+1:

Thorinwasher (approved):
Perfect, looks good to me

PR #1510, opened 2025-12-28
Fix tag pipeline

4everTheOne:
Fixing TAG pipeline broken by: https://github.com/MockBukkit/MockBukkit/pull/1497 (Sorry!)

The text is [multiline](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands?utm_source=copilot.com#example-of-a-multiline-string)


Thorinwasher (approved):
Let's see what this does.

PR #1511, opened 2025-12-28
Upgrade v1.21.11 (Help wanted)

4everTheOne:
This is a PR open for contributions, the ideas is to work together on v1.21.11.
Feel free to open a PR for this branch.

PR #1517, opened 2025-12-30
Add 1.21.11 components

PuppyTransGirl:
# Description
This PR is related with #1511 and adds support for the new 1.21.11 components.

# Checklist

### DIFF
--- a/src/main/java/io/papermc/paper/datacomponent/item/AttackRangeMock.java
+++ b/src/main/java/io/papermc/paper/datacomponent/item/AttackRangeMock.java
@@ -1,0 +1,127 @@
+package io.papermc.paper.datacomponent.item;
+
+import com.google.common.base.Preconditions;
+import org.jetbrains.annotations.Range;
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+@SuppressWarnings({ "NonExtendableApiUsage", "UnstableApiUsage" })
+public class AttackRangeMock implements AttackRange
+{
+	private final float minReach;
+	private final float maxReach;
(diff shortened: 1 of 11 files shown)


PuppyTransGirl:
I think the KineticWeapon.Condition could be better but I don't really know a good way of implementing it

4everTheOne:
Thank you for your work @PuppyTransGirl! I have a suggestion for the KineticWeapon.Condition, maybe we could create a nested class something this:
```java
public class KineticWeaponMock implements KineticWeapon
{
// ...
    public static class ConditionMock implements Condition
    {
    // ...
    }
// ...
}
```
What do you think?

PuppyTransGirl:
What do you think would be better between a simple Constructor or a Builder inside the ConditionMock class ?

4everTheOne:
> What do you think would be better between a simple Constructor or a Builder inside the ConditionMock class ?

I think a simple constructor would be enough. if later we need to expand it to a Builder, it can be easily done :)

PuppyTransGirl:
I made it a Record class and I have put the NonNegative verification in the `ItemComponentTypesBridgeMock#kineticWeaponCondition`
Though, it doesn't have a verification if a dev uses the constructor of KineticWeapon.ConditionMock directly

4everTheOne:
It looks good to me! I just have an additional proposal, to include the unit tests :)
https://github.com/PuppyTransGirl/MockBukkit/pull/1
Other than that looks good to me

PuppyTransGirl:
Merged unit tests :3

4everTheOne (approved):

PR #1519, opened 2025-12-30
Make pipelines run on every pull request

4everTheOne:
# Description

If a PR is open to another branch, we should also run the basic tests.

# Checklist


svaningelgem (approved):
Shouldn't you also change the "push" then? Because that's also limited to `v1.*` right now?

Thorinwasher:
> Shouldn't you also change the "push" then? Because that's also limited to `v1.*` right now?

That one is meant to be used as a reference for coverage as far as I know, which should always be the main branch. Did not set up this though, I think it was @thelooter

4everTheOne:
> Shouldn't you also change the "push" then? Because that's also limited to `v1.*` right now?

This one i'm not sure what's the impact. This proposal was to allow us to work together on branches/PR's similar to what was done for 1.21.11.

So if anyone opens a PR to another PR it would be nice for us to have at least the build result.

---

Regarding your question @Thorinwasher , i'm no sure too but we could also remove the change in coverage.yaml and only have the build a lint result. (But i think that codecov can differentiate pull requests from main branch)

Thorinwasher (dismissed):

thelooter (approved):

PR #1520, opened 2025-12-31
feat: implement LivingEntityMock:isHandRaised with state management

svaningelgem:
- Added handRaised boolean field to track if a hand is raised
- Added raisedHand field to track which hand is raised (HAND or OFF_HAND)
- Implemented isHandRaised() to return the hand raised state
- Implemented getHandRaised() to return which hand is raised
- Added setHandRaised(boolean, EquipmentSlot) to allow setting state
- getHandRaised() throws IllegalStateException if no hand is raised
- Added comprehensive tests for all hand raised behaviors

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
@@ -104,6 +104,8 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 	private boolean gliding = false;
 	private boolean jumping = false;
 	private boolean riptiding = false;
+	private boolean handRaised = false;
+	private @Nullable EquipmentSlot raisedHand = null;
 
 	/**
 	 * The attributes this entity has.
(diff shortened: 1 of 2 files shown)


src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
```diff
@@ -1242,15 +1244,29 @@ public int getHandRaisedTime()
...
+	 * Sets whether a hand is raised and which hand is raised.
+	 *
+	 * @param handRaised Whether a hand is raised.
+	 * @param hand The hand that is raised, or null if no hand is raised.
+	 */
+	public void setHandRaised(boolean handRaised, @Nullable EquipmentSlot hand)
+	{
+		this.handRaised = handRaised;
+		this.raisedHand = hand;
 	}
```
4everTheOne: Suggestion: Instead of using 2 arguments here, maybe we could just have:
```java
public void setHandRaised(@Nullable EquipmentSlot hand)
{
// ...
}
```
This would represent:
- Null: No hand raised
- [HAND](https://jd.papermc.io/paper/1.21.11/org/bukkit/inventory/EquipmentSlot.html#HAND): Main hand raised
- [OFF_HAND](https://jd.papermc.io/paper/1.21.11/org/bukkit/inventory/EquipmentSlot.html#OFF_HAND): Other hand raised

Note: Would also be good to add a precondition validating the [EquipmentSlot](https://jd.papermc.io/paper/1.21.11/org/bukkit/inventory/EquipmentSlot.html) is one of null or HAND or OFF_HAND. To avoid people using [SADDLE ](https://jd.papermc.io/paper/1.21.11/org/bukkit/inventory/EquipmentSlot.html#SADDLE) for example.

With this, we can replace the `isHandRaised` with:
```java
@Override
public boolean isHandRaised()
{
    return this.raisedHand != null;
}
```

What do you think?

4everTheOne (commented):
Thank you for your contribution @svaningelgem! I have left a suggestion, let me know your opinion ;)

src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
svaningelgem: Hey @4everTheOne , I removed the isHandRaised etc methods now in favor of the getActiveItem-methods because I saw internally inside paper it's redirecting them to those anyhow.
Could you check if it makes sense?

Thanks

svaningelgem: I kept the isRaised-tests in place in order to check these methods/entries as well. But they're not needed inside the code because they're handled by paper already.

4everTheOne (approved):
Sorry for the delay @svaningelgem , and thank you for this contribution!

PR #1521, opened 2026-01-04
Scheduler miscellaneous improvements

frengor:
# Description

While implementing #1509 I noticed a couple of improvements that could be applied to the scheduler code. I would suggest to review this commit by commit. I decided to make only one PR, but I can split it up if you prefer so.

Major changes introduced in this PR:
* Tasks' cancel listeners are not run again if the task is already cancelled. This ensures the `cancel` method can be called multiple times
* Implements `getPendingTasks()`
* `isCurrentlyRunning` was implemented incorrectly. I checked the new implementation behaves like the latest Spigot/Paper 1.21.11 build and added the test I used to the unit tests (`taskRunningQueuedPending`).

# Checklist


4everTheOne (approved):

Issue #1506, opened 2025-12-16
Folia scheduler, is* methods


devdinc:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the feature

Currently methods like #getGlobalScheduler , #regionScheduler etc is not implemented. Would be lovely to have an implementation even if it is single threaded and every scheduler is the same.

PR #1523, opened 2026-01-08
feat: add single threaded implementations of RegionScheduler, GlobalR…

devdinc:
# Description

Solves issue #1506 

- implements single threaded RegionScheduler, EntityScheduler, GlobalRegionScheduler
- implements methods isOwnedByCurrentRegion, isGlobalTickThread in ServerMock
- implements method getScheduler in EntityMock
- implements tests regarding schedulers.


4everTheOne:
Hello @devdinc , thank you for your contribution.

This implementation will make the folia scheduler delegate the run to `BukkitScheduler`, i think that's a decision needs to be done as a team. Should we keep it as it is or separate it? @MockBukkit/maintainers what are your opinions?

devdinc:
> Hello @devdinc , thank you for your contribution.
> 
> This implementation will make the folia scheduler delegate the run to `BukkitScheduler`, i think that's a decision needs to be done as a team. Should we keep it as it is or separate it? @MockBukkit/maintainers what are your opinions?

Hello @4everTheOne. I agree that there is a decision to be made by @MockBukkit(@MockBukkit/project-members) team. I want to offer my views on the subject while informing you of my decisions:

- The decision to delegate to `BukkitScheduler` was intentional and follows the existing design. As you can see, `FoliaAsyncScheduler` already does the same.  I believe keeping this approach is way to go, while documenting the behaviour.

   That said, given Folia’s multithreaded nature, this approach may indeed be suboptimal, as you suggested it out.

- If we wanted to implement this differently, I believe this would cause major issues.
Most `ensureTickThread` checks are concentrated around `Level` logic and tick methods in entities and tile entities in Folia. Introducing a scheduler-level system that throws runtime exceptions for incorrect threading would be extremely problematic. It is unclear how such a mechanism could be implemented while still adhering to the current APIs and method signatures.

  In my view, clearly documenting the single-threaded assumptions and the lack of runtime enforcement is sufficient. Adding aggressive runtime checks at the scheduler level would likely cause more harm than benefit.

Thorinwasher:
> Hello @devdinc , thank you for your contribution.
> 
> This implementation will make the folia scheduler delegate the run to `BukkitScheduler`, i think that's a decision needs to be done as a team. Should we keep it as it is or separate it? @MockBukkit/maintainers what are your opinions?

We are mocking a paper server, which means that we should imitate the behavior of PaperMC. I think paper just delegates that to the "main  thread", as it doesn't have any concepts such as regions.

I will note that introducing this scheduler could technically make mockbukkit FoliaMC compatible. That would require a lot more work though, as many checks would need to be done to make sure that everything is managed from the right region. One thing to note is that a FoliaMC compatibility does not necessarily mean we remove the concept of a main thread, we could just add some context somewhere about what something is allowed to access at specific moments of time. Probably switch this depending on an environment variable if we ever went to do it.

src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
```diff
@@ -2768,49 +2775,54 @@ public int getSpawnLimit(@NotNull SpawnCategory spawnCategory)
...
+	public @NotNull EntityScheduler getEntityScheduler()
+	{
+		return this.foliaEntityScheduler;
 	}
 
 	@Override
 	public boolean isOwnedByCurrentRegion(@NotNull World world, @NotNull Position position)
 	{
-		throw new UnimplementedOperationException();
+		return true;
```
devdinc: Instead of unconditionally returning `true`, it would be preferable to rely on overloaded methods in conjunction with `Bukkit.isPrimaryThread()`. This provides a more accurate and intention-revealing threading check while maintaining compatibility with the existing single-threaded execution model.

Additionally, given the ongoing discussion around properly supporting the FoliaMC fork while still allowing single-threaded usage, we should consider a region-based abstraction rather than a global or environment-variable-driven approach. One viable option is to maintain a `List<Region>`, where a `Region` represents a bounded two-dimensional (or sectional) area derived from `position + @Nullable world`. Access checks would then be performed against these registered regions.

Rather than rebuilding regions on every tick or reacting implicitly to state changes, the responsibility for defining and updating regions would be delegated to the library consumer via an explicit API call (e.g., a `defineRegions(...)` method, along with `buildRegions(@Nullable world, threadAmount)`). This approach has several advantages:

* Eliminates per-tick region recomputation overhead.
* Makes threading and region ownership explicit rather than implicit.
* Avoids reliance on environment variables for behavioral control.
* Easier control over configuration of FoliaMC.
* If the consumer does not want FoliaMC, they simply do not call the methods in tearUp.

PaperMC support with this will be just, by default `defineRegions(buildRegions(null,1))`

Inside each tick method then we can simply call `isOwnedByCurrentRegion`, and if it isn't we explicitly record the exception, possibly consumer would have to call `getLastThreadingException` to check if a threading error occurred.

devdinc:
> > Hello @devdinc , thank you for your contribution.
> > This implementation will make the folia scheduler delegate the run to `BukkitScheduler`, i think that's a decision needs to be done as a team. Should we keep it as it is or separate it? @MockBukkit/maintainers what are your opinions?
> 
> We are mocking a paper server, which means that we should imitate the behavior of PaperMC. I think paper just delegates that to the "main thread", as it doesn't have any concepts such as regions.
> 
> I will note that introducing this scheduler could technically make mockbukkit FoliaMC compatible. That would require a lot more work though, as many checks would need to be done to make sure that everything is managed from the right region. One thing to note is that a FoliaMC compatibility does not necessarily mean we remove the concept of a main thread, we could just add some context somewhere about what something is allowed to access at specific moments of time. Probably switch this depending on an environment variable if we ever went to do it.

@Thorinwasher your suggested implementation of FoliaMC support sounds reasonable, but I might have a different approach you may find interesting.

[#r2691274095](https://github.com/MockBukkit/MockBukkit/pull/1523/changes#r2691274095)

src/test/java/org/mockbukkit/mockbukkit/scheduler/paper/FoliaEntitySchedulerTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @MockBukkitInject
+    private ServerMock server;
+
+    @MockBukkitInject
+    private BukkitSchedulerMock bukkitScheduler;
+
+    private FoliaEntityScheduler scheduler;
+    private PluginMock plugin;
+    // private EntityMock entity; // for testing retire maybe?
```
devdinc: Yeah, should test retire too.

src/main/java/org/mockbukkit/mockbukkit/scheduler/paper/FoliaEntityScheduler.java
```diff
@@ -0,0 +1,81 @@
...
+            @NotNull Plugin plugin,
+            @NotNull Runnable run,
+            @Nullable Runnable retired,
+            long delayTicks) {
+        Preconditions.checkNotNull(plugin, PLUGIN_CANNOT_BE_NULL);
+        Preconditions.checkNotNull(run, RUNNABLE_CANNOT_BE_NULL);
+
+        scheduler.runTaskLater(plugin, run, delayTicks);
+
+        // Entity is never retired in MockBukkit?
```
devdinc: Actually, entities can retire, by using `entity.remove()`. We can use `entity.isValid()` to check if the entity is removed.
Retire runnable should be used.

src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
Thorinwasher: > Rather than rebuilding regions on every tick or reacting implicitly to state changes, the responsibility for defining and updating regions would be delegated to the library consumer via an explicit API call (e.g., a defineRegions(...) method, along with buildRegions(@Nullable world, threadAmount)). This approach has several advantages:

This might actually help with FoliaMC debugging, as the main issue you always encounter in folia is that. (Also sorry for long inactivity)

Thorinwasher:
@devdinc I think this looks good enough to merge (sorry again for long inactivity). I don't think you need to include folia support in this PR, that would be too extensive, but also interesting to do.

devdinc:
> @devdinc I think this looks good enough to merge (sorry again for long inactivity). I don't think you need to include folia support in this PR, that would be too extensive, but also interesting to do.

No worries about the inactivity.

I agree that adding Folia support in this PR would significantly increase the scope. Proper implementation — especially testing — would require considerable time, so it makes sense to exclude it from this PR.

I’ve already shared the approach I had in mind for Folia support, and I believe it would be better handled in a separate, dedicated PR. At the moment, I don’t have the capacity to work on it further, but if a maintainer or contributor would like to take it on, that approach should serve as a solid starting point.

I’ll mark the PR as ready for review today or tomorrow once I’ve addressed the remaining feedback.

Thorinwasher (approved):

PR #1525, opened 2026-01-10
feat: implement WorldMock.spawnArrow methods with basic arrow flight simulation

svaningelgem:
## Summary
  Implements the two previously unimplemented `spawnArrow` methods in `WorldMock` and adds basic arrow flight simulation via a `tick()` method in `AbstractArrowMock`.

  ## Changes
  - ✅ Implement `spawnArrow(Location, Vector, float, float)` as shorthand for `Arrow.class`
  - ✅ Implement `spawnArrow(Location, Vector, float, float, Class<T>)` with full functionality
  - ✅ Add precondition validation: speed must be between 0 and 1 (0 = not flying, 1 = fully drawn)
  - ✅ Add precondition validation: spread must be non-negative
  - ✅ Add `tick()` method to `AbstractArrowMock` to simulate arrow movement
  - ✅ Support for spawning different arrow types (Arrow, SpectralArrow, etc.)
  - ✅ Random spread application when specified
  - ✅ Respect `noPhysics` flag (arrows don't move when set)

  ## Test Coverage
  Added 15 comprehensive tests covering:
  - Basic spawning and velocity calculations
  - Spread randomization
  - Different arrow types (SpectralArrow)
  - Precondition validation (null checks, speed bounds, negative spread)
  - Arrow flight simulation via tick()
  - NoPhysics flag behavior

  ## Limitations
  ⚠️ **Note:** The arrow flight simulation is intentionally basic and far from perfect. It simply moves the arrow by its velocity vector each tick without:
  - Gravity simulation
  - Air resistance/drag
  - Collision detection with blocks or entities
  - Trajectory arcing
  - Speed degradation over time

  This is sufficient for most testing scenarios where exact physics accuracy isn't required. Future PRs could enhance the flight simulation if needed.

  ## Testing
  All tests passing ✅

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
@@ -164,6 +164,7 @@ public class WorldMock implements World
 	private final BiomeProviderMock biomeProviderMock = new BiomeProviderMock();
 	private final @NotNull Map<Coordinate, Biome> biomes = new HashMap<>();
 	private @NotNull Difficulty difficulty = Difficulty.NORMAL;
+	private final Random random = new Random(0);
 	private boolean bonusChest = false;
 
 	private boolean allowAnimals = true;
@@ -2332,28 +2333,25 @@ public class WorldMock implements World
 		Preconditions.checkNotNull(location, "Location cannot be null");
 		Preconditions.checkNotNull(direction, "Direction cannot be null");
 		Preconditions.checkNotNull(clazz, "Class cannot be null");
-		Preconditions.checkArgument(speed >= 0 && speed <= 1, "Speed must be between 0 and 1 (inclusive), got %s", speed);
 		Preconditions.checkArgument(spread >= 0, "Spread must be non-negative, got %s", spread);
 
 		// Spawn the arrow entity
 		T arrow = spawn(location, clazz, CreatureSpawnEvent.SpawnReason.CUSTOM);
 
-		// Calculate velocity with spread
+		// Calculate velocity: normalize direction, apply spread, then scale by speed
 		Vector velocity = direction.clone().normalize();
 
-		// Apply random spread if specified
+		// Apply spread using triangular distribution matching Minecraft source
 		if (spread > 0)
 		{
-			Random random = ThreadLocalRandom.current();
 			velocity.add(new Vector(
-				(random.nextDouble() - 0.5) * spread,
-				(random.nextDouble() - 0.5) * spread,
-				(random.nextDouble() - 0.5) * spread
+				0.0172275 * spread * (random.nextDouble() - random.nextDouble()),
+				0.0172275 * spread * (random.nextDouble() - random.nextDouble()),
+				0.0172275 * spread * (random.nextDouble() - random.nextDouble())
 			));
 		}
 
-		// Normalize and apply speed
-		velocity.normalize().multiply(speed);
+		velocity.multiply(speed);
 		arrow.setVelocity(velocity);
 
 		// Mark the arrow as having been shot
--- a/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
@@ -847,8 +847,7 @@ public class WorldMock implements World
 	@Override
 	public @NotNull Arrow spawnArrow(Location location, Vector direction, float speed, float spread)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return spawnArrow(location, direction, speed, spread, Arrow.class);
 	}
 
 	@Override
@@ -2330,8 +2329,37 @@ public class WorldMock implements World
 	public <T extends AbstractArrow> @NotNull T spawnArrow(Location location, Vector direction, float speed, float spread,
 														   Class<T> clazz)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Preconditions.checkNotNull(location, "Location cannot be null");
+		Preconditions.checkNotNull(direction, "Direction cannot be null");
+		Preconditions.checkNotNull(clazz, "Class cannot be null");
+		Preconditions.checkArgument(speed >= 0 && speed <= 1, "Speed must be between 0 and 1 (inclusive), got %s", speed);
+		Preconditions.checkArgument(spread >= 0, "Spread must be non-negative, got %s", spread);
+
+		// Spawn the arrow entity
+		T arrow = spawn(location, clazz, CreatureSpawnEvent.SpawnReason.CUSTOM);
+
+		// Calculate velocity with spread
+		Vector velocity = direction.clone().normalize();
+
+		// Apply random spread if specified
+		if (spread > 0)
+		{
+			Random random = ThreadLocalRandom.current();
+			velocity.add(new Vector(
+				(random.nextDouble() - 0.5) * spread,
+				(random.nextDouble() - 0.5) * spread,
+				(random.nextDouble() - 0.5) * spread
+			));
+		}
+
+		// Normalize and apply speed
+		velocity.normalize().multiply(speed);
+		arrow.setVelocity(velocity);
+
+		// Mark the arrow as having been shot
+		arrow.setHasBeenShot(true);
+
+		return arrow;
 	}
 
 	@Override
--- a/src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
+++ b/src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
@@ -2900,4 +2900,209 @@ class WorldMockTest
 
 	}
 
+	@Nested
+	class SpawnArrowTests
+	{
+
+		@Test
+		void spawnArrow_BasicSpawn_CreatesArrow()
+		{
+			WorldMock world = new WorldMock();
+			Location location = new Location(world, 10, 20, 30);
+			org.bukkit.util.Vector direction = new org.bukkit.util.Vector(1, 0, 0);
+			Arrow arrow = world.spawnArrow(location, direction, 0.6f, 0);
+
+			assertNotNull(arrow);
+			assertEquals(location, arrow.getLocation());
+			assertTrue(arrow.hasBeenShot());
+		}
+
+		@Test
+		void spawnArrow_WithSpeed_SetsCorrectVelocity()
+		{
+			WorldMock world = new WorldMock();
+			Location location = new Location(world, 0, 0, 0);
+			org.bukkit.util.Vector direction = new org.bukkit.util.Vector(1, 0, 0);
+			float speed = 0.8f;
+
+			Arrow arrow = world.spawnArrow(location, direction, speed, 0);
+
+			org.bukkit.util.Vector velocity = arrow.getVelocity();
+			// Velocity should be normalized direction * speed
+			assertEquals(speed, velocity.length(), 0.001);
+		}
+
+		@Test
+		void spawnArrow_WithSpread_AppliesRandomization()
+		{
+			WorldMock world = new WorldMock();
+			Location location = new Location(world, 0, 0, 0);
+			org.bukkit.util.Vector direction = new org.bukkit.util.Vector(1, 0, 0);
+
+			Arrow arrow1 = world.spawnArrow(location, direction, 0.5f, 0.1f);
+			Arrow arrow2 = world.spawnArrow(location, direction, 0.5f, 0.1f);
+
+			// With spread, arrows should have different velocities
+			// (this might occasionally fail due to randomness, but is very unlikely)
+			assertNotEquals(arrow1.getVelocity(), arrow2.getVelocity());
+		}
+
+		@Test
+		void spawnArrow_WithSpectralArrowClass_CreatesSpectralArrow()
+		{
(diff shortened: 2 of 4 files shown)


src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -2309,8 +2308,37 @@ public void removePluginChunkTickets(Plugin plugin)
...
+		Vector velocity = direction.clone().normalize();
+
+		// Apply random spread if specified
+		if (spread > 0)
+		{
+			Random random = ThreadLocalRandom.current();
+			velocity.add(new Vector(
+				(random.nextDouble() - 0.5) * spread,
+				(random.nextDouble() - 0.5) * spread,
+				(random.nextDouble() - 0.5) * spread
```
Thorinwasher: Randomness is usually not a good idea when unit testing

Thorinwasher (commented):
Looks good on most parts. I'm unsure if randomness is a good idea to use here, might be better to replace with some kind of pseoudorandom thing instead. I would think just storing a random in world mock with a seed would work well for that purpose.

One really important thing here is that the tests should be reproducible

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -2309,8 +2308,37 @@ public void removePluginChunkTickets(Plugin plugin)
 	public <T extends AbstractArrow> @NotNull T spawnArrow(Location location, Vector direction, float speed, float spread,
 														   Class<T> clazz)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		Preconditions.checkNotNull(location, "Location cannot be null");
+		Preconditions.checkNotNull(direction, "Direction cannot be null");
+		Preconditions.checkNotNull(clazz, "Class cannot be null");
+		Preconditions.checkArgument(speed >= 0 && speed <= 1, "Speed must be between 0 and 1 (inclusive), got %s", speed);
```
sdeschaetzen-ccv: I cannot find any documentation that seems to indicate this constraint exists

Thorinwasher: We usually look into the papermc source code for these situations. I haven't confirmed this, but I expect these constraints to be copied directly from the papermc source.

src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
```diff
@@ -2900,4 +2900,209 @@ void spawnParticle_testingTimings()
...
+	@Nested
+	class SpawnArrowTests
+	{
+
+		@Test
+		void spawnArrow_BasicSpawn_CreatesArrow()
+		{
+			WorldMock world = new WorldMock();
+			Location location = new Location(world, 10, 20, 30);
+			org.bukkit.util.Vector direction = new org.bukkit.util.Vector(1, 0, 0);
```
sdeschaetzen-ccv: Why did you use a fully-qualified type here?

src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
```diff
@@ -2900,4 +2900,209 @@ void spawnParticle_testingTimings()
...
+			WorldMock world = new WorldMock();
+			Location location = new Location(world, 0, 0, 0);
+			org.bukkit.util.Vector direction = new org.bukkit.util.Vector(1, 0, 0);
+			float speed = 0.5f;
+
+			Arrow arrow = world.spawnArrow(location, direction, speed, 0);
+			Location initialLocation = arrow.getLocation().clone();
+
+			// Tick the arrow (cast to mock to access tick method)
+			((org.mockbukkit.mockbukkit.entity.ArrowMock) arrow).tick();
```
sdeschaetzen-ccv: No need for a fully qualified type here I think

src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
```diff
@@ -2900,4 +2900,209 @@ void spawnParticle_testingTimings()
...
+				world.spawnArrow(location, direction, 0.5f, -1f));
+		}
+
+		@Test
+		void spawnArrow_ArrowTick_UpdatesPosition()
+		{
+			WorldMock world = new WorldMock();
+			Location location = new Location(world, 0, 0, 0);
+			org.bukkit.util.Vector direction = new org.bukkit.util.Vector(1, 0, 0);
+			float speed = 0.5f;
```
sdeschaetzen-ccv: You can just inline this like in the other tests

src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
Thorinwasher: Good point, should just be in an import

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
sdeschaetzen-ccv: Not seeing in the papermc source either: https://github.com/PaperMC/Paper/blob/main/paper-server/src/main/java/org/bukkit/craftbukkit/CraftWorld.java

Though it does call into decompiled minecraft, perhaps it does such a check?

Thorinwasher: Eh, let me just double check. Thanks for looking into these things, by the way :3

src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
```diff
@@ -2900,4 +2900,209 @@ void spawnParticle_testingTimings()
...
+			org.bukkit.util.Vector direction = new org.bukkit.util.Vector(1, 0, 0);
+
+			Arrow arrow = world.spawnArrow(location, direction, 0.5f, 0);
+			int initialLifetime = arrow.getLifetimeTicks();
+
+			// Tick the arrow (cast to mock to access tick method)
+			((org.mockbukkit.mockbukkit.entity.ArrowMock) arrow).tick();
+
+			assertEquals(initialLifetime + 1, arrow.getLifetimeTicks());
+		}
```
sdeschaetzen-ccv: Should these test possible be in AbstractArrowMockTest instead?

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -840,8 +840,7 @@ public boolean refreshChunk(int x, int z)
 	@Override
 	public @NotNull Arrow spawnArrow(Location location, Vector direction, float speed, float spread)
```
sdeschaetzen-ccv, line 842: Perhaps also change this to ArrowMock?

src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
Thorinwasher: I don't think it matters too much. Might be a bit better though, as abstract arrow mock test is going to be a smaller class.

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
Thorinwasher: Minecraft source code:
```java
new Vec3(x, y, z)
            .normalize()
            .add(
                this.random.triangle(0.0, 0.0172275 * inaccuracy),
                this.random.triangle(0.0, 0.0172275 * inaccuracy),
                this.random.triangle(0.0, 0.0172275 * inaccuracy)
            )
            .scale(velocity);
```
I think that means you just have a random value between -0.0172275 * inaccuracy and 0.0172275 * inaccuracy appended to the velocity

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -2309,8 +2308,37 @@ public void removePluginChunkTickets(Plugin plugin)
...
+			Random random = ThreadLocalRandom.current();
+			velocity.add(new Vector(
+				(random.nextDouble() - 0.5) * spread,
+				(random.nextDouble() - 0.5) * spread,
+				(random.nextDouble() - 0.5) * spread
+			));
+		}
+
+		// Normalize and apply speed
+		velocity.normalize().multiply(speed);
```
Thorinwasher: It does not look like the velocity is normalized again in minecraft source

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
Thorinwasher: When looking into minecraft source, I could not find any check that does this, so you're correct

Thorinwasher (changes requested):
When looking into this more thorougly, thanks to the help of @sdeschaetzen-ccv , I have some more changes that needs to be applied, and also that which @sdeschaetzen-ccv mentioned

I might actually get back in track with these reviews again.

svaningelgem:
Thanks for the thorough review! Here's what I've addressed:

**Speed constraint removed** — Per Thorinwasher and sdeschaetzen-ccv's investigation, the Minecraft source has no speed validation. Removed the 0–1 constraint entirely (negative and >1 speeds are technically valid, same as Minecraft).

**Spread formula fixed** — Updated to match Minecraft source exactly:
```java
velocity.add(new Vector(
    0.0172275 * spread * (random.nextDouble() - random.nextDouble()),
    0.0172275 * spread * (random.nextDouble() - random.nextDouble()),
    0.0172275 * spread * (random.nextDouble() - random.nextDouble())
));
```
This uses the triangular distribution (`triangle(0.0, 0.0172275 * inaccuracy)`) from the Minecraft source.

**Re-normalization removed** — The velocity is not normalized after applying spread (matching Minecraft source: normalize → add spread → scale by speed, no re-normalize).

**Seeded Random** — Added `private final Random random = new Random(0);` to `WorldMock`, replacing `ThreadLocalRandom.current()`. Tests are now fully deterministic.

**Tick tests moved** — Moved the three tick tests from `WorldMockTest` to `AbstractArrowMockTest`, plus added a `tick_NotShot_DoesNotMove` test for full branch coverage (the `hasBeenShot=false` path).

**Imports fixed** — Removed all fully-qualified `org.bukkit.util.Vector` references; added proper import.

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
svaningelgem: Fixed — replaced  with  as a field on , so spread is fully deterministic.

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
svaningelgem: Fixed — constraint removed entirely. Per the Minecraft source (thanks to both of you for looking into it), there is no speed validation.

src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
svaningelgem: Fixed — added a proper import for org.bukkit.util.Vector and removed all fully-qualified references.

src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
svaningelgem: Fixed — imported ArrowMock properly.

src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
svaningelgem: Done — the tick tests were moved to AbstractArrowMockTest instead, so the inline was no longer relevant there.

src/test/java/org/mockbukkit/mockbukkit/world/WorldMockTest.java
svaningelgem: Done — moved all three tick tests to AbstractArrowMockTest. Also added a tick_NotShot_DoesNotMove test there for full branch coverage.

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
svaningelgem: Left as Arrow.class — the spawn() method handles the mapping to ArrowMock internally via EntityTypesMock, so Arrow.class is correct here (it's the Bukkit API type the method is declared to return).

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
svaningelgem: Fixed — updated to match Minecraft source exactly: 0.0172275 * spread * (nextDouble() - nextDouble()) per axis, which implements the triangular distribution.

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
svaningelgem: Fixed — removed the re-normalization. The velocity is now: normalize direction, add spread offset, multiply by speed — matching the Minecraft source.

Thorinwasher:
Will review when I get the time

Thorinwasher (approved):
LGTM, gradle plugin portal is down though. We wait a bit

PR #1544, opened 2026-03-12
feat: implement PlayerCommandPreprocessEvent

maxikg:
# Description

This PR implements the call of the [`PlayerCommandPreprocessEvent`](https://jd.papermc.io/paper/1.21.11/org/bukkit/event/player/PlayerCommandPreprocessEvent.html), when a player performs a command. This event is triggered when a player performs a command, and before execution is handled by a `CommandExecutor`. It enables command executions to be cancelled or altered. In the case of MockBukkit, commands are performed via either  `Server#dispatchCommand(CommandSender, String)` or `Player#performCommand(String)`.

# Checklist


4everTheOne (approved):

PR #1574, opened 2026-06-18
Alternate unimplemented operation handling

### LINKED ISSUE #1507: Tests being ignored due to wrong paper version

### Describe the feature

Right now my paper version is wrong. The result is that all my tests are ignored. Not wrong. Ignored.

This leads my CI to pass. Which gives the false sense of security.

My suggestion would be to fail when the paper version is wrong. Not auto-ignore them.

Thorinwasher:
# Description

Fixes #1507. Checks for system property `mockbukkit.unimplementedoperation.fail` = `true`, and uses a separate exception in that case. 

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/exception/UnimplementedOperationException.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/exception/UnimplementedOperationException.java
@@ -25,7 +25,7 @@ public class UnimplementedOperationException extends TestAbortedException
 	 * Constructs a new  with a default message.
 	 */
 	@ApiStatus.Internal
-	public UnimplementedOperationException()
+	private UnimplementedOperationException()
 	{
 		this("Not implemented");
 	}
@@ -36,9 +36,37 @@ public class UnimplementedOperationException extends TestAbortedException
 	 * @param message The message.
 	 */
 	@ApiStatus.Internal
-	public UnimplementedOperationException(@NotNull String message)
+	private UnimplementedOperationException(@NotNull String message)
 	{
 		super(message);
+	}
+
+	/**
+	 *
+	 * @return Either a {@link UnimplementedOperationException} or {@link UnimplementedOperationFailure}
+	 */
+	@ApiStatus.Internal
+	public static RuntimeException exception()
+	{
+		return exception("Not implemented");
+	}
+
+	/**
+	 *
+	 * @param message The error message
+	 * @return Either a {@link UnimplementedOperationException} or {@link UnimplementedOperationFailure}
+	 */
+	@ApiStatus.Internal
+	public static RuntimeException exception(String message)
+	{
+		if ("true".equalsIgnoreCase(System.getenv("mockbukkit.unimplementedoperation.fail")))
+		{
+			return new UnimplementedOperationFailure(message);
+		}
+		else
+		{
+			return new UnimplementedOperationException(message);
+		}
 	}
 
 }
--- a/src/main/java/org/mockbukkit/mockbukkit/exception/UnimplementedOperationException.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/exception/UnimplementedOperationException.java
@@ -65,7 +65,7 @@ public class UnimplementedOperationException extends TestAbortedException
 		}
 		else
 		{
-			return new UnimplementedOperationException(message);
+			return UnimplementedOperationException.exception(message);
 		}
 	}
 
--- a/src/main/java/org/mockbukkit/mockbukkit/exception/UnimplementedOperationException.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/exception/UnimplementedOperationException.java
@@ -59,7 +59,7 @@ public class UnimplementedOperationException extends TestAbortedException
 	@ApiStatus.Internal
 	public static RuntimeException exception(String message)
 	{
-		if ("true".equalsIgnoreCase(System.getenv("mockbukkit.unimplementedoperation.fail")))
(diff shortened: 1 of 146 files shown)


svaningelgem:
Found it! The bulk is just `new X()` -> `X.exception()`, the factory branch is the actual change.

One thing on that branch: it reads `System.getenv("mockbukkit.unimplementedoperation.fail")`, but the description says system property. Different mechanisms, and a dotted env var name is awkward to set in most shells. I think you meant `System.getProperty(...)`, which also matches the wording.

src/main/java/org/mockbukkit/mockbukkit/exception/UnimplementedOperationException.java
```diff
@@ -36,9 +36,37 @@ public UnimplementedOperationException()
...
+
+	/**
+	 *
+	 * @param message The error message
+	 * @return Either a {@link UnimplementedOperationException} or {@link UnimplementedOperationFailure}
+	 */
+	@ApiStatus.Internal
+	public static RuntimeException exception(String message)
+	{
+		if ("true".equalsIgnoreCase(System.getenv("mockbukkit.unimplementedoperation.fail")))
```
walker84837: This should check for the *system property*, not for the environment variable.

It can be simplified with `Boolean.getBoolean("mockbukkit.unimplementedoperation.fail")`, which essentially does the same thing.

P.S. Didn't notice [svaningelgem](https://github.com/svaningelgem) also pointed it out.

walker84837:
We might also want to add a simple unit test, just to make sure the toggle works properly.

Thorinwasher:
> Found it! The bulk is just `new X()` -> `X.exception()`, the factory branch is the actual change.
> 
> One thing on that branch: it reads `System.getenv("mockbukkit.unimplementedoperation.fail")`, but the description says system property. Different mechanisms, and a dotted env var name is awkward to set in most shells. I think you meant `System.getProperty(...)`, which also matches the wording.

Whops, should be getproperty 😅

src/main/java/org/mockbukkit/mockbukkit/exception/UnimplementedOperationException.java
4everTheOne: I have an alternative suggestion, which one do you prefer?
```java
	@ApiStatus.Internal
	public static RuntimeException exception(String message)
	{
		if (Boolean.parseBoolean(System.getenv("MOCKBUKKIT_FAIL_ON_UNIMPLEMENTED")))
		{
			return new UnimplementedOperationFailure(message);
		}
		else
		{
			return new UnimplementedOperationException(message);
		}
	}
```

```yml
env:
  MOCKBUKKIT_FAIL_ON_UNIMPLEMENTED: true

steps:
  - run: ./gradlew test

```

Or:
```yml
env:
  mockbukkit.unimplementedoperation.fail: true

steps:
  - run: ./gradlew test

```

4everTheOne (commented):
First of all, great work! :) The code looks good to me, I just have an alternative sugestion for the variable name :)

4everTheOne:
Should we also update the README with the information?

src/main/java/org/mockbukkit/mockbukkit/exception/UnimplementedOperationException.java
Thorinwasher: Environment variables are generally capital case, yep. I was going for a system property, but messed up. Should we have both property support and environment variable support?

On a separate note, should it fail completely if the property is invalidly configured?

Thorinwasher:
> Should we also update the README with the information?

Yep, I think we even have a section specifically for this.

src/main/java/org/mockbukkit/mockbukkit/exception/UnimplementedOperationException.java
svaningelgem: I think we should go for a system property. That's more how Java/Gradle works than via envvars...

walker84837: I agree. From how it's cased and given the previous discussion in #1507 , it should be changed to be a system property.

4everTheOne:
Hmm weird, i may be doing something wrong.. But i was trying to test the feature on MockBukkit because we have some tests that are throwing the `UnimplementedOperationException`.

I tried to run the following command and it seems to pass anyway instead of throwing the failure.
```
./gradlew test -Dmockbukkit.unimplementedoperation.fail=true
``` 

Any ideia why this could be? At first look look like the system property is not being passed to the test itself only for the gradle jvm

svaningelgem:
The `-D` only lands on the Gradle daemon JVM, not the forked test JVM.

The `test {}` task doesn't pass it through.

Can you try to add this to build.gradle.kts so it gets forwarded:

```gradle
test {
    systemProperty("mockbukkit.unimplementedoperation.fail",
    System.getProperty("mockbukkit.unimplementedoperation.fail") ?: "")
}
```

Thorinwasher:
> The `-D` only lands on the Gradle daemon JVM, not the forked test JVM.
> 
> The `test {}` task doesn't pass it through.
> 
> Can you try to add this to build.gradle.kts so it gets forwarded:
> 
> ```groovy-gradle
> test {
>     systemProperty("mockbukkit.unimplementedoperation.fail",
>     System.getProperty("mockbukkit.unimplementedoperation.fail") ?: "")
> }
> ```

Yep, system properties can be a bit confusing to work with, as we're dealing with multiple isolated JVM's

4everTheOne (approved):

PR #1577, opened 2026-06-23
fix(meta): ItemMeta equals/hashCode ignores CustomModelDataComponent

kccricket:
# Description

## Problem

`ItemMetaMock` stored custom model data in two independent fields -- a legacy `Integer customModelData` and a separate `CustomModelDataComponent customModelDataComponent` -- but `equals()` and `hashCode()` only consulted the integer. Two `ItemMeta` instances that differed solely in their `CustomModelDataComponent` (e.g. different `strings`, `floats`, or `colors`) compared as equal and produced the same hash code, diverging from Paper's production behavior.

## Changes

**`ItemMetaMock`** -- consolidates both fields into a single `CustomModelDataComponent customModelData` field, matching `CraftMetaItem`'s design. The legacy `int` API (`hasCustomModelData` / `getCustomModelData` / `setCustomModelData`) becomes a thin view over the component's first float value. `equals()` and `hashCode()` now compare the full component, so all sub-fields participate.

**`CustomModelDataComponentMock`** -- adds a copy constructor `CustomModelDataComponentMock(CustomModelDataComponent source)` mirroring `CraftCustomModelDataComponent(CraftCustomModelDataComponent)`. This is used by `ItemMetaMock`'s copy constructor to produce independent clones, and is more future-proof than manually enumerating the four sub-fields at each call site.

## Tests added

- `equals_setCustomModelDataAndSetCustomModelDataComponentEquivalent_True` -- `setCustomModelData(42)` and `setCustomModelDataComponent(builder().floats([42f]).build())` produce equal metas.
- `equals_CustomModelDataComponentDifferent_False` -- two metas with differing component `strings` are not equal.
- `equals_CustomModelDataComponentOneWithout_False` -- asymmetric presence is not equal.
- `clone_CustomModelDataComponent_IsIndependent` -- mutating the original's component after cloning does not affect the clone.
- `CopyConstructor` suite on `CustomModelDataComponentMock` -- copies all four sub-fields and produces an independent instance.

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ItemMetaMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ItemMetaMock.java
@@ -122,7 +122,6 @@ public class ItemMetaMock implements ItemMeta, Damageable, Repairable
 	private @NotNull Set<ItemFlag> hideFlags = EnumSet.noneOf(ItemFlag.class);
 	private @NotNull PersistentDataContainerMock persistentDataContainer = new PersistentDataContainerMock();
 	private boolean unbreakable = false;
-	private @Nullable Integer customModelData = null;
 	private boolean hideTooltip;
 	private boolean fireResistant;
 	private boolean glider;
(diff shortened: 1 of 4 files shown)


4everTheOne (approved):

kccricket:
@4everTheOne Do you do maintenance releases on older versions? If so, I wouldn't mind backporting it to the 1.21 branch.

4everTheOne:
At the moment I don't think we have a process for that. In theory we just need to cherry pick this commit into the older branch. I'll do that tomorrow ;)

PR #1578, opened 2026-06-25
fix(meta): ItemMeta equals/hashCode ignores CustomModelDataComponent …

4everTheOne:
Backport https://github.com/MockBukkit/MockBukkit/pull/1577
---
# Description

## Problem

`ItemMetaMock` stored custom model data in two independent fields -- a legacy `Integer customModelData` and a separate `CustomModelDataComponent customModelDataComponent` -- but `equals()` and `hashCode()` only consulted the integer. Two `ItemMeta` instances that differed solely in their `CustomModelDataComponent` (e.g. different `strings`, `floats`, or `colors`) compared as equal and produced the same hash code, diverging from Paper's production behavior.

## Changes

**`ItemMetaMock`** -- consolidates both fields into a single `CustomModelDataComponent customModelData` field, matching `CraftMetaItem`'s design. The legacy `int` API (`hasCustomModelData` / `getCustomModelData` / `setCustomModelData`) becomes a thin view over the component's first float value. `equals()` and `hashCode()` now compare the full component, so all sub-fields participate.

**`CustomModelDataComponentMock`** -- adds a copy constructor `CustomModelDataComponentMock(CustomModelDataComponent source)` mirroring `CraftCustomModelDataComponent(CraftCustomModelDataComponent)`. This is used by `ItemMetaMock`'s copy constructor to produce independent clones, and is more future-proof than manually enumerating the four sub-fields at each call site.

## Tests added

- `equals_setCustomModelDataAndSetCustomModelDataComponentEquivalent_True` -- `setCustomModelData(42)` and `setCustomModelDataComponent(builder().floats([42f]).build())` produce equal metas.
- `equals_CustomModelDataComponentDifferent_False` -- two metas with differing component `strings` are not equal.
- `equals_CustomModelDataComponentOneWithout_False` -- asymmetric presence is not equal.
- `clone_CustomModelDataComponent_IsIndependent` -- mutating the original's component after cloning does not affect the clone.
- `CopyConstructor` suite on `CustomModelDataComponentMock` -- copies all four sub-fields and produces an independent instance.

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ItemMetaMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ItemMetaMock.java
@@ -121,7 +121,6 @@ public class ItemMetaMock implements ItemMeta, Damageable, Repairable
 	private @NotNull Set<ItemFlag> hideFlags = EnumSet.noneOf(ItemFlag.class);
 	private @NotNull PersistentDataContainerMock persistentDataContainer = new PersistentDataContainerMock();
 	private boolean unbreakable = false;
-	private @Nullable Integer customModelData = null;
 	private boolean hideTooltip;
 	private boolean fireResistant;
 	private boolean glider;
@@ -135,7 +134,7 @@ public class ItemMetaMock implements ItemMeta, Damageable, Repairable
 	private @Nullable NamespacedKey tooltipStyle;
 	private @Nullable NamespacedKey damageResistant;
 
-	private @Nullable CustomModelDataComponent customModelDataComponent;
+	private @Nullable CustomModelDataComponent customModelData;
 	private @Nullable UseCooldownComponent useCooldown;
 	private @Nullable FoodComponent foodComponent;
 	private @Nullable ToolComponent toolComponent;
@@ -193,7 +192,10 @@ public class ItemMetaMock implements ItemMeta, Damageable, Repairable
 			this.persistentDataContainer = new PersistentDataContainerMock(m.persistentDataContainer);
 		}
 		unbreakable = meta.isUnbreakable();
-		customModelData = meta.hasCustomModelData() ? meta.getCustomModelData() : null;
+		if (meta.hasCustomModelDataComponent())
+		{
+			this.customModelData = new CustomModelDataComponentMock(meta.getCustomModelDataComponent());
+		}
 		hideTooltip = meta.isHideTooltip();
 		fireResistant = meta.isFireResistant();
 		if (meta.hasMaxStackSize())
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockbukkit/mockbukkit/inventory/meta/components/CustomModelDataComponentMock.java
```diff
@@ -34,6 +35,11 @@ private CustomModelDataComponentMock(List<Float> floats, List<Boolean> flags, Li
 		this.colors = List.copyOf(colors);
 	}
 
+	public CustomModelDataComponentMock(@NotNull CustomModelDataComponent source)
```
Thorinwasher, line 38:
```suggestion
    @ApiStatus.Internal
	public CustomModelDataComponentMock(@NotNull CustomModelDataComponent source)
```

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/ItemMetaMock.java
```diff
@@ -135,7 +134,7 @@ public class ItemMetaMock implements ItemMeta, Damageable, Repairable
 	private @Nullable NamespacedKey tooltipStyle;
 	private @Nullable NamespacedKey damageResistant;
 
-	private @Nullable CustomModelDataComponent customModelDataComponent;
+	private @Nullable CustomModelDataComponent customModelData;
```
Thorinwasher, line 137: Doesn't matter
```suggestion
	private @Nullable CustomModelDataComponent customModelDataComponent;
```

src/test/java/org/mockbukkit/mockbukkit/inventory/meta/ItemMetaMockTest.java
```diff
@@ -621,6 +621,52 @@ void equals_CustomModelDataOneWithout_False()
 		assertNotEquals(meta.hashCode(), meta2.hashCode());
 	}
 
+	@Test
+	@SuppressWarnings("UnstableApiUsage")
+	void equals_setCustomModelDataAndSetCustomModelDataComponentEquivalent_True()
+	{
+		ItemMetaMock meta2 = new ItemMetaMock();
+		meta.setCustomModelData(42);
```
Thorinwasher, line 629: The meaning of life =

src/test/java/org/mockbukkit/mockbukkit/inventory/meta/ItemMetaMockTest.java
```diff
@@ -621,6 +621,52 @@ void equals_CustomModelDataOneWithout_False()
...
+	void equals_CustomModelDataComponentOneWithout_False()
+	{
+		ItemMetaMock meta2 = new ItemMetaMock();
+		meta.setCustomModelDataComponent(CustomModelDataComponentMock.builder().strings(List.of("plugin:variant_a")).build());
+		assertNotEquals(meta, meta2);
+		assertNotEquals(meta2, meta);
+	}
+
+	@Test
+	@SuppressWarnings("UnstableApiUsage")
```
Thorinwasher, line 659: We should be suppressing these warnings globally in the project.

Thorinwasher (changes requested):

PR #1581, opened 2026-06-28
Upgrade paper from build 57 to 72

4everTheOne:
# Description

Upgrade paper mc from build 57 to 72

# Checklist


Thorinwasher (approved):
We should setup some automatic bot to bump the version, I think we're using dependabot?

Now that I think about it, some things can't be done automatically. This update includes other elements as well, like paper API additions. I'm thinking this:
- A bot starts a PR with the updated release
- If build fails, then API changed a bit. We should then be able to manually fix build issues

Issue #1582, opened 2026-06-29
EntityMock#getName has wrong semantics compared with actual Paper behavior


SkytAsul:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Are you using the latest version of MockBukkit?

- [x] I am using the latest version of MockBukkit.

### Minecraft Version

1.21

### Describe the bug

EntityMock#getName returns the name of the entity even if a custom name has been defined:
https://github.com/MockBukkit/MockBukkit/blob/ec5a2dcc4bf30a3933bce705f74046d05af667fb/src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java#L474-L477

However, in Paper, Entity#getName returns **the custom name if defined,** else the name:
https://github.com/PaperMC/Paper/blob/0ae1b4239e9559516917f86bfcac404a091c710c/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java#L779-L781 (Paper source)
https://mcsrc.dev/1/26.1.2/net/minecraft/world/entity/Entity#L2904-2907 (NMS entity source)

I believe this is a mistake on MockBukkit part. There is even a unit test testing this flawed behavior:
https://github.com/MockBukkit/MockBukkit/blob/ec5a2dcc4bf30a3933bce705f74046d05af667fb/src/test/java/org/mockbukkit/mockbukkit/entity/EntityMockTest.java#L315-L320
Here, `entity.getName()` should also return `"Some Custom Name"`.

### Reproducible Test

Unit test mentioned above.

### Anything else?

The companion Adventure method EntityMock#name also has this wrong behavior.

PR #1583, opened 2026-06-29
feat: add Adventure pointers to EntityMock and PlayerMock

SkytAsul:
# Description
This PR adds the following Adventure identity pointers to both EntityMock and PlayerMock: name, display name, uuid and locale. Those pointers have the same semantics as the ones actually implemented in Paper:
https://github.com/PaperMC/Paper/blob/0ae1b4239e9559516917f86bfcac404a091c710c/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java#L228-L233
https://github.com/PaperMC/Paper/blob/0ae1b4239e9559516917f86bfcac404a091c710c/paper-server/src/main/java/org/bukkit/craftbukkit/entity/CraftEntity.java#L91-L95

One test is disabled because of a difference in semantics between MockBukkit and Paper (reported at #1582) 

# Checklist

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
@@ -7,6 +7,7 @@ import io.papermc.paper.datacomponent.DataComponentType;
 import io.papermc.paper.entity.LookAnchor;
 import io.papermc.paper.entity.TeleportFlag;
 import io.papermc.paper.threadedregions.scheduler.EntityScheduler;
+import net.kyori.adventure.identity.Identity;
 import net.kyori.adventure.pointer.Pointers;
 import net.kyori.adventure.pointer.PointersSupplier;
 import net.kyori.adventure.text.Component;
@@ -93,8 +94,8 @@ public abstract class EntityMock extends Entity.Spigot implements Entity, Messag
 
 	private static final AtomicInteger ENTITY_COUNTER = new AtomicInteger();
 	static final PointersSupplier<EntityMock> POINTERS_SUPPLIER = PointersSupplier.<EntityMock>builder()
-			.resolving(net.kyori.adventure.identity.Identity.DISPLAY_NAME, EntityMock::name)
-			.resolving(net.kyori.adventure.identity.Identity.UUID, EntityMock::getUniqueId)
+			.resolving(Identity.DISPLAY_NAME, EntityMock::name)
+			.resolving(Identity.UUID, EntityMock::getUniqueId)
 			.build();
 
 	private final Set<String> tags = Sets.newHashSet();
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
```diff
@@ -90,6 +92,10 @@ public abstract class EntityMock extends Entity.Spigot implements Entity, Messag
 {
 
 	private static final AtomicInteger ENTITY_COUNTER = new AtomicInteger();
+	static final PointersSupplier<EntityMock> POINTERS_SUPPLIER = PointersSupplier.<EntityMock>builder()
```
Thorinwasher, line 96:
```suggestion
	private static final PointersSupplier<EntityMock> POINTERS_SUPPLIER = PointersSupplier.<EntityMock>builder()
```

Thorinwasher (changes requested):
Looks good, except a minor nitpick

Thorinwasher:
Actually when double checking; this might not fix your exact issue with the getName method. I suggest testing that method as well :)

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
SkytAsul: No, this field is accessed in PlayerMock so it has to be package-private.

SkytAsul:
> Actually when double checking; this might not fix your exact issue with the getName method.

Yeah the getName issue is not completely related to the pointers so I did not want to fix it in this PR.

> I suggest testing that method as well :)

I actually did! But I put `@Disabled` in front of the test since the "correct" (tested) behavior is actually incorrect.

Do you want me to create a separate PR to fix the getName thing or can I include the fix in this one?

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
```diff
@@ -90,6 +92,10 @@ public abstract class EntityMock extends Entity.Spigot implements Entity, Messag
...
+			.resolving(net.kyori.adventure.identity.Identity.DISPLAY_NAME, EntityMock::name)
+			.resolving(net.kyori.adventure.identity.Identity.UUID, EntityMock::getUniqueId)
```
4everTheOne: Can we replace the FQN with the import?
```java
			.resolving(Identity.DISPLAY_NAME, EntityMock::name)
			.resolving(Identity.UUID, EntityMock::getUniqueId)
```

4everTheOne (commented):
First of all, sorry for the delay in the review and a big thank you for your contribution! I just have a minor request, other than that looks good to me.

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
SkytAsul: Sure!

4everTheOne (approved):

PR #1585, opened 2026-06-30
ci: harden and clean up GitHub Actions workflows

thelooter:
## Description

Hardening and cleanup pass over the GitHub Actions workflows. No application code changes. Each concern is a separate commit so it can be reviewed (and reverted) independently. Grouped below by category.

### Correctness — release pipeline

- **`tag` job had no JDK/Gradle setup.** It ran `./gradlew printPaperDetails` relying on whatever JDK the runner image happened to ship, while every other workflow explicitly sets up JDK 25 + Gradle. Added the same setup, and gated the Gradle steps on a release label being present so Gradle only runs when a release is actually being cut.
- **Publish version was computed inside a `with:` value:** `version: $(echo ${{ ... }} | sed -e 's:v::')`. `with:` inputs are not shell-evaluated — this only worked by accident because the called workflow later splices the value into a `run:` line where the shell finally expands the `$(...)`. Replaced with a dedicated step using shell parameter expansion (`${NEW_VERSION#v}`), passing the tag via an env var (no interpolation into the command) and exposing a clean job output.
- **No concurrency guard on the release workflow → racing releases.** `tag` runs on every push to `minecraft/v*`. With two PRs merged close together, the second Tag run could read the same "latest tag" as the first before it pushed, compute a colliding next version, and double-publish to Maven Central (which is immutable — so one release fails or tags collide). Added a per-branch `concurrency` group with `cancel-in-progress: false`, so releases queue and finish one at a time and are never cancelled mid-publish.
- **Publish was called via a hardcoded, mutable branch ref:** `uses: …/publish.yml@minecraft/v26`. That pin had to be bumped at every Minecraft-version promotion, and a branch ref is mutable (anyone who can push the branch can rewrite the job that runs with the signing/Maven Central secrets). Switched to the same-repo local-path form `uses: ./.github/workflows/publish.yml`, which always resolves `publish.yml` from the exact commit that triggered the run. On `minecraft/v26` this is byte-for-byte the current behaviour; it just removes the hardcoded version and always matches the branch being released.
- **Gradle wrapper download spam leaked into release tag messages.** The `printPaperDetails` step captured `./gradlew … 2>&1` into the `paper_details` output, which is spliced into the annotated tag. On a cold runner the wrapper writes its `Downloading …10%…100%` distribution-bootstrap progress to **stdout** (confirmed by forcing a cold wrapper download locally — it's stdout, not stderr, so `--quiet`/`--console=plain`/dropping `2>&1` don't suppress it), so releases got the download dump in their notes (e.g. `v4.113.4`). The step now filters to just the `Paper …:` lines with `grep '^Paper'`, drops `2>&1`, and uses a non-colliding heredoc delimiter (closing a latent `GITHUB_OUTPUT` injection footgun where a captured line equal to `EOF` would have truncated the block).

### Security — supply chain & least privilege

- **No `permissions:` blocks anywhere.** Every workflow ran with the repository-default token scope. Scoped each to what it needs: `contents: read` for build/lint/coverage/publish, `pull-requests: read` for the label check (it only does a GraphQL read of the PR's labels), and `contents: read` by default for `tag` with the tag job elevated to `contents: write` + `pull-requests: read`.
- **All external actions pinned to commit digests.** A mutable tag can be repointed to malicious code that then runs with the workflow's privileges — the attack class behind the `tj-actions/changed-files` and recent TanStack/npm compromises. This matters most for the `tag` job, which runs five `actions-ecosystem/*` actions (a dormant third-party org) with `contents: write`. The pins are behaviour-neutral: each action already resolved to the release pinned here. The version is kept in a trailing comment, and Renovate keeps digests current.
- **Label-checker was on `:latest`.** A mutable tag on a required merge gate. Pinned to the v1.6.65 image digest (currently identical to `:latest`).
- **`secrets: inherit` → explicit secrets.** The publish call forwarded every repository secret to the reusable workflow. It now passes only the four secrets `publish.yml` declares (`OSSRH_USERNAME`, `OSSRH_TOKEN`, `KEY_PASS`, `PRIV_KEY`).

### Maintenance

- **`actions/checkout` v6 → v7** and **`codecov/codecov-action` v6 → v7** — both were a major version behind. checkout v7's only notable change (blocking fork-PR checkout under `pull_request_target`/`workflow_run`) is a no-op here; codecov v7 is a GPG key-account rotation. These are the only updates that change runtime behaviour.
- **Migrated test results upload off the deprecated `codecov/test-results-action`.** That standalone action is deprecated (and still ships a Node 20 entrypoint) in favour of folding test result uploads into the main action. The step now uses `codecov/codecov-action@v7` with `report_type: test_results`, keeping the same files glob, token, and `codecov_yml_path`. Verified working on this PR's run — the upload succeeds and the deprecation warning is gone.

### Efficiency & runner hygiene

- **Dropped `--refresh-dependencies`** from the Gradle invocations. It was added when the Paper API was a mutable `-SNAPSHOT` coordinate, where the same coordinate could resolve to different artifacts over time. Paper now ships immutable, fully-qualified coordinates (e.g. `26.1.2.build.57-stable`), so a version bump is a new coordinate and triggers a fresh download on its own. Forcing a refresh every run defeated the `setup-gradle` dependency cache; removing it lets the cache work, with downloads still happening exactly when a dependency version changes.
- **Tests ran twice per PR.** The `Build` check ran `./gradlew test`, but `Coverage` runs `./gradlew jacocoTestReport`, which already runs the whole suite (`jacocoTestReport dependsOn test`) — so every PR executed the test suite on two separate runners. `Build` now runs `./gradlew assemble testClasses`: it compiles main + test sources and packages the jar without running tests, checkstyle, or jacoco. Tests run once (in `Coverage`); the three checks stay distinct files with clearer, non-overlapping concerns — **Build = it compiles & packages**, **Coverage = tests pass + coverage**, **Linting = style**. The `Build / build` check name is unchanged, so branch-protection requirements keep working untouched.
- **Per-job `timeout-minutes` on every job** (build/coverage 20, linting 15, tag 15, publish 30, label-check 10). Without them a wedged Gradle daemon or a stuck upload runs against GitHub's 6-hour default.
- **Cancel superseded PR runs.** Added a `concurrency` group with `cancel-in-progress: true` to the PR-gating workflows (build, coverage, linting), so a new push to a PR cancels its now-stale in-flight runs. The release `tag` workflow deliberately uses its own group with `cancel-in-progress: false` (queue, never cancel).

## Note for reviewers

- The `permissions:` blocks are the one change worth watching on the first real run — particularly `bump-label-check` (a required gate) and the release `tag` job — to confirm the scoped tokens are sufficient.
- The `Build` job no longer runs tests (it compiles/assembles). Tests are still gated on every PR via `Coverage`. If `Build` is a required check in branch protection, it remains valid — the check name is unchanged.

## Checklist

- [ ] Unit tests added (if applicable). — N/A, CI configuration only.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -42,8 +42,8 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.2.0
 
-      - name: Build with Gradle
-        run: ./gradlew build -x test
+      - name: Assemble artifacts
+        run: ./gradlew assemble
 
       - name: Run unit tests
         run: ./gradlew test
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -33,13 +33,13 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Set up JDK 25
-        uses: actions/setup-java@v5
+        uses: actions/setup-java@1bcf9fb12cf4aa7d266a90ae39939e61372fe520 # v5.4.0
         with:
           java-version: 25
           distribution: 'temurin'
 
       - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@v6
+        uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.2.0
 
       - name: Build with Gradle
         run: ./gradlew build -x test --refresh-dependencies
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -42,7 +42,7 @@ jobs:
         uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.2.0
 
       - name: Build with Gradle
-        run: ./gradlew build -x test --refresh-dependencies
+        run: ./gradlew build -x test
 
       - name: Run unit tests
         run: ./gradlew test
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,6 +21,10 @@ on:
         required: true
         type: string
         description: "The version to publish (no v prefix)"
+
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,6 +28,7 @@ permissions:
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
 
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -30,7 +30,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Set up JDK 25
         uses: actions/setup-java@v5
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,9 @@ name: Build
 on:
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Set up JDK 25
         uses: actions/setup-java@v5
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,13 +14,13 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Set up JDK 25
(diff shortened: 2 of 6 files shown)


.github/workflows/publish.yml
```diff
@@ -21,24 +21,29 @@ on:
...
           java-version: 25
           distribution: 'temurin'
 
       - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@v6
+        uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.2.0
 
       - name: Build with Gradle
-        run: ./gradlew build -x test --refresh-dependencies
+        run: ./gradlew build -x test
```
Insprill: We skip tests here differently from the build workflow, which only calls `assemble testClasses`. It's probably fine now but it'd be possible for build to end up running some task that assemble doesn't, and we'll end up with publish failing even though all PR checks passed. I think we should use this version in the build task to ensure everything gets run except tests.

Insprill (changes requested):

Insprill (approved):
LGTM :)

.github/workflows/publish.yml
```diff
@@ -21,24 +21,29 @@ on:
...
         with:
           java-version: 25
           distribution: 'temurin'
 
       - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@v6
+        uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6.2.0
 
-      - name: Build with Gradle
-        run: ./gradlew build -x test --refresh-dependencies
```
Thorinwasher, line 41: Not testing again on publish makes sense, I'm more wondering why this was decided upon previously

Thorinwasher (approved):
LGTM

.github/workflows/publish.yml
Insprill: The `-x test` part prevented tests from running again, so it's basically the same; it's just some other misc tasks that we're avoiding by using `assemble`. Or were you referring to why we skip tests in publish?

Thorinwasher: Haha 😂 I think I was a too early in the morning when looking at this

4everTheOne:
LGTM!
Should we also consider this in renovate?
https://github.com/renovatebot/renovate/discussions/42031#discussioncomment-16229220

4everTheOne (approved):

PR #1588, opened 2026-07-07
fix(scheduler): fix #1586 - flaky longRunningTask_Throws_RunTimeException test

### LINKED ISSUE #1586: Randomly failing test for BukkitSchedulerMock

### Minecraft Version

N/A

### Describe the bug

This failure occurs randomly during CI runs:
```
BukkitSchedulerMockTest > longRunningTask_Throws_RunTimeException() FAILED
    org.opentest4j.AssertionFailedError at BukkitSchedulerMockTest.java:335
```

### Reproducible Test

Inconsistent behavior that occurs **randomly**

### Anything else?

_No response_

kccricket:
To be perfectly frank, I haven't put any effort into learning the nuance of Java's threading model, so I leaned on an LLM's analysis on this one.

You won't hurt my feelings if you reject this PR.

---

The async task read the shared testTask field before the main thread had finished assigning it, a data race that could NPE and kill the worker early, making getActiveRunningCount() read 0 instead of 1. Guard the reference with an AtomicReference, assert the latch actually fires instead of ignoring its result, and poll getActiveRunningCount() briefly since it is inherently approximate.

Fixes #1586

Claude-Session: https://claude.ai/code/session_01FdwuB4vK3jUk6QbQhLWPUn

### DIFF
--- a/src/test/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMockTest.java
+++ b/src/test/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMockTest.java
@@ -32,6 +32,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -307,13 +308,18 @@ class BukkitSchedulerMockTest
 
 		final CountDownLatch countDownLatch = new CountDownLatch(1);
 		final AtomicBoolean alive = new AtomicBoolean(true);
+		// Guarded by an AtomicReference (rather than the plain `testTask` field) so the task below
+		// never observes a null/stale reference: the task may start running before runTaskAsynchronously
+		// returns and assigns testTask, so reading that field directly here is a data race.
+		final AtomicReference<BukkitTask> taskRef = new AtomicReference<>();
 
 		testTask = scheduler.runTaskAsynchronously(null, () ->
 		{
 			countDownLatch.countDown();
 			while (alive.get())
 			{
-				if (testTask.isCancelled())
+				BukkitTask task = taskRef.get();
+				if (task != null && task.isCancelled())
 				{
 					alive.set(false);
 				}
@@ -329,9 +335,15 @@ class BukkitSchedulerMockTest
 				}
 			}
 		});
-		countDownLatch.await(1, TimeUnit.SECONDS);
+		taskRef.set(testTask);
+		assertTrue(countDownLatch.await(2, TimeUnit.SECONDS));
 
 		assertTrue(alive.get());
+		long deadline = System.currentTimeMillis() + 1000;
+		while (scheduler.getActiveRunningCount() != 1 && System.currentTimeMillis() < deadline)
+		{
+			Thread.sleep(10);
+		}
 		assertEquals(1, scheduler.getActiveRunningCount());
 		scheduler.performTicks(10);
 		scheduler.setShutdownTimeout(10);
--- a/src/test/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMockTest.java
+++ b/src/test/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMockTest.java
@@ -32,7 +32,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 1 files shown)


Thorinwasher:
Could you explain more in detail about what issue you have, and what this PR fixes?

Thorinwasher:
Aha, I see that you linked one issue to this. So nevermind

src/test/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMockTest.java
```diff
@@ -329,9 +335,15 @@ void longRunningTask_Throws_RunTimeException() throws InterruptedException
...
-		countDownLatch.await(1, TimeUnit.SECONDS);
+		taskRef.set(testTask);
+		assertTrue(countDownLatch.await(2, TimeUnit.SECONDS));
 
 		assertTrue(alive.get());
+		long deadline = System.currentTimeMillis() + 1000;
+		while (scheduler.getActiveRunningCount() != 1 && System.currentTimeMillis() < deadline)
+		{
+			Thread.sleep(10);
+		}
```
Thorinwasher: This seems a bit counterproductive; While the next upcomming assertion will fail, sleep for 10 ms.

Thorinwasher (commented):
I'm a bit unsure with what the test is trying to test. That might be a better approach to go about fixing this problem. I would personally have used locks or something instead of sleep.

kccricket:
It looks like the test is intended to assert that long-running tasks are properly handled during a server shutdown. Good feedback. I'll take it back and rework it.

kccricket:
Let's try this. Use a CountDownLatch in testStarted to indicate to the test that the scheduled task has started. Use an infinitely waiting CountDownLatch in the task to hang the thread. Only continue the test once taskStarted resolves. Assert that there is one active task, set the shutdown timer to something tiny, and then shut down the scheduler, asserting the exception.

I'd expect this to be more reliable than the previous implementation.

Apologies for the churn. I didn't do my due diligence on the previous iteration.

src/test/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMockTest.java
```diff
@@ -305,35 +305,27 @@ void longRunningTask_Throws_RunTimeException() throws InterruptedException
...
-				}
+				// Block until the scheduler force-interrupts us during shutdown.
+				new CountDownLatch(1).await(); // never counted down
+			}
+			catch (InterruptedException e)
+			{
+				throw new TaskCancelledException("Interrupted", e);
 			}
 		});
-		countDownLatch.await(1, TimeUnit.SECONDS);
```
frengor, line 332: Shouldn't increasing the timeout here (or just using `countDownLatch.await()`) resolve the issue? It seems to me the other changes are just AI slop.

frengor: Also, `await` doesn't throw an exception when the timeout expires, it just returns `false` ([await javadoc](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/CountDownLatch.html#await(long,java.util.concurrent.TimeUnit))). So basically the test is not checking properly if the task has been started. The entire await call should be inside an `assertTrue` here like in [this test](https://github.com/MockBukkit/MockBukkit/blob/fc492ec6d006be2fd85fb259e4ac832e7b92105b/src/test/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMockTest.java#L615).

kccricket: It's a little half and half: LLM-assisted, but I'm by no means "vibe coding." I generally keep a tight grip. I use it to fill my knowledge gaps, because Java's not my day job.

The original test was intended to assert that long-running scheduler threads properly throw on scheduler shutdown (correct me if I'm wrong). However, it was also inadvertently testing how fast the testbed could run the tests.

With a no-arg `taskStarted.await()`, as I've made it, it's not making guesses about how fast the testbed is. This does mean if the task never starts it will hang until GitHub kills the action, but any occurrence of that would be a useful signal, in my opinion.

I don't mind rolling back to the original version (or some hybrid) and `assertTrue` on `countDownLatch` with a longer timeout. It just depends on your (the maintainers') tolerance for non-determinism. I could set the timeout to five seconds, and it'd *probably* never flake on a slow testbed, but it's not guaranteed.

thelooter:
I traced this through `BukkitSchedulerMock` and I don't think a longer/unbounded `countDownLatch.await()` alone would have fixed the flake — there are actually two independent races in the old test, and the timeout is only one of them.

**Race 1 (the one the latch can't fix):** 
The task body reads the `testTask` *instance field*:

```java
testTask = scheduler.runTaskAsynchronously(null, () -> {
    countDownLatch.countDown();
    while (alive.get()) {
        if (testTask.isCancelled()) { ... }   // reads the field
        ...
    }
});
``` 

`runTaskAsynchronously` calls `pool.execute(...)` *before* it returns ([`BukkitSchedulerMock.java:561-571`](https://github.com/kccricket/MockBukkit/blob/fix/flaky-scheduler-test-1586/src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java#L561-L572)), so the worker can be running before the main thread assigns `testTask` — and `Executor.execute` only establishes happens-before for writes made *before* the submit. If the worker wins that race, `testTask.isCancelled()` NPEs. Crucially, `AsyncRunnable.run()` swallows that NPE into `asyncException` ([`BukkitSchedulerMock.java:844-855`](https://github.com/kccricket/MockBukkit/blob/fix/flaky-scheduler-test-1586/src/main/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMock.java#L844-L855) and the worker dies silently. Since `countDown()` already fired, the latch releases normally and the main thread then sees `getActiveRunningCount() == 0` — which is exactly the reported failure at line 335. No timeout value changes this; the fix is removing the shared field read, which this PR does by blocking on a latch instead. So that part of the change is load-bearing, not incidental.


**Race 2** is the ignored `await` return value, which this PR also fixes.

On the unbounded `taskStarted.await()` question: I'd suggest the middle ground of

```java
assertTrue(taskStarted.await(30, TimeUnit.SECONDS));
```

30s is far above any plausible thread-start latency, so it keeps the "a slow
machine only makes the test slower, never changes the verdict" property in
practice — but if the pool somehow never runs the task, we get a readable
assertion failure instead of a hung job that runs until GitHub kills it. It also
matches the existing convention elsewhere in this file:
https://github.com/MockBukkit/MockBukkit/blob/fc492ec6d006be2fd85fb259e4ac832e7b92105b/src/test/java/org/mockbukkit/mockbukkit/scheduler/BukkitSchedulerMockTest.java#L615

One last nit: the PR description still describes an earlier iteration
(`AtomicReference` + polling `getActiveRunningCount()`) — worth updating it to
match the current diff before merge so the history stays accurate.

With the bounded await + description update, this LGTM.

Btw, given the origin of this PR, it's only fair to admit that this review was aided by Claude Fable 5

frengor:
> `runTaskAsynchronously` calls `pool.execute(...)` before it returns

I don't know why it was implemented like this, but from what I can see from Paper's scheduler impl, this should be incorrect. `runTaskAsynchronously` should behave like `runTask` in respect to the delay, i.e. it should be implemented by calling `runTaskLaterAsynchronously` with a delay of 1, not (be scheduled to) run immediately. Changing the `runTaskAsynchronously` impl should fix the first race, since the task won't run until `scheduler.performTicks(...)` is called.

PR #1592, opened 2026-07-10
Upgrade to PaperMC v26.2

4everTheOne:
# Description

# Checklist

---
<a href="https://app.blacksmith.sh/MockBukkit/codesmith/MockBukkit/pr/1592"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786279154&installation_id=145442630&pr_number=1592&repository=MockBukkit%2FMockBukkit&return_to=https%3A%2F%2Fgithub.com%2FMockBukkit%2FMockBukkit%2Fpull%2F1592&signature=502e3bc340d4eecf05b1b5370fe3d27bc40137b1d27b0eedd023662bccfbdb94"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>


src/main/java/org/mockbukkit/mockbukkit/entity/PlayerMock.java
```diff
@@ -3324,10 +3328,7 @@ public void sendMessage(@NotNull ChatMessageType position, @Nullable UUID sender
 		{
 			Preconditions.checkNotNull(position, "Position must not be null");
 			Preconditions.checkNotNull(components, "Component must not be null");
-			Component comp = BungeeComponentSerializer.get().deserialize(components);
-			String serialized = LegacyComponentSerializer.legacySection().serialize(comp);
-			comp = LegacyComponentSerializer.legacySection().deserialize(serialized);
-			PlayerMock.this.sendMessage(sender == null ? Identity.nil() : Identity.identity(sender), comp);
```
4everTheOne, line 3330: This migration might not be correct and needs to be validated

src/main/java/org/mockbukkit/mockbukkit/command/ConsoleCommandSenderMock.java
```diff
@@ -246,7 +245,7 @@ public void sendMessage(@Nullable UUID sender, @NotNull BaseComponent... compone
 		{
 			Preconditions.checkNotNull(components, "Components must not be null");
 			Component comp = BungeeComponentSerializer.get().deserialize(components);
-			ConsoleCommandSenderMock.this.sendMessage(sender == null ? Identity.nil() : Identity.identity(sender), comp);
+			ConsoleCommandSenderMock.this.sendMessage(comp);
 		}
```
4everTheOne, line 250: This migration might not be correct and needs to be checked

src/main/java/org/mockbukkit/mockbukkit/entity/SulfurCubeMock.java
4everTheOne, line 1: We should add tests here also

x9nico (approved):

thelooter (commented):
Reviewed the hand-written Java against a local Paper 26.2 checkout (`paper-api` + `paper-server`), and verified each behavioural finding by running it against this branch rather than reading only. The ~10k lines of `resources-autogenerated` churn I spot-checked instead of line-reviewing — it reads as insertions plus ordinal renumbering, as expected from a regen.

Three of the comments below are behavioural (`VexMock`, `WorldMock`, `TropicalFishBucketMetaMock`); the rest are cleanup, a nit, and one test-coverage gap. Nothing architectural.

Nice work on `BungeeComponentUtils` in particular — routing through the JSON chat representation to dodge the Adventure 5 `downsampleColors()` removal is the right call, and the javadoc explaining *why* will save the next person a lot of time.

src/main/java/org/mockbukkit/mockbukkit/entity/VexMock.java
```diff
@@ -104,6 +106,18 @@ public void setSummoner(@Nullable Mob summoner)
...
+	public @Nullable LivingEntity getOwner()
+	{
+		return this.owner;
+	}
+
+	@Override
+	public void setOwner(@Nullable LivingEntity owner)
+	{
+		this.owner = owner;
+	}
```
thelooter, line 105: In 26.2 `getSummoner()` / `setSummoner()` became **default methods on `Vex`** that delegate to the owner:

```java
// paper-api — Vex.java
@Deprecated(forRemoval = true, since = "26.2")
default @Nullable Mob getSummoner() {
    return this.getOwner() instanceof Mob mob ? mob : null;
}

@Deprecated(forRemoval = true, since = "26.2")
default void setSummoner(final @Nullable Mob summoner) {
    this.setOwner(summoner);
}
```

Here the new `owner` field (L30) sits alongside the pre-existing `summoner` field and **both** pairs of accessors are overridden, so the two halves never see each other. Verified against this branch:

```
after setSummoner(mob) -> getOwner()    == null
after setOwner(mob)    -> getSummoner() == null
```

Suggest dropping the `summoner` field and the `getSummoner()`/`setSummoner()` overrides entirely so the interface defaults take over — that gets the delegation semantics for free and keeps them correct if Paper changes the mapping.

`VexMockTest` has separate `SetSummoner` and `SetOwner` nests that never cross; one test writing through one accessor and reading through the other would pin this.

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -1497,6 +1497,13 @@ public <T> void playEffect(@NotNull Location location, @NotNull Effect effect, @
 		return new ChunkSnapshotMock(chunkX, chunkZ, getMinHeight(), getMaxHeight(), getName(), getFullTime(), chunkBlockData.build(), (includeBiome || includeBiomeTempRain) ? chunkBiomes.build() : null);
 	}
 
+	@Override
+	public void setAllowMonsterSpawning(boolean allowMonsters)
+	{
+		// TODO Auto-generated method stub
+		throw new UnimplementedOperationException();
+	}
```
thelooter, line 1504: `setSpawnFlags(boolean, boolean)` is `@Deprecated(since = "26.2", forRemoval = true)` and its javadoc points users straight here:

> @deprecated the vanilla server no longer maintains this functionality. See `{@link #setAllowMonsterSpawning(boolean)}` if you want to dis/allow monster spawning.

So as it stands, the replacement throws while the deprecated method it replaces works — following Paper's own migration advice moves you from working code to an exception.

This isn't really "not implemented yet": the backing state already exists. `setSpawnFlags` immediately below writes `this.allowMonsters`, and `getAllowMonsters()` reads it.

```java
@Override
public void setAllowMonsterSpawning(boolean allowMonsters)
{
	this.allowMonsters = allowMonsters;
}
```

Matches `CraftWorld#setAllowMonsterSpawning`, which likewise only touches the monster flag.

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/TropicalFishBucketMetaMock.java
```diff
@@ -116,9 +116,28 @@ public void setPattern(TropicalFish.@NotNull Pattern pattern)
...
+	public boolean hasBodyColor()
+	{
+		return this.bodyColor != null;
+	}
+
+	@Override
+	public boolean hasPatternColor()
+	{
+		return this.patternColor != null;
+	}
```
thelooter, line 117: `hasVariant()` just below now matches `CraftMetaTropicalFishBucket` exactly — good catch on the AND→OR.

These three new predicates can't actually distinguish the properties, though, because every setter calls `checkVars()`, which eagerly defaults **all three** fields:

```java
protected void checkVars()
{
	if (this.patternColor == null) this.patternColor = DyeColor.WHITE;
	if (this.bodyColor == null)    this.bodyColor    = DyeColor.WHITE;
	if (this.pattern == null)      this.pattern      = TropicalFish.Pattern.KOB;
}
```

Observed on this branch:

| | `hasPattern` | `hasBodyColor` | `hasPatternColor` |
|---|---|---|---|
| fresh meta | `false` | `false` | `false` |
| after `setPattern(KOB)` only | `true` | `true` :x: | `true` :x: |

`getBodyColor()` then returns `WHITE` where Paper has nothing set at all.

Since the entire point of splitting `hasVariant()` into three predicates is granularity, `checkVars()` should only default the field being written — or go away.

src/main/java/org/mockbukkit/mockbukkit/entity/SulfurCubeMock.java
```diff
@@ -0,0 +1,221 @@
...
+		//TODO: Auto-generated method stub
+		throw new UnimplementedOperationException();
+	}
+
+	@Override
+	public void setBreed(boolean breed)
+	{
+		//TODO: Auto-generated method stub
+		throw new UnimplementedOperationException();
+	}
```
thelooter: Minor / free win: `CraftSulfurCube` hardcodes both of these rather than deriving them from state, so they don't need to throw:

```java
@Override
public boolean canBreed() {
    return false;
}

@Override
public void setBreed(final boolean breed) {
}
```

For what it's worth, the rest of this class cross-references cleanly against `CraftSulfurCube` — `setFuseTicks`'s precondition is character-identical (and `PrimedTnt.NO_FUSE == -1` checks out), and the bucket item / pickup sound / `fromBucket` default all match vanilla `SulfurCube`.

src/main/java/org/mockbukkit/mockbukkit/util/UnsafeValuesMock.java
```diff
@@ -399,92 +299,6 @@ public Entity deserializeEntity(byte[] data, World world, boolean preserveUUID)
...
-		// e.g: Material#STONE is a block (but also an obtainable item in the inventory). However, the translation key is always "block.minecraft.stone".
-		if (material.isBlock())
-		{
-			return formatTranslatable("block", material);
-		}
-		// not an edge case
-		return null;
-	}
-
 	private <T extends Keyed & Translatable> String formatTranslatable(String prefix, T translatable, boolean fromItemStack)
```
thelooter: This PR removed every caller of `formatTranslatable` from this class — `getBlockTranslationKey`, `getItemTranslationKey`, both `getTranslationKey` overloads and `handleTranslateItemEdgeCases` all went.

Both overloads are `private`, and the only remaining reference is the 2-arg one at L322 calling this one, so the pair is now mutually-recursive dead weight. Safe to delete outright (~25 lines, including the wall-sign handling and the `emptyEffects` set) — the logic lives on in `ItemStackMock#translationKey()` and the generated keys.

src/main/java/org/mockbukkit/mockbukkit/MockBukkitInternalAPIBridge.java
```diff
@@ -123,4 +230,29 @@ public Set<Pose> validMannequinPoses()
...
+			{
+				value += ".effect.empty";
+			}
+		}
+		return String.format("%s.%s.%s", prefix, translatable.key().namespace(), value);
+	}
+
+	private <T extends Keyed & Translatable> String formatTranslatable(String prefix, T translatable)
+	{
+		return formatTranslatable(prefix, translatable, false);
```
thelooter: The `fromItemStack` flag is dead here: the 2-arg overload below hardcodes `false`, and its only caller is `getTranslationKey(EntityType)` at L78. Both overloads are `private`, so nothing outside the class can reach them either.

That makes the whole `translatable instanceof Material` block unreachable — the wall-sign/banner/torch/head handling *and* the `emptyEffects` `.effect.empty` suffix. An `EntityType` is never a `Material`.

Can collapse to:

```java
private <T extends Keyed & Translatable> String formatTranslatable(String prefix, T translatable)
{
	return String.format("%s.%s.%s", prefix, translatable.key().namespace(), translatable.key().value());
}
```

The item-side logic now has a better home in `ItemStackMock#translationKey()` plus the generated `item.json`/`block.json` keys.

src/main/java/org/mockbukkit/mockbukkit/MockBukkitInternalAPIBridge.java
```diff
@@ -45,12 +71,93 @@ public DamageEffect getDamageEffect(String key)
...
 
+	@Override
+	public String getTranslationKey(EntityType type)
+	{
+		Preconditions.checkArgument(type.getName() != null, "Invalid name of EntityType %s for translation key", type);
+		return formatTranslatable("entity", type);
+	}
+
+	@Override
+	public ItemStack deserializeItem(byte[] data)
```
thelooter, line 87: This is now the live `deserializeItem` — it's what `ItemStack.deserializeBytes()` routes through. But `UnsafeValuesMock#deserializeItem` (L235) still exists with byte-for-byte identical logic and a **different** result for the same input:

| | truncated / header-only data |
|---|---|
| `MockBukkitInternalAPIBridge` (here) | `ItemStack.empty()` |
| `UnsafeValuesMock#deserializeItem` | `null` |

To be clear, keeping the `UnsafeValuesMock` copy is the right call — `serializeItem`/`deserializeItem` are gone from the `UnsafeValues` interface in 26.2, but plugins may still call them on the mock, and since `ItemStack#serializeAsBytes()` NPEs on the null `craftDelegate`, `UnsafeValuesMock#serializeItem` is currently the only working way to get item bytes out of MockBukkit at all.

The two just shouldn't disagree. Suggest having one delegate to the other, the way `deserializeStack` now goes through `deserializeStackStatic`.

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -37,6 +37,9 @@
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 
+import static org.mockbukkit.mockbukkit.util.UnsafeValuesMock.PROPERTY_SCHEMA_VERSION;
+import static org.mockbukkit.mockbukkit.util.UnsafeValuesMock.RENAME_JSON_PROPERTY;
```
thelooter: Nit: these two constants were widened from `private` to `public` on `UnsafeValuesMock` purely to enable this static import.

Since `ItemStackMock#serialize()` is now the only producer of this format, they'd sit more naturally as private constants *here* than as new public fields on a class annotated `@Deprecated(since = "1.7.2")`. Not blocking if you'd rather keep them reachable.

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -605,4 +611,71 @@ public static ItemStack deserialize(@NotNull Map<String, Object> args)
...
+				serializedMeta.remove(originalName);
+			}
+			result.put("components", serializedMeta);
+		}
+
+		return result;
+	}
+
+	@Override
+	public @NotNull String translationKey()
```
thelooter, line 769: Note: the `.effect.empty` suffix is appended unconditionally for the four potion materials, regardless of what potion data the stack actually carries. Real Paper returns `item.minecraft.potion.effect.<type>` once a potion type is set and only falls back to `.effect.empty` when there is none.

This is carried over from the old `UnsafeValuesMock#getTranslationKey(ItemStack)` so it's not a regression — but the comment on `EMPTY_EFFECT_ITEMS` says "when no effect is present", which is the one thing the code doesn't check.

Otherwise this rewrite is a clear improvement: driving off `ItemType`/`BlockType` means the generated data supplies the real vanilla strings instead of the old prefix-guessing.

src/test/java/org/mockbukkit/mockbukkit/util/UnsafeValuesTest.java
```diff
@@ -524,30 +505,6 @@ static Stream<Arguments> wallMaterialTranslationKeyProvider()
...
-	{
-		return Stream.of(
-				Arguments.of("item.minecraft.saddle", new ItemStackMock(Material.SADDLE)),
-				Arguments.of("block.minecraft.stone", new ItemStackMock(Material.STONE)),
-				Arguments.of("item.minecraft.wheat", new ItemStackMock(Material.WHEAT)),
-				Arguments.of("item.minecraft.nether_wart", new ItemStackMock(Material.NETHER_WART))
-		);
-	}
-
 	@ParameterizedTest
```
thelooter, line 552: `testItemStackTranslationKey` was deleted here along with the `UnsafeValues` methods it covered, but nothing replaced it against the new `ItemStackMock#translationKey()`.

I checked the behaviour is preserved — all four of the old expectations still hold on this branch:

```
STONE       -> block.minecraft.stone
WHEAT       -> item.minecraft.wheat
NETHER_WART -> item.minecraft.nether_wart
SADDLE      -> item.minecraft.saddle
```

So no regression, but the net is gone. The surviving `itemStackEmptyEffectTranslationKeyProvider` only exercises the four potion items, which leaves the interesting branch untested: block-items like `STONE` take the `isItem()` path and still have to resolve to a `block.` key. Worth porting the old cases over to `itemStack.translationKey()`.

src/main/java/org/mockbukkit/mockbukkit/entity/VexMock.java
4everTheOne: Resolved in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/1d80d5f59179b29b6b59c5610b10e160401b821d

src/test/java/org/mockbukkit/mockbukkit/util/UnsafeValuesTest.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/89428e01820027e4c35bcbc7a09ffe14de4cf72e

src/main/java/org/mockbukkit/mockbukkit/entity/SulfurCubeMock.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/0371c70fba9dfbce28454f33c607b6a270f5eab8

src/main/java/org/mockbukkit/mockbukkit/util/UnsafeValuesMock.java
4everTheOne: Removed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/ecc9471ade302cb80f67841083ff71a64f43a880

src/main/java/org/mockbukkit/mockbukkit/MockBukkitInternalAPIBridge.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/ff17ede23230da8666ee7e18a7bc9bc260d1da77

src/main/java/org/mockbukkit/mockbukkit/MockBukkitInternalAPIBridge.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/270a195223ddc8351e4d76b362a17f9998ef66b9

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
4everTheOne: That is a good point, but is the current behavior in the system. I tried with this payload on this branch and on the main branch and they have the same behaviour:
```java
static Stream<Arguments> itemStackTranslationKeyProvider()
	{
		ItemStack healing = ItemStack.of(Material.POTION);
		healing.editMeta(PotionMeta.class, potionMeta -> {
			potionMeta.setBasePotionType(PotionType.HEALING);
		});

		return Stream.of(
				Arguments.of("item.minecraft.saddle", ItemStack.of(Material.SADDLE)),
				Arguments.of("block.minecraft.stone", ItemStack.of(Material.STONE)),
				Arguments.of("item.minecraft.wheat", ItemStack.of(Material.WHEAT)),
				Arguments.of("item.minecraft.nether_wart", ItemStack.of(Material.NETHER_WART)),
				Arguments.of("item.minecraft.potion.effect.empty", ItemStack.of(Material.POTION)),
				Arguments.of("item.minecraft.potion.effect.healing", healing)
		);
	}
```

If you agree, I would suggest opening a new issue with this information to be worked in parallel.

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/03596571388f1a6c7b2b38c3551fcef77df2da40 by moving the logic inside the ItemStack. Now he is the owner of the logic

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/TropicalFishBucketMetaMock.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/ad5b6cc33bf0097a7c66ee0efee10df09824e211

src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/ef038da666dfd945fec4157bd807fe7cc80a1f40

thelooter (commented):
Re-reviewed at `ef038da6`. Nine of ten previous comments addressed; build blocker gone — green against build 111, tests and checkstyle pass.

Adversarial pass turned up the below. Each is tagged **[this PR]** or **[pre-existing]**, verified against a worktree at the merge base (`21cc6c48`) rather than guessed.

Nothing blocking. Worth fixing here: the TropicalFish getter exception type, and `setRotation(Angle, Angle)`.

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/TropicalFishBucketMetaMock.java
```diff
@@ -111,14 +95,32 @@ public void setBodyColor(@NotNull DyeColor color)
...
+	@Override
+	public boolean hasBodyColor()
+	{
+		return this.bodyColor != null;
+	}
+
+	@Override
+	public boolean hasPatternColor()
+	{
+		return this.patternColor != null;
```
thelooter, line 116: **[partly this PR]**

The AND→OR fix and the `checkVars()` removal are both right. But they change how often this path runs, and the getters underneath diverge from Paper.

Paper uses `Preconditions.checkState` with an actionable message:

```java
// CraftMetaTropicalFishBucket
public DyeColor getBodyColor() {
    Preconditions.checkState(this.hasBodyColor(), "Body color is absent, check hasBodyColor first!");
    ...
}
```

MockBukkit uses commons-lang3 `Validate.notNull`, so callers get `NullPointerException` instead of `IllegalStateException`.

On attribution — I built a worktree at the merge base (`21cc6c48`) and ran both sides:

```
BASE  fresh.getBodyColor()                      -> NullPointerException
BASE  after setPattern() only -> getBodyColor() =  WHITE      <- checkVars filled it in
PR    fresh.getBodyColor()                      -> NullPointerException
PR    after setPattern() only -> getBodyColor() -> NullPointerException
```

So the exception *type* is pre-existing and untouched by this PR. What changed is **reachability**: `checkVars()` meant any setter populated all three fields, so the unset branch was only hit on a completely fresh meta. Now unset is the normal state and this sits on the common path.

Since the exposure and the fix arrive together, I'd switch the three getters (`getPatternColor` L63, `getBodyColor` L76, `getPattern` L89) to `IllegalStateException` here. `TropicalFishBucketMetaMockTest` currently pins `assertThrowsExactly(NullPointerException.class, ...)`, so those three assertions move with it.

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
```diff
@@ -1238,6 +1242,13 @@ public void setRotation(float yaw, float pitch)
 		location.setPitch(pitch);
 	}
 
+	@Override
+	public void setRotation(@NotNull Angle yaw, @NotNull Angle pitch)
```
thelooter, line 1246: **[this PR]**

`setRotation(float, float)` directly above works; this overload throws. `Angle` is a thin wrapper — `degrees()` plus a `relative()` flag — so it's delegable rather than a stub:

```java
@Override
public void setRotation(@NotNull Angle yaw, @NotNull Angle pitch)
{
	setRotation(
			yaw.relative() ? location.getYaw() + yaw.degrees() : yaw.degrees(),
			pitch.relative() ? location.getPitch() + pitch.degrees() : pitch.degrees());
}
```

Worth noting Paper does **not** deprecate the float overload here (unlike `World#setSpawnFlags`), so this is a coverage gap rather than a broken migration path — materially lower stakes than the `setAllowMonsterSpawning` case. Every entity mock inherits this one, so it's a single fix.

src/main/java/org/mockbukkit/mockbukkit/MockBukkitInternalAPIBridge.java
```diff
@@ -123,4 +235,9 @@ public Set<Pose> validMannequinPoses()
 		return Set.of(Pose.STANDING, Pose.SNEAKING, Pose.SWIMMING, Pose.FALL_FLYING, Pose.SLEEPING);
 	}
 
+	public static <T extends Keyed & Translatable> String formatTranslatable(String prefix, T translatable)
```
thelooter: **[this PR]**

Widened from `private` to `public static` in ff17ede2, but it still has exactly one caller: `getTranslationKey(EntityType)` at L76, in this same class. Nothing outside the class references it.

The class is `@ApiStatus.Internal` so this isn't an API commitment in spirit, but `private static` communicates the same thing without widening the surface.

(Contrast the `PROPERTY_SCHEMA_VERSION` / `RENAME_JSON_PROPERTY` move — those had a genuine cross-class consumer, and relocating them into `ItemStackMock` so the owner holds them was exactly right.)

src/test/java/org/mockbukkit/mockbukkit/inventory/meta/TropicalFishBucketMetaMockTest.java
```diff
@@ -110,20 +101,6 @@ void setPattern_SetsPattern()
...
-	}
-
-	@Test
-	void equals_DifferentObjects_SameValues_True()
-	{
-		meta.checkVars();
-		assertEquals(meta, meta.clone());
-	}
-
 	@Test
```
thelooter, line 104: **[this PR]**

`equals_SameInstance_True` and `equals_DifferentObjects_SameValues_True` were dropped here because they called `checkVars()`. That removes the only assertion that `meta.equals(meta.clone())` — and the copy constructor is precisely what this PR rewrote, now branching on `hasPattern()` / `hasBodyColor()` / `hasPatternColor()`.

Both surviving clone tests (`cloneConstructor_CopiesValues`, `clone_CopiesValues`) set all three fields, so the new conditional branches are only exercised in the all-set case. A partial-state case would cover the change directly:

```java
@Test
void clone_PartialState_PreservesAbsence()
{
	meta.setBodyColor(DyeColor.RED);
	TropicalFishBucketMetaMock clone = meta.clone();

	assertTrue(clone.hasBodyColor());
	assertFalse(clone.hasPattern());
	assertEquals(meta, clone);
}
```

I ran that shape against the branch and it passes — the behaviour is correct, it just isn't pinned.

src/main/java/org/mockbukkit/mockbukkit/damage/DamageSourceMock.java
```diff
@@ -70,6 +80,13 @@ public boolean scalesWithDifficulty()
 		};
 	}
 
+	@Override
+    @ApiStatus.Experimental
```
thelooter: **[this PR]**

Nit: this line is indented with spaces while the rest of the file uses tabs. Checkstyle doesn't catch it.

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
```diff
@@ -602,7 +630,149 @@ public int hashCode()
...
+
+				// Skip the key if it does not exist
+				if (!components.containsKey(originalName))
+				{
+					continue;
+				}
+
+				var value = components.get(originalName);
+				components.put(newName, value);
+				components.remove(originalName);
```
thelooter: **[pre-existing]**

**Not caused by this PR** — raising it because the block moved here and this is its home now.

These ~20 lines came over verbatim from `UnsafeValuesMock.deserializeStack` (base `UnsafeValuesMock.java:672-690`) because Paper dropped `serializeStack`/`deserializeStack` from `UnsafeValues` in 26.2. I built a worktree at the merge base (`21cc6c48`) and confirmed all four behaviours below reproduce identically there, so none of this is a regression.

**1. Empty stacks can't round-trip.** `serialize()` omits `count` for an empty stack, but `count` is unboxed at L644 — *above* the air early-return further down:

```
ItemStack.empty().serialize()  ->  {id=minecraft:air, schema_version=1, DataVersion=1}
ItemStack.deserialize(that)    ->  NullPointerException
```

Identical on base: 26.1.2's `ItemStack.serialize()` compiled to `Bukkit.getUnsafe().serializeStack(this)` (checked with `javap`), so the same two-call sequence already failed. Moving the air check above the `count` read fixes it.

**2. The caller's map is mutated in place** (L660-661). The nested `components` map is rewritten under the caller:

```
before = {meta-type=UNSPECIFIC, minecraft:unbreakable=true}
after  = {meta-type=UNSPECIFIC, Unbreakable=true}
```

Deserialize an item out of a YAML config and that config's in-memory map now holds MockBukkit's internal key names; re-saving persists the wrong format. A defensive copy of `components` fixes it.

**3. Unchecked casts on caller-supplied data** (L643-645) surface as `NullPointerException` / `ClassCastException` rather than `IllegalArgumentException`:

| input | result |
|---|---|
| `{id: "minecraft:stone"}` | `NullPointerException` |
| `{id: "minecraft:stone", count: "3"}` | `ClassCastException` |
| `{id: 42, count: 1}` | `ClassCastException` |

**4. `version` (L642) is computed and never read**, kept quiet by `@SuppressWarnings({"java:S1481","java:S1854"})`, next to the TODO about the unimplemented legacy-schema branch.

Entirely reasonable to spin these out into a separate issue rather than grow this PR — I just didn't want them to get lost now that they live at a new address.

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
thelooter: Agreed on deferring this to a separate issue. One correction for the write-up, though — I couldn't reproduce the `healing` case.

Ran your payload against `ef038da6`:

```
plain POTION        = item.minecraft.potion.effect.empty
HEALING POTION      = item.minecraft.potion.effect.empty     <- expected item.minecraft.potion.effect.healing
healing stack class = org.mockbukkit.mockbukkit.inventory.ItemStackMock
```

`ItemStackMock#translationKey()` decides the suffix purely from `EMPTY_EFFECT_ITEMS.contains(material)` and never reads the stack's `PotionMeta`, so it can't return `.effect.healing` for any input — the `Arguments.of("item.minecraft.potion.effect.healing", healing)` row would fail.

The committed test in 486b8dbb only added the `.effect.empty` row, so CI never exercised the healing case either.

Doesn't change the conclusion — still pre-existing, still fine as a follow-up. Just worth the issue recording "always returns `.effect.empty` regardless of potion data" rather than "already matches main", since the fix is to read the meta.

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/9e17c726e26d216f50716161afe88cd87275669f

src/main/java/org/mockbukkit/mockbukkit/damage/DamageSourceMock.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/a429acfd4130d7619252834d3606409574743d94

src/main/java/org/mockbukkit/mockbukkit/inventory/meta/TropicalFishBucketMetaMock.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/85e305eb6a8312a4efeb38ae389b39af35075886

src/test/java/org/mockbukkit/mockbukkit/inventory/meta/TropicalFishBucketMetaMockTest.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/87177b0a79720f0335b65d9ca4cd50ee3c336351

src/main/java/org/mockbukkit/mockbukkit/inventory/ItemStackMock.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/92134d47ebb81a8ef980427378da0eef01f4390d

src/main/java/org/mockbukkit/mockbukkit/MockBukkitInternalAPIBridge.java
4everTheOne: Fixed in https://github.com/MockBukkit/MockBukkit/pull/1592/changes/c0fd79fbc4dcd1629677e2d7d7e0da5b73b4a5f6

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
```diff
@@ -1238,6 +1242,34 @@ public void setRotation(float yaw, float pitch)
...
+	{
+		NumberConversions.checkFinite(yaw.degrees(), "yaw not finite");
+		NumberConversions.checkFinite(pitch.degrees(), "pitch not finite");
+
+		float yawValue = Location.normalizeYaw(yaw.degrees());
+		float pitchValue = Location.normalizePitch(pitch.degrees());
```
thelooter: **Bug: a relative pitch delta gets clamped twice, so large deltas are silently truncated.**

`Location.normalizePitch` *clamps* to ±90 (unlike `normalizeYaw`, which wraps mod 360). Pre-normalizing here truncates the delta before it is ever added to the current pitch.

Reproduced against this branch:

```java
entity.setRotation(0f, -80f);
entity.setRotation(Angle.relative(0f), Angle.relative(100f));
// actual:   10.0
// expected: 20.0   -> normalizePitch(-80 + 100)
```

`pitchValue` becomes `normalizePitch(100) == 90`, then `normalizePitch(-80 + 90) == 10`. Any relative pitch delta with `|degrees| > 90` is wrong.

The yaw path is fine — `normalizeYaw(a + normalizeYaw(d)) == normalizeYaw(a + d)` holds, including across the ±180 wrap — but it is simpler to treat both the same way and normalize exactly once, at the end:

```java
float yawValue = yaw.relative() ? location.getYaw() + yaw.degrees() : yaw.degrees();
float pitchValue = pitch.relative() ? location.getPitch() + pitch.degrees() : pitch.degrees();

location.setYaw(Location.normalizeYaw(yawValue));
location.setPitch(Location.normalizePitch(pitchValue));
```

The new `setRotation_AngleRelative_AddsToCurrent` test uses deltas of `+30`/`-10`, which stay in range and so cannot catch this. Worth adding a case with an out-of-range relative pitch.

src/main/java/org/mockbukkit/mockbukkit/entity/SulfurCubeMock.java
```diff
@@ -0,0 +1,219 @@
...
+import org.bukkit.entity.SulfurCube;
+import org.bukkit.inventory.ItemStack;
+import org.jspecify.annotations.NullMarked;
+import org.mockbukkit.mockbukkit.ServerMock;
+import org.mockbukkit.mockbukkit.exception.UnimplementedOperationException;
+
+import java.util.UUID;
+
+@NullMarked
+public class SulfurCubeMock extends MobMock implements SulfurCube
```
thelooter: **Suggestion: extend `AgeableMock` instead of `MobMock`.**

`SulfurCube extends AbstractCubeMob (→ Creature), Shearable, Bucketable, Ageable`, and `AgeableMock extends CreatureMock extends MobMock` — so it is a drop-in replacement for the superclass here, and the `(ServerMock, UUID)` constructor signature is unchanged.

As written, this class hand-rolls the entire `Ageable` contract that `AgeableMock` already implements with real, working behaviour:

* `getAge`, `setAge`, `setAgeLock`, `getAgeLock`, `setBaby`, `setAdult`, `isAdult` → `UnimplementedOperationException`
* `canBreed()` → hardcoded `false`
* `setBreed(boolean)` → silent no-op

Switching the superclass drops all nine overrides (~60 lines) and makes age-related plugin logic actually testable against this mock. `DolphinMock` and `SquidMock` are the existing precedent for a non-`Animals` `Ageable`.

Two related notes while you are in here:

* The comment on `setBreed` — *"Nothing to do here, Paper does not implement this"* — is not quite right. `Ageable.setBreed` is `@Deprecated(since = "1.16.2")` in favour of `Breedable`, but it has real semantics and `AgeableMock` implements it. A setter that silently discards its argument is worth avoiding.
* `setFuseTicks` enforces `ticks == -1 || ticks > 0`. The API javadoc documents no constraint, so rejecting `0` is an invented rule that could fail plugin tests which would pass on a real server — worth confirming against the server implementation.

src/main/java/org/mockbukkit/mockbukkit/entity/EntityMock.java
4everTheOne: Fixed by https://github.com/MockBukkit/MockBukkit/pull/1592/changes/26eb0e8afa1b619f23284a821885b9ed7fd11437

src/main/java/org/mockbukkit/mockbukkit/entity/SulfurCubeMock.java
4everTheOne: Fixed by https://github.com/MockBukkit/MockBukkit/pull/1592/changes/f4985cf7e6e9f45edd3778950bfb3f702f3e7dc0

thelooter (approved):

Issue #1610, opened 2026-08-15
Async scheduler behaviour is undocumented and cannot be controlled from a test


svaningelgem:
Async scheduler behaviour is not documented, and there is no way to control it from a test.

### The documentation half

`BukkitSchedulerMock.runTaskAsynchronously` does `pool.execute(...)` — a real thread pool, run immediately on submit. That is a perfectly reasonable choice, but it is not written down anywhere, and from the outside it is indistinguishable from "runs on performTicks", "runs inline" or "never runs".

We could not tell which it was, so we tested our drain method directly and left the scheduling wiring untested. A bug in that wiring would not have been caught. I suspect we are not the only ones who did this.

Even with no code change, stating the behaviour in the javadoc would be worth having.

### The capability half

Real threads make async tests non-deterministic — you either sleep or you race. What I would like:

```java
server.getScheduler().setAsyncMode(AsyncMode.INLINE);      // run on submit, same thread
server.getScheduler().setAsyncMode(AsyncMode.ON_TICK);     // run during performTicks
server.getScheduler().setAsyncMode(AsyncMode.REAL_THREADS); // today's behaviour
server.getScheduler().awaitAsyncTasks(Duration.ofSeconds(1));
```

`INLINE` would make most async plugin code testable with no sleeping at all.

### In scope

- Documenting the current behaviour
- The three modes and `awaitAsyncTasks`
- `REAL_THREADS` stays the default, so nothing changes for existing tests

### Not in scope

- Changing the default
- Region/folia schedulers

### What I would like input on

1. Would you take the documentation change on its own, if the modes are too much?
2. If the modes are welcome — is `REAL_THREADS` the right default to keep, or would `INLINE` be a better default in a future major?

Same on `v1.21` and `minecraft/v26.2`.

Issue #1609, opened 2026-08-15
No controllable clock for tests that measure elapsed time


svaningelgem:
Tests that involve time have no way to control it, so they have to sleep.

`BukkitSchedulerMock.performTicks` is good and we lean on it. What is missing is a clock that plugin code can read and a test can advance, for anything that measures elapsed wall-clock time rather than ticks.

### Two problems we hit

**Granularity.** Consecutive `Instant.now()` calls can return the same instant on Windows, so code that orders by timestamp is untestable without a sleep. We have eleven `Thread.sleep` calls in one test suite, every one a workaround, and each is a flake risk on a loaded CI box in both directions.

**Expiry windows.** A ten-second expiry cannot be waited out in a test, so we added a package-private constructor taking the duration purely so a test can pass 20 ms. That is production code shaped by a test limitation, which is the part that actually bothers me.

### What I would like to do

A clock on `ServerMock` that tests can advance, and that `performTicks` advances too:

```java
server.getClock().advance(Duration.ofSeconds(11));
server.getScheduler().performTicks(220);   // also advances it
```

### In scope

- A `Clock` obtainable from `ServerMock`
- `performTicks` advancing it by 50 ms per tick

### Not in scope

- Making anything in MockBukkit itself depend on it, beyond the scheduler
- Intercepting `System.currentTimeMillis` or `Instant.now` — plugins would opt in by reading the server clock

### What I would like input on

1. Is this something you want in MockBukkit at all, or do you consider clock injection the plugin's own problem?
2. If yes — expose `java.time.Clock`, or a MockBukkit type with `advance`?

Nothing exists on either `v1.21` or `minecraft/v26.2` today.

PR #1611, opened 2026-08-15
Implement stored-state entity flags

svaningelgem:
Implements four getter/setter pairs in the `entity` package that were unimplemented stubs. All of them are plain stored state with no behaviour attached.

| Class | Methods | Default |
|---|---|---|
| `LivingEntityMock` | `getRemoveWhenFarAway` / `setRemoveWhenFarAway` | `true` |
| `EntityMock` | `isVisibleByDefault` / `setVisibleByDefault` | `true` |
| `DisplayMock` | `getBillboard` / `setBillboard` | `FIXED` |
| `ItemMock` | `isUnlimitedLifetime` / `setUnlimitedLifetime`, `getThrower` / `setThrower` | `false`, `null` |

`setBillboard` rejects null with `Preconditions.checkArgument`, matching the neighbouring `setTransformation`.

### Why

`UnimplementedOperationException` extends `TestAbortedException`, so a test touching any of these is reported as *skipped* rather than failed and the build stays green. That makes the gap easy to miss: in our own suite, seven tests across two classes had been silently skipping on `setRemoveWhenFarAway` because the drop-ship entity we spawn sets it.

Worth mentioning as evidence — when I wrote the tests in this PR before the implementation, **10 of the 11 reported as skipped rather than failed**.

### Tests

11 tests covering the default, the round trip in both directions, and the null argument for `setBillboard`. 100% line and branch coverage on all eight methods (`setBillboard` 3/3 lines, 2/2 branches; the rest have no branches).

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/DisplayMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/DisplayMock.java
@@ -10,7 +10,6 @@ import org.joml.Matrix4f;
 import org.joml.Quaternionf;
 import org.joml.Vector3f;
 import org.mockbukkit.mockbukkit.ServerMock;
-import org.mockbukkit.mockbukkit.exception.UnimplementedOperationException;
 
 import java.util.UUID;
 
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/DisplayMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/DisplayMock.java
@@ -18,6 +18,7 @@ public class DisplayMock extends EntityMock implements Display
 {
 
 	private Matrix4f transformationMatrix;
+	private @NotNull Billboard billboard = Billboard.FIXED;
 	private int interpolationDuration = 0;
 	private int teleportDuration = 0;
 	private float viewRange = 1.0f;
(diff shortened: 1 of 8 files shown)


svaningelgem:
The `build` failure here looks unrelated to this change — could someone re-run it? I do not have permission to.

Failure is `BukkitSchedulerMockTest.longRunningTask_Throws_RunTimeException` at line 335, `assertEquals(1, scheduler.getActiveRunningCount())`.

Why I think it is a flake rather than this PR:

- This PR only touches stored-state flags on `DisplayMock`, `EntityMock`, `ItemMock` and `LivingEntityMock`. Nothing it adds is reachable from the scheduler.
- That test runs a real thread, sleeps, and waits on `countDownLatch.await(1, TimeUnit.SECONDS)` before asserting the running count — so a loaded runner can miss the window.
- It passes locally on this exact branch.
- `checkstyle` is green again after the unused-import fix, and #1612 and #1613 built fine from the same base.

Incidentally this is a small example of what #1609 and #1610 are about: with a controllable clock, or an inline async mode, this assertion would not depend on how busy the runner is.

Thorinwasher:
Do you have any reason for opening the PR against the 1.21 branch instead of the minecraft/v26.2 branch?

Same applies to all other PRs

svaningelgem:
Yes, as discussed and approved on Discord:

https://discord.com/channels/792754410576019477/809866105275023381/1537810749395832856

> @4everTheOne 
> Hello Steve! Sorry for the late reply I missed your message. Each version has their own branch. For example for v1.21 is branch "v1.21". Feel free to clone this branch, propose the changes and open a PR for it. Us maintainers will take care of the versioning ☺️

What I want to do is get these in the codebase, and then I merge them against the 2.26 branch.

That way no additional checks need to be done.

(And I need it for a 1.21.11 plugin - so plugging it into the 26 branch won't work for me sadly enough)

PR #1612, opened 2026-08-15
Implement numberFormat on ObjectiveMock and ScoreMock

svaningelgem:
Implements Paper's scoreboard number-format API, which was an unimplemented stub on both `ObjectiveMock` and `ScoreMock`:

- `ObjectiveMock#numberFormat()` / `numberFormat(NumberFormat)`
- `ScoreMock#numberFormat()` / `numberFormat(NumberFormat)`

Stores the format and hands it back, nullable in both directions, matching how the neighbouring `displayName` accessors behave.

### Why

A plugin that blanks a sidebar's score column with `objective.numberFormat(NumberFormat.blank())` cannot be exercised under MockBukkit at all — the call aborts the test. Because `UnimplementedOperationException` extends `TestAbortedException` that shows up as a skip, not a failure, so it is easy to miss.

We currently work around it by installing a `ServerMock` subclass that returns a stubbed scoreboard chain purely to get past this one call, which means nothing in our suite asserts on sidebar rendering. This lets that scaffolding go.

### Tests

6 tests, three per class: the default is null, the value round trips, and null is accepted to clear it. 100% line coverage on all four methods; no branches.

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/scoreboard/ScoreMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/scoreboard/ScoreMock.java
@@ -20,6 +20,7 @@ public class ScoreMock implements Score
 	private @Nullable OfflinePlayer player = null;
 	private int score = 0;
 	private boolean set = false;
+	private @Nullable NumberFormat numberFormat = null;
 
 	/**
 	 * Constructs a new {@link ScoreMock} for the provided objective with the specified entry.
@@ -132,15 +133,13 @@ public class ScoreMock implements Score
 	@Override
 	public @Nullable NumberFormat numberFormat()
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		return this.numberFormat;
 	}
 
 	@Override
 	public void numberFormat(@Nullable NumberFormat numberFormat)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		this.numberFormat = numberFormat;
 	}
 
 }
--- a/src/main/java/org/mockbukkit/mockbukkit/scoreboard/ObjectiveMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/scoreboard/ObjectiveMock.java
@@ -31,6 +31,7 @@ public class ObjectiveMock implements Objective
 	private @NotNull Component displayName;
 	private @Nullable DisplaySlot displaySlot;
 	private RenderType renderType;
+	private @Nullable NumberFormat numberFormat = null;
 
 	/**
 	 * Constructs a new {@link ObjectiveMock} with the provided parameters.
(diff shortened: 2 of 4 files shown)


src/main/java/org/mockbukkit/mockbukkit/scoreboard/ScoreMock.java
```diff
@@ -132,15 +133,13 @@ public void customName(@Nullable Component customName)
...
+		return this.numberFormat;
 	}
 
 	@Override
 	public void numberFormat(@Nullable NumberFormat numberFormat)
 	{
-		// TODO Auto-generated method stub
-		throw new UnimplementedOperationException();
+		this.numberFormat = numberFormat;
 	}
```
4everTheOne, line 143: Looking into the PaperMc code, the logic is delgated to the Objective. So i would expect this to delegate to this.objective.numberFormat(...)

4everTheOne (changes requested):
Just a small change to match Paper implementation

src/main/java/org/mockbukkit/mockbukkit/scoreboard/ScoreMock.java
svaningelgem: Thanks for looking. I went to check `CraftScore` in the 1.21.11 dev bundle before changing it, and I think the delegation is narrower than it looks — could you confirm which part you meant?

What Paper delegates to the objective is the **state check and the route to the scoreboard**, not the format itself:

```java
@Override
public io.papermc.paper.scoreboard.numbers.NumberFormat numberFormat() {
    this.objective.checkState();

    ReadOnlyScoreInfo scoreInfo = this.objective.getScoreboard().getHandle()
        .getPlayerScoreInfo(this.entry, this.objective.getHandle());
    ...
}

@Override
public void numberFormat(io.papermc.paper.scoreboard.numbers.NumberFormat format) {
    this.objective.checkState();

    final net.minecraft.world.scores.ScoreAccess access = this.objective.getScoreboard()
        .getHandle().getOrCreatePlayerScore(this.entry, this.objective.getHandle());
    ...
    access.numberFormatOverride(io.papermc.paper.util.PaperScoreboardFormat.asVanilla(format));
}
```

The setter writes `numberFormatOverride` on that score's own `ScoreAccess`, and the getter reads that score's `ScoreInfo` — whereas `CraftObjective#numberFormat` reads and writes `this.objective.numberFormat()`, which is the objective's own field. They are two different pieces of state, with the score's acting as an override.

So if `ScoreMock` delegated the format to the objective, setting one score's format would change every score on that objective, which I do not think is what you want.

Happy to be wrong if you are looking at something I have missed.

**What I can take from this**, if it is the delegation you were pointing at: Paper routes the score's accessors through `this.objective.checkState()`, so a score belonging to an unregistered objective throws rather than answering. `ObjectiveMock` already has a private `validate()` doing exactly that check — it just has no callers anywhere in the codebase today. I would happily wire both `ScoreMock` accessors and the `ObjectiveMock` ones through it, either here or in a follow-up, whichever you prefer.

Which would you like?

PR #1613, opened 2026-08-15
Credit indirect kills to the player behind them

svaningelgem:
`LivingEntityMock#setHealth(0)` derives the killer from the damage source's **direct** entity only:

```java
if (getLastDamageCause() != null
        && getLastDamageCause().getDamageSource().getDirectEntity() instanceof Player player)
{
    setKiller(player);
}
```

For an arrow kill the direct entity is the arrow, not the shooter, so `getKiller()` stays null here while a real server sets it. The same applies to anything indirect: projectiles, TNT, potions.

This is a behavioural divergence from `CraftLivingEntity` rather than a missing feature, which is why it is worth fixing even though a workaround exists — a test that calls `setKiller` by hand asserts on state the test itself set, not on state the mock derived.

### The fix

Resolves in the order `CraftLivingEntity` uses:

1. direct entity, if it is a player
2. causing entity, if it is a player
3. the shooter behind a projectile, if it is a player

Extracted into a private `deriveKiller()` with a javadoc explaining the ordering.

### Tests

6 tests, one per branch: no last damage, a direct player hit, a projectile with a causing player, a projectile whose shooter is a player, a projectile with no shooter, and a non-player damager.

Worth noting how the red phase looked — the four tests describing *existing* behaviour passed immediately and only the two projectile cases failed, which isolates the defect to that one condition.

100% coverage: `deriveKiller` 11/11 lines and 10/10 branches, `setHealth` 16/16 lines and 6/6 branches.

### Deliberately not included

Paper also applies a 100-tick kill-credit window, so a victim that dies long after the last hit is not credited. That needs damage-time tracking and would change behaviour for existing death tests, so it felt like a separate change rather than part of a divergence fix. Happy to add it here or in a follow-up if you would prefer it bundled.

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
@@ -191,6 +191,38 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 		return !isDead();
 	}
 
+	/**
+	 * Works out which player, if any, should be credited with this entity's death, following the same order
+	 * {@code CraftLivingEntity} uses: the entity that dealt the damage directly, then the entity that caused it, then
+	 * the shooter behind a projectile. Checking only the direct entity would miss every indirect kill, so an arrow
+	 * fired by a player would leave the killer unset here while setting it on a real server.
+	 *
+	 * @return the player to credit, or null if the last damage cannot be traced back to one.
+	 */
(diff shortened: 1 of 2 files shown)


svaningelgem:
Opened #1614 for the 100-tick kill-credit window mentioned above, since it needs a direction on whether time-based behaviour belongs in the mock before anyone writes it. This PR stands on its own without it.

4everTheOne (approved):
Thanks for this contribution

PR #1616, opened 2026-08-15
Stop a command outliving its plugin from throwing

svaningelgem:
`ServerMock#reload()` throws `NullPointerException` if any Brigadier command has been registered.

```
java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:233)
	at org.mockbukkit.mockbukkit.command.brigadier.ApiCommandMetaMock.plugin(ApiCommandMetaMock.java:23)
	at org.mockbukkit.mockbukkit.command.brigadier.BrigadierUtils.wrapNode(BrigadierUtils.java:96)
	at ...bukkit.BukkitBrigadierForwardingMapMock.nodeToEntry(BukkitBrigadierForwardingMapMock.java:103)
	at ...bukkit.BukkitBrigadierForwardingMapMock.entrySet(BukkitBrigadierForwardingMapMock.java:95)
	at org.bukkit.command.SimpleCommandMap.clearCommands(SimpleCommandMap.java:179)
	at org.mockbukkit.mockbukkit.ServerMock.reload(ServerMock.java:1667)
```

### Cause

```java
@Nullable
public Plugin plugin()
{
	return this.pluginMeta == null ? null : Objects.requireNonNull(Bukkit.getPluginManager().getPlugin(this.pluginMeta.getName()));
}
```

The method is annotated `@Nullable` and then makes it impossible to return null for a command that has plugin meta. That holds right up until the plugin stops being registered — which is exactly what reload does, before walking the command map to clear it. The command outlives the plugin instance that added it, the lookup returns null, and `requireNonNull` turns a normal state into a crash.

### Fix

Return the lookup result as-is, matching the declared `@Nullable`, and document why null is a real answer rather than a defect.

### Tests

- Clearing the command map after the owning plugin is gone no longer throws. This is the reload path, reproduced directly — `reload()` itself cannot re-instantiate a builder-made `PluginMock`, which is a separate limitation and would have made the test about the wrong thing.
- `plugin()` returns null once the plugin is gone, and null when there is no plugin meta at all.

Full suite green: 20610 tests, 0 failures. `ApiCommandMetaMock#plugin` at 100% line and branch.

### Note

Found while working on #1615, but it reproduces on an unmodified checkout of this branch — registering a command through the `COMMANDS` lifecycle event, dispatching once so the old lazy initialisation runs, then calling `reload()`. Independent of that change, so it is here on its own.

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/command/brigadier/ApiCommandMetaMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/command/brigadier/ApiCommandMetaMock.java
@@ -6,7 +6,6 @@ import org.bukkit.plugin.Plugin;
 import org.jspecify.annotations.Nullable;
 
 import java.util.List;
-import java.util.Objects;
 
 public record ApiCommandMetaMock(@Nullable PluginMeta pluginMeta, @Nullable String description, List<String> aliases,
 								 @Nullable String helpCommandNamespace, boolean serverSideOnly)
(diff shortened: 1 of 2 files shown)


4everTheOne (approved):
Thanks for this contribution

PR #1617, opened 2026-08-15
Implement teleportAsync

svaningelgem:
`EntityMock#teleportAsync` was unimplemented, so any code that moves an entity the Paper way could not be exercised.

```java
entity.teleportAsync(destination, TeleportCause.PLUGIN);
```

Worth noting how that presents: `UnimplementedOperationException` extends `TestAbortedException`, so the test is reported as *skipped* rather than failed and the build stays green. A test can stop running and nothing says so.

### The fix

Delegate to the synchronous `teleport` and hand back an already-completed future. The mock teleports immediately — there is nothing to await — and going through the same path means the teleport event fires exactly as it does for the blocking call, rather than the async variant quietly behaving differently.

### Tests

Two: the entity ends up at the destination with a future that is already done and reports success, and the teleport event still fires.

Full suite green: 20609 tests, 0 failures. `teleportAsync` at 100% line coverage, checkstyle clean.


4everTheOne (approved):
Thanks for this contribution

PR #1622, opened 2026-08-15
Implement server tick metrics

svaningelgem:
`ServerMock#getTPS`, `getTickTimes` and `getAverageTickTime` were all unimplemented, so anything reporting on server load could not be tested.

```java
Bukkit.getTPS();
```

### The fix

`getTPS` and `getTickTimes` return stored values; `getAverageTickTime` is **derived from the tick times** rather than stored alongside them. That is the part worth a second look: if the average were its own field, a test could set tick times and an average that disagreed with each other, and code under test would see a server that cannot exist. Deriving it means the two cannot contradict.

Both getters hand back a copy, so a caller cannot reach in and mutate the server's state through the returned array.

### Why there are setters

A mock cannot observe how long its own ticks took — there is no real tick loop to measure. So `setTPS` and `setTickTimes` let a test state what the server should report, which is the only way code that reacts to load can be exercised at all. Neither has a Bukkit setter; both are MockBukkit-specific and documented as such.

`setTickTimes` rejects null and empty, since an empty array would make the average a division by zero.

### Tests

Ten: defaults for all three, both getters returning copies rather than the live array, the setters round-tripping, the average being the mean in milliseconds, and the two rejection cases.

Full suite green: 20617 tests, 0 failures. 12 added lines, 0 uncovered, checkstyle clean.

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/ServerMock.java
@@ -205,6 +205,8 @@ public class ServerMock extends Server.Spigot implements Server
 {
 
 	private Component motd = Component.text("A Minecraft Server");
+	private double[] tps = { 20.0, 20.0, 20.0 };
+	private long[] tickTimes = new long[100];
 	private static final Component NO_PERMISSION = Component.text("I'm sorry, but you do not have permission to perform this command. Please contact the server administrators if you believe that this is in error.", NamedTextColor.RED);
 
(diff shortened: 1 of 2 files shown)


Thorinwasher:
Do you use these methods for your plugin?

svaningelgem:
yes: one of my plugins calls Bukkit.getTPS() and Bukkit.getAverageTickTime().

Its BukkitAPITest currently asserts the exception to document the boundary

PR #1623, opened 2026-08-15
Implement client brand name and current input on PlayerMock

svaningelgem:
`PlayerMock#getClientBrandName` and `getCurrentInput` were both unimplemented, so anything that reacts to what the client is or what it is pressing could not be tested.

```java
player.getClientBrandName();   // org.mockbukkit.mockbukkit.exception.UnimplementedOperationException
player.getCurrentInput();      // same
```

### Why there are setters

Neither value has a Bukkit setter, and both are things the *client* tells the server: the brand arrives over the `minecraft:brand` plugin channel after login, and the input is sent as the player presses keys. There is no server-side call that sets either.

That leaves a test with no way to arrange the state it wants to exercise, so each gets a MockBukkit-specific `setClientBrandName` / `setCurrentInput`, documented as standing in for what the client would have sent.

### Why InputMock exists

`org.bukkit.Input` is an interface with seven accessors and no implementation anywhere in the API — nothing to hand to `setCurrentInput`. `InputMock` is a record implementing it, plus `InputMock.none()` for the common case of a player standing still, which is also what `getCurrentInput` returns by default.

### Tests

Seven: the brand defaults to null and round trips including back to null, the input defaults to nothing pressed, a set input is returned as given, null is rejected, and `InputMock` reports each of its seven keys independently plus the `none()` shorthand.

Full suite green: 20614 tests, 0 failures. 16 added lines, 0 uncovered, checkstyle clean.

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/InputMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/InputMock.java
@@ -1,6 +1,7 @@
 package org.mockbukkit.mockbukkit.entity;
 
 import lombok.Builder;
+import lombok.Getter;
 import org.bukkit.Input;
 
 /**
@@ -17,8 +18,8 @@ import org.bukkit.Input;
  * @param sprint   Whether the sprint key is pressed.
  */
 @Builder
-public record InputMock(boolean forward, boolean backward, boolean left, boolean right, boolean jump, boolean sneak,
-						boolean sprint) implements Input
+public record InputMock(@Getter boolean forward, @Getter boolean backward, @Getter boolean left,
+						@Getter boolean right, @Getter boolean jump, @Getter boolean sneak, @Getter boolean sprint) implements Input
 {
 
 	/**
@@ -29,48 +30,6 @@ public record InputMock(boolean forward, boolean backward, boolean left, boolean
 	public static InputMock none()
 	{
 		return builder().build();
-	}
-
-	@Override
-	public boolean isForward()
-	{
-		return this.forward;
-	}
-
-	@Override
-	public boolean isBackward()
-	{
-		return this.backward;
-	}
-
-	@Override
-	public boolean isLeft()
-	{
-		return this.left;
-	}
-
-	@Override
-	public boolean isRight()
-	{
-		return this.right;
-	}
-
-	@Override
-	public boolean isJump()
-	{
-		return this.jump;
-	}
-
-	@Override
-	public boolean isSneak()
-	{
-		return this.sneak;
-	}
-
-	@Override
-	public boolean isSprint()
-	{
-		return this.sprint;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


src/main/java/org/mockbukkit/mockbukkit/entity/InputMock.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @param forward  Whether the forward key is pressed.
+ * @param backward Whether the backward key is pressed.
+ * @param left     Whether the left key is pressed.
+ * @param right    Whether the right key is pressed.
+ * @param jump     Whether the jump key is pressed.
+ * @param sneak    Whether the sneak key is pressed.
+ * @param sprint   Whether the sprint key is pressed.
+ */
+public record InputMock(boolean forward, boolean backward, boolean left, boolean right, boolean jump, boolean sneak,
```
4everTheOne: A Builder pattern would be usefull here, so if someone wants can simply do:

```java
var input = InputMock.builder()
    .forward(true)
    .right(true)
    .build()
```

This can be achieved with the `@Builder` annotation from lombok.
To make it even simpler to use we could create alias like this:

```java
var input = InputMock.builder()
    .forward()
    .right()
    .build()
```

Here is an example in how to do it:
```java
@Builder
public record InputMock(boolean forward, boolean backward, boolean left, boolean right, boolean jump, boolean sneak, boolean sprint) implements Input
{
   // ...
   
	public static class InputMockBuilder
	{

		public InputMockBuilder backward()
		{
			return this.backward(true);
		}

		public InputMockBuilder forward()
		{
			return this.backward(true);
		}
		
		// ...

	}
}
```

4everTheOne (commented):
Looks good to me, just a have a improvement to make the input easier to use

src/main/java/org/mockbukkit/mockbukkit/entity/InputMock.java
svaningelgem: Done in 48ce150.

Lombok `@Builder` on the record, plus a hand written `InputMockBuilder` with the shorthands you sketched:

```java
InputMock.builder().forward().right().build()
```

One wrinkle worth recording: Lombok skips a builder setter when a method of that **name** already exists, regardless of signature. Declaring `forward()` therefore suppressed the generated `forward(boolean)`, so the explicit setters are written out next to the shorthands. Lombok still supplies the builder fields and `build()`, and both forms now work:

```java
InputMock.builder().forward().right().build()
InputMock.builder().forward(true).right(pressed).build()
```

`none()` delegates to `builder().build()`. Three tests added for the shorthands, the explicit values, and the empty default; full suite green (20617 tests, 0 failures), checkstyle clean.

svaningelgem:
Build fails due to flaky `BukkitSchedulerMockTest.longRunningTask_Throws_RunTimeException()`

This is already addressed in #1588

PR #1624, opened 2026-08-15
Let simulateDamage carry the real damage cause

svaningelgem:
`simulateDamage` reports `ENTITY_ATTACK` whenever the damage source has a direct entity:

```java
EntityDamageEvent event;
if (source.getDirectEntity() != null)
{
	event = new EntityDamageByEntityEvent(source.getDirectEntity(), livingEntityMock,
			EntityDamageEvent.DamageCause.ENTITY_ATTACK, source, amount);
}
```

An explosion, a projectile and a splash potion all have a direct entity too. So all three arrive at the listener claiming to be a melee hit, and a listener that switches on `getCause()` cannot be tested for any of them — there is no way to make `simulateDamage` produce `PROJECTILE`, `BLOCK_EXPLOSION`, `MAGIC` or anything else.

The workaround is to stop using `simulateDamage` and hand-build `EntityDamageByEntityEvent` instead, which is its own problem: every constructor is deprecated.

### The fix

An overload taking the cause:

```java
livingEntity.simulateDamage(1.0, source, DamageCause.PROJECTILE);
```

The existing two-argument forms are unchanged. They now derive the same default they always produced — `ENTITY_ATTACK` with a direct entity, `CUSTOM` without — so nothing that calls them behaves differently.

Exposed on `LivingEntitySimulation` and mirrored on `LivingEntityMock` alongside the existing delegates.

### Tests

Eight. The new overload reports the given cause both with and without a direct entity, and rejects a null source or cause. The existing defaults are pinned by tests of their own so a later change cannot quietly alter them, and the previously uncovered paths — a cancelled event leaving health alone, and a non-human damager being a mob attack — are covered too.

Full suite green: 20615 tests, 0 failures. 9 added lines, 0 uncovered, checkstyle clean. All three `simulateDamage` overloads now at 100% branch coverage, up from two of them having gaps.

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/entity/LivingEntityMock.java
@@ -306,6 +306,20 @@ public abstract class LivingEntityMock extends EntityMock implements LivingEntit
 	}
 
 	/**
+	 * Simulate damage to this entity and throw an event, stating what caused it.
+	 *
+	 * @param amount The amount of damage to be done.
+	 * @param source The damage source.
+	 * @param cause  The cause to report on the event.
+	 * @return The EntityDamageEvent that got thrown.
+	 */
+	public EntityDamageEvent simulateDamage(double amount, @NotNull DamageSource source,
+			EntityDamageEvent.@NotNull DamageCause cause)
+	{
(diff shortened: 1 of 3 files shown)


4everTheOne (approved):
Thank you for your contribution. We can't merge right now, because we will need to resolve the problem with tags in old versions before merging this PR

PR #1634, opened 2026-08-16
Implement rayTraceBlocks

### LINKED ISSUE #1607: WorldMock.rayTraceBlocks is unimplemented in every overload
`WorldMock.rayTraceBlocks` is unimplemented in every overload, as is the `LivingEntityMock` pair.

Any plugin that does line-of-sight or block-picking work cannot be tested. I am willing to implement it, but it is the largest item on my list and I would rather agree the shape before writing it.

### What I would like to do

A straightforward voxel walk (Amanatides–Woo) over the blocks between start and `start + direction * maxDistance`, returning the first block that matches, with the hit position and face filled in.

### In scope

- `WorldMock.rayTraceBlocks(...)`, all overloads
- `LivingEntityMock.rayTraceBlocks(...)`
- Honouring `FluidCollisionMode`, `ignorePassableBlocks` and the `canCollide` predicate

### Not in scope

- Per-block collision shapes — I would treat every non-passable block as a full cube. Stairs, slabs and fences would report a hit at the cube boundary rather than the true surface.
- `rayTraceEntities` and the combined `rayTrace`, unless you want them in the same change

### What I would like input on

1. Is full-cube collision an acceptable simplification, or does this need real block shapes to be worth merging?
2. Should `rayTraceEntities` land in the same PR or a separate one?

Unimplemented on both `v1.21` and `minecraft/v26.2`.

svaningelgem:
Closes #1607.

Every `WorldMock#rayTraceBlocks` overload was unimplemented, as were both on `LivingEntityMock`, so any line-of-sight or block-picking code was untestable.

```java
world.rayTraceBlocks(start, direction, 10);
```

### How it traces

A voxel walk between `start` and `start + direction * maxDistance`. At each step it crosses whichever axis boundary comes next and reports the face it entered through, so `getHitBlockFace` is `WEST` for a ray travelling east, `UP` for one travelling down, and so on.

Starting inside a block that already collides is an immediate hit with a null face, matching Paper — there is no face to have entered through.

The `LivingEntityMock` overloads trace from the eye location along the entity's facing.

### The trade-off, stated plainly

**Collision is a full cube.** A stair, slab, fence or open trapdoor reports a hit at the block boundary rather than at its real surface. For line-of-sight tests — the common case, "can this entity see that one", "what block is the player looking at" — that is indistinguishable from the real answer. For anything measuring precise geometry against a partial block, it is not.

Adding real block shapes is a bigger change and would want its own review. Happy to do it if you would rather have it, or to note the limitation in the javadoc if full cubes are acceptable.

`FluidCollisionMode`, `ignorePassableBlocks` and the `canCollide` predicate are all honoured properly.

### One small scope addition

`BlockMock#isPassable` was itself an unimplemented stub, and `ignorePassableBlocks` cannot be honoured without it, so it is implemented here as `!material.isSolid()` — `Material` already carries that distinction, so there is no table to maintain. Flagging it rather than burying it in the diff; happy to split it out if you would prefer.

### Deliberately out of scope

`rayTraceEntities` and the combined `rayTrace` are still stubs. They need an entity-intersection model rather than a block walk, which felt like a separate change. Do you want them as a follow-up?

### Tests

21: a hit ahead with the correct block and face, all six directions, nothing in the way, stopping at max distance, starting inside a block, leaving the world upwards and downwards, all three fluid modes including flowing water and a liquid with no level, passable blocks with the flag on and off, the `canCollide` predicate, zero distance, and the four argument rejections.

Full suite green: 20628 tests, 0 failures. 74 added lines, 0 uncovered, checkstyle clean.

### DIFF
--- a/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
+++ b/src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
@@ -37,6 +37,8 @@ import org.bukkit.WorldCreator;
 import org.bukkit.WorldType;
 import org.bukkit.block.Biome;
 import org.bukkit.block.Block;
+import org.bukkit.block.data.Levelled;
+import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
 import org.bukkit.block.data.BlockData;
 import org.bukkit.block.data.Waterlogged;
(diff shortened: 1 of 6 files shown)


src/main/java/org/mockbukkit/mockbukkit/world/WorldMock.java
```diff
@@ -3045,4 +3125,60 @@ private static <MODERN, LEGACY> MODERN convertLegacyValue(GameRule<MODERN> rule,
...
+	}
+
+	private static int signum(double value)
+	{
+		if (value > 0)
+		{
+			return 1;
+		}
+		return value < 0 ? -1 : 0;
+	}
```
ltuffery: Why not use Java’s `Math.signum()` method?

svaningelgem: Correct. I've adjusted it.

